PR #48, opened 2017-01-27
PacMan "cabinet" hardware support skeleton

zx70:


suborb:
The .opt file is no longer required.

Also sections are needed the DEFVARS support was removed as soon as sections were available.

zx70:
tnx, fixed.

PR #79, opened 2017-02-20
Allow sccz80 to do r2l calls

suborb:
This isn't enabled by default yet. Command flag -set-r2l-by-default has to be supplied to turn on this behaviour.


suborb:
Shows up too many errors on linux builds. Needs investigation.

suborb:
Let's try again...

PR #88, opened 2017-02-25
Fix #86: Allow r2l functions to be defined

### LINKED ISSUE #85: sccz80: When calling r2l add the '_' prefix to library functions
This way the implementation can be shared with sdcc and we don't have duplicate library code.

### LINKED ISSUE #86: sccz80: Handle function definitions that are r2l
Although we can pass parameters in r2l order, we can't define functions in r2l order yet.

Need to jiggle the local symbol table and set the offsets appropriately.

suborb:
Fix #85: Use the r2l entry point when in r2l mode.

This starts to mean we can share library code between sdcc and sccz80

@aralbrec can you give this a bit of a whirl with your test programs?


aralbrec:
The LIB change where leading underscores now appear in function names makes it incompatible with the new c lib currently so I was unable to test tonight.

suborb:
I've backed out the prefix change to sccz80.

aralbrec:
I've tried test programs against sccz80 with -set-r2l-by-default and all seem to work fine.  The __smallc attribute is being honoured and the rest of the functions are r2l.

suborb:
Thanks Alvin, that matches what I was seeing. I'm going to merge this one to get it in and then open another issue to tidy some bits I thought about (changing the option name + providing __stdc option to allow a slow migration.

PR #102, opened 2017-03-10
Feature/use ticks for tests + fix const long promotion + fix doubles

suborb:
Import the hooks in and tweak.


suborb:
@aralbrec when this is merged, you'll need to add dswap to the new library (sccz80 flavours)

PR #126, opened 2017-03-22
Feature/sccz80 enhancements v2

suborb:
A lot of changes here.

* Optimisations for common int/long operations.
* Allow static initialisation of doubles
* Better handling of initialisation without a need to cast in code the type.
* Fold double constant expressions.
* Add #pragma printf and #pragma scanf to pragma
* Optimisation rules

Probably best to hold off on merge until @aralbrec has finished the benchmarking - it changes quite a few of the results there (for the better!)


aralbrec:
If it's ready to merge, go ahead and merge.  I've finished the first pass of the benchmarks that have been committed and they can be revisited any time if the Z88DK date information is updated.

Issue #159, opened 2017-04-11
z80asm: List file handling improvements


suborb:
As part of improving the classic build to support multiple configurations I'm creating a load of extra .lst files which are identical apart from part of the path.

It would be handy if there could be sort of substitution on the contents of these files so that an environment variable could be substituted, e.g.:

    stdio/obj/${CONFIG_TYPE}/fopen.o

and if:

    CONFIG_TYPE=stdesc

then the file that ends up being added to the file list is:

    stdio/obj/stdesc/fopen.o

Should the environment variable not exist, then replace with an empty string (to match zcc behaviour).

Additionally, if comment lines could be supported that would be great.

PR #160, opened 2017-04-11
Issue #159. Allow environment variable substitution in list files

suborb:
I've taken some code from zcc to do this.

I think there's possibly a leak every time a filename has a substitution in it.


suborb:
Thanks Paulo, I wasn't sure how you were doing tests in z80asm so I'll watch and learn.

pauloscustodio:
Changes accepted and commited.
Extended expansion of environment variables to all command line options and arguments.

PR #173, opened 2017-04-16
Fix #171: Remove dead and old win32 project files

### LINKED ISSUE #171: z88dk: Cleanup windows build files
There's a whole bunch of windows batch and solution files hanging around the place:

````
src/*/gccmake.bat
src/*/vscmake.bat
src/*/*.vcxproj
win32/*/*.vcxproj
````

The ones which are needed should remain, and the ones that aren't should be removed.

Given that z88dk compiles with Visual Studio Express and we're heavily dependent on a functioning make, that suggests the .bat files can be removed.

I'm guessing that the win32/*/*.vcxproj files are the ones that are actually used?

suborb:


suborb:
Alvin, Stefano I don't use windows so can't check myself, but can you give an ok to removing these files?

zx70:
The last time I checked that stuff it was all in working order.  Things may have changed an perhaps they are redundant.. I'm not sure.

aralbrec:
I successfully built everything in the branch using vs2015.

Stef, he is getting rid of the vs batch files.  I don't think they are needed anymore.  The win32 binaries are either cross-compiled via the makefile or built natively on vs2015 using the sln in win32/

zx70:
I can leave without them obviously.  They were meant to permit the free command line only version of visual c to be used, I don't know how the licensing logic is now.  Similarly the MinGw environment could build z88dk without msys. I admit that the lack of the make tool would have hit the library build soon  and that keeping them updated is a bit a waste of energy ..just wipe em

suborb:
Thanks, I'm just going through cleaning up some of the cruft we've gathered over the years. I suspect docs will be my next target.

PR #204, opened 2017-04-28
Use GNU make on BSD.

vext01:
This makes z88dk at least build on OpenBSD (and probably NetBSD and FreeBSD too).

There are some other issues, which I will raise over the weekend.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -39,13 +39,20 @@ export CC
 export CFLAGS
 export CCOPT
 
-MAKE="make"
-INSTALL="install"
-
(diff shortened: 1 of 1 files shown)


aralbrec:
Looks good to me.  I'll let someone whose main box is linux do the merge.

suborb:
Looks good to me, and builds on travis.

I've just tried a build on an OpenBSD box and it ran through without issues so I'll be interested in the other issues you've found and try to add in some tests for them.

vext01:
Thanks!

Issue #186, opened 2017-04-22
appmake: create sna snapshots for zx target


aralbrec:
This need is arising from the generation of 128k bankswitched programs on the zx in the new c lib.  The question is how appmake should handle the extra banks. 

A 128k .sna snapshot would be convenient for emulator development and if that's going to be done we might as well do a 48k snapshot too.  If appmake is in tap mode, the extra banks could either be output as separate tap files with or without header depending on existing options, or ignored so that it is up to the user to turn them into tap files.  I don't think an automatic tap loader for the whole thing would be general enough.

aralbrec:
Completed some time ago.  `-subtype=sna` on the compile line will generate 48k or 128k snapshots.

PR #218, opened 2017-04-30
Various build issues

suborb:
Sort out various build issues:

* Change build.sh to not override CC and CFLAGS that are set in the environment
* Sort out make install so things go to the right place (including fakeroot)
* Add a definition for INSTALL - it may not be set in all environments
* Remove zcc.cfg and the default configuration support (so removes all the legacy config file support)

@vext01 would appreciate your feedback on this, I think it addresses all of the issues you raised.


vext01:
Thanks. This is a clear improvement.

Note also, the following files have hard-coded `-g`:
 * src/sccz80/Makefile
 * src/ticks/Makefile

Makefile
```diff
@@ -12,15 +12,18 @@ prefix_share = $(prefix)/share
 git_rev = $(shell git rev-parse --short HEAD)
 version := $(shell date +%Y%m%d)
 
-# The default machine, the lib/config/DEFAULT.cfg file is copied to zcc.cfg
-DEFAULT = z88
+INSTALL ?= install
+CFLAGS ?= -g -O2
```
vext01: Did you mean to hard-code `-g`?

Makefile
```diff
@@ -12,15 +12,18 @@ prefix_share = $(prefix)/share
...
+CC ?= gcc
 
 # --> End of Configurable Options
 
-all: setup appmake copt zcpp sccz80 z80asm zcc zpragma zx7 z80nm ticks z80svg config testsuite
+export CC INSTALL CFLAGS
```
vext01: I didn't know you could do that in a Makefile!

Makefile
```diff
@@ -69,20 +72,18 @@ ticks:
...
 
 install-libs:
-	mkdir -p $(prefix)/lib/z88dk/lib/config
-	cp -R lib/config/* $(prefix)/lib/z88dk/lib/config/
-	find $(prefix)/lib/z88dk/lib/config -type f | xargs chmod 644
-	mkdir -p $(prefix)/lib/z88dk/lib/clibs
-	cp -R lib/clibs/* $(prefix)/lib/z88dk/lib/clibs/
-	find $(prefix)/lib/z88dk/lib/clibs -type f | xargs chmod 644
+	mkdir -p $(DESTDIR)/$(prefix_share)/z88dk/lib/config
+	cp -R lib/config/* $(DESTDIR)/$(prefix_share)/z88dk/lib/config/
```
vext01: Strictly speaking, you'd use $INSTALL, but if there's lots of sub-dirs, could be annoying.

build.sh
```diff
@@ -56,7 +54,8 @@ esac
 
 export INSTALL
 
-$MAKE -e
+$MAKE 
 $MAKE -C `pwd`/libsrc
 $MAKE -C `pwd`/libsrc install
 $MAKE -C `pwd`/libsrc/_DEVELOPMENT
+$MAKE -C test
```
vext01: Ah yes, I noticed that the tests are now run as part of the build. I will probably separate them for porting, but I think it's fine to run the tests like this for the main repo.

src/ticks/Makefile
```diff
@@ -8,7 +8,7 @@ INSTALL ?= install
 all: ticks$(EXESUFFIX)
 
 ticks$(EXESUFFIX):	$(OBJS)
-	${CC} -o ticks$(EXESUFFIX) ${CFLAGS} ${CCOPT} ${OBJS}
+	$(CC) -o ticks$(EXESUFFIX) $(CFLAGS)  $(OBJS)
```
vext01: double space?

src/zcc/Makefile
```diff
@@ -13,7 +13,7 @@ OBJS += $(REGEX_OBJS)
 all: zcc$(EXESUFFIX)
 
 zcc$(EXESUFFIX):	$(OBJS)
-	${CC} -o zcc$(EXESUFFIX) ${CFLAGS} ${CCOPT} ${OBJS}
+	$(CC) -o zcc$(EXESUFFIX) $(CFLAGS)  $(OBJS)
```
vext01: Another double space?

src/zcc/zcc.c
```diff
@@ -2699,7 +2706,7 @@ void tempname(char *filen)
 *
 *    If ZCCCFG doesn't exist then we take the c_install_dir/lib/config/zcc.cfg
 */
-int find_zcc_config_fileFile(char *arg, int gc, char *buf, size_t buflen)
+int 	find_zcc_config_fileFile(const char *program, char *arg, int gc, char *buf, size_t buflen)
```
vext01: Is the leading space before the func name intentional?

src/zcc/zcc.c
```diff
@@ -2733,34 +2740,10 @@ int find_zcc_config_fileFile(char *arg, int gc, char *buf, size_t buflen)
...
-		}
-		snprintf(buf, buflen, "%s/zcc.cfg", cfgfile);
-	}
-	else {
-		// Fall back to a sensible default
-		snprintf(buf, buflen, "%slib/config/zcc.cfg", c_install_dir);
-	}
-	return (gc);
+	} 
+	// Without a config file, we should just exit
```
vext01: Did we want to print some error message as well, so the user knows what is wrong?

support/graphics/Makefile
```diff
@@ -12,7 +12,7 @@ CFLAGS += `$(XML2CONFIG) --cflags`
 all: z80svg$(EXESUFFIX)
 
 z80svg$(EXESUFFIX):	$(OBJS)
-	$(CC) -o z80svg$(EXESUFFIX) $(CFLAGS) $(CCOPT) $(OBJS) $(LIBS)
+	$(CC) -o z80svg$(EXESUFFIX) $(CFLAGS)  $(OBJS) $(LIBS)
```
vext01: double space?

vext01:
While I have a channel to the z88dk devs, can I ask. Can I use the assembler to make a .tap image, like pasmo does? I know you can do this for C, but I don't think zcc understands asm files at present, right?

vext01:
Looks like I could do (from your slowdos repo):
```
appmake +zx -b slowdos.bin --org  32768 -o slowdos.tap
```

build.sh
vext01: Alternatively you could add a `-t` switch to run the tests? Up to you.

aralbrec:
Yes "appmake" is a separate utility within z88dk and can be called in its own right to manipulate output binaries.  Within zcc what happens is if there is a "-create-app" switch then zcc will call appmake as a final processing step to generate an appropriate file from the toolchain's output.

What appropriate means is determined in the target's cfg, eg [z88dk/lib/config/zx.cfg](https://github.com/z88dk/z88dk/blob/master/lib/config/zx.cfg) for the zx target.  In there you will see on the OPTIONS line that "subtype=default" and below you will see the "SUBTYPE default" line where "-Cz+zx" is added to the compile line.  "-Cz" passes the following argument to appmake so that zcc will invoke "appmake +zx" here plus add a few other switches so that appmake makes a standard tap file with loader by default.  The user can override this, for example, by adding "-subtype=rom" on the compile line to make an if2 cartridge.  Or he can just make up his own custom appmake invocation by adding -Cz"blah blah" to the compile line.  This will override the defaults in zcc.cfg.

So you can generate binaries and then call appmake directly yourself to make tap files, roms, whatever.  Entering "appmake" by itself on the command line will show all its flavours and "appmake +zx" will give options for the zx flavour specifically.

How it can be used directly can also be seen in [pietro bros](https://bitbucket.org/CmGonzalez/pietro_bros/src/4839eb9d5a85b0f8fd9a72932b62c08723a40576/zcompile.bat?at=master&fileviewer=file-view-default).  That's a windows batch file but you can back up and look at a makefile instead if you prefer that.  In that one, appmake is used to create all separately loaded bits of the game and the final tap is combined with copy or cat.

> Can I use the assembler to make a .tap image, like pasmo does? I know you can do this for C, but I don't think zcc understands asm files at present, right?

You can call the chain or z80asm directly and stop at binary output and then create a tap yourself with separate invocation of appmake.

But zcc can do it all.. it understands pretty much all source types: .c, .o, .asm, .m4 (m4 macros),.. the list is [here](https://github.com/z88dk/z88dk/blob/master/src/zcc/zcc.c#L1664).  So you can go straight to tap for a pure asm project or a mixed asm/c project too.  The only complication arises from the fact that z88dk uses z80asm as a linking assembler, that is it defines sections in order to be able to place code and data into roms and into bankswitched memory.  Assemblers like pasmo are simpler in that there is no linking capability and no memory map definable so that it knows where things should go (multiple org has many pitfalls).

When using zcc for asm files, the asm is ideally written with all code and data assigned to sections so that the target's default memory map can be used.  This way the asm can call into the assembly language library (if desired) without incident, you can specify options to the toolchain per normal, code and data will automatically be placed so that roms can be generated and/or bankswitched binaries can be produced, and the toolchain will operate just like it does for C.

This:

```
SECTION code_user
PUBLIC _main

; called by the crt as entry to program
_main:
   ld hl,10
   ret

SECTION bss_user
PUBLIC myvar

; zeroed by crt when program started
myvar:
   defs 100

SECTION data_user
PUBLIC name

; initialized in ram by crt if program is in rom
name: 
   defm "John Smith"
   defb 0

SECTION smc_user
PUBLIC oops

; self modifying code moved to ram if program is in rom
oops:
   ld hl,0
   ld (oops),hl
   ret

SECTION rodata_user
PUBLIC placedinrom

; kept in rom if program is in rom
placedinrom:
   defb 1,2,3,4

SECTION BANK_06
PUBLIC zxbank6

; varies by target.
; for zx in new c lib in current build, placed in bank6 of 128k spectrum
; classic lib does not define 128 banks (yet)
aymusic:
   BINARY "rawdata.bin"

SECTION code_crt_init
EXTERN stufftodo

; wedged before call to main inside crt so initialization
; can happen before program is started

call stufftodo

; notice NO RET!
```

can be turned into a tap automatically like this:

(classic) zcc +zx -v test.asm -lndos -create-app
(new ) zcc +zx -v -a -startup=31 -clib=new test.asm -create-app

BANK_06 will not be included in the tap (or any other user defined section) as we've not figured out what to do with the 128 banks yet. #186 

The classic crt will define a lightweight stdio which is why -lndos is there.  The new clib is being told to use the [crt specified by startup=31](https://github.com/z88dk/z88dk/blob/master/libsrc/_DEVELOPMENT/target/zx/startup/zx_crt_31.asm.m4) which is a minimal one with no stdio attached.  The options in that crt are controlled by pragmas.

There are other ways to use zcc without using a crt but this is how it should be done because the crts do a lot to set up the environment and make sure things go smoothly.  If you want to look into it, "zcc +zx --no-crt" will make sure zcc does not use one of the crts and it will instead treat the first file on the compile line as the crt.  This means this first file will be responsible for setting up the memory map (if any) and so on.

src/zcc/zcc.c
```diff
@@ -2741,7 +2741,7 @@ int 	find_zcc_config_fileFile(const char *program, char *arg, int gc, char *buf,
 		*/
 		return (gc);
 	} 
-	// Without a config file, we should just exit
+	// Without a config file, we should just print usage and then exit
```
vext01: I mean, should we `fprintf(stderr, "No config file!\n");` before printing usage?

Makefile
suborb: I've switched it use install now, with a bit of find magic beforehand.

I've removed all the -g as well now (they were there mostly for my convenience I think!)

src/zcc/zcc.c
suborb: Think that's a detabbing problem, I'm not sure I've sanitised zcc yet.

vext01:
Thanks so much. I really appreciate this.

I wonder if it's worth using the text you wrote on sections/linkage above to make some documentation?

As soon as the next version of z88dk is out, I will submit a port to OpenBSD :)

suborb:
No worries, it was broken and needed fixing and besides, I'm sure I used my old sparc64/OpenBSD box to tease out a few issues in the past.

I've just realised I'm not installing the new library, but I'll sort that out before release.

aralbrec:
> I wonder if it's worth using the text you wrote on sections/linkage above to make some documentation?

I will try to write some overview documentation again soon.  The problem in z88dk is the amount of things not written down is daunting -- you could probably fill a large book to describe how to use things and what is in it.

aralbrec:
A very brief overview of the z88dk toolset:
https://github.com/z88dk/z88dk/blob/master/doc/overview.md

Issue #207, opened 2017-04-28
sccz80: Implement __critical


suborb:
Following on from #203 for zsdcc, we should support the same within sdcc. Both at a statement level and a function level.

suborb:
This will need library support for z80 and z180.

```
l_push_di
l_pop_ei
```

`l_push_di` should di and save onto the stack the state of the interrupt. `l_pop_ei` should pop the state and enable interrupts as necessary.

`l_pop_ei` needs to preserve the return value (i.e. `hide`/`FA`)

Rabbit support is embedded in the compiler by using `ipset 3` and `ipres`

Issue #205, opened 2017-04-28
z88dk: improve access to block i/o instructions


aralbrec:
At the moment the new c lib contains the following sort of functions to access the zilog family's block i/o.  The classic lib has something similar.
~~~
extern uint8_t z180_inp(uint16_t port);
extern void *z180_inir(void *dst,uint8_t port,uint8_t num);
extern void *z180_indr(void *dst,uint8_t port,uint8_t num);
extern void z180_outp(uint16_t port,uint8_t data);
extern void *z180_otir(void *src,uint8_t port,uint8_t num);
extern void *z180_otdr(void *src,uint8_t port,uint8_t num);
extern void *z180_otimr(void *dst,uint8_t port,uint8_t num);
extern void *z180_otdmr(void *src,uint8_t port,uint8_t num);
~~~
The implementations themselves are very brief and simple.  For example, "z180_otdmr" is just "otdmr; ret".

These things can be inlined in post-processing either by copt or sdcc peephole rule.

aralbrec:
Maybe we should add regular block memory instructions too (ldir, lddr, cpir, cpdr)?  The builtins kind of do this already for memcpy, memset and so on.  This would be a second avenue to inline these instructions but not quite as good if the function call overhead can't be optimized away.

suborb:
For reference, in order for sccz80 to inline a call the following conditions must be met:

* memset(b,c,len) - at least len needs to be const
* memcpy(dst,src,len) - at least len needs to be const
* strcpy(dst,src) - always inlined
* strchr(ptr, c) - c must be const

No stack will be used for strcpy/strchr and at most one push/pop for memset/memcpy

PR #219, opened 2017-04-30
Implement __critical for sccz80

suborb:
This addresses #207 

Missing lib support, but will pick up Alvins work on #205 when that's complete.

Merge after #218 has been merged


suborb:
Wrong way round..

PR #220, opened 2017-04-30
Implement __critical for sccz80

suborb:
This addresses #207

Missing lib support, but will pick up Alvins work on #205 when that's complete.

Merge after #218 has been merged


aralbrec:
You can use the existing library routines in the new c lib for this.

[asm_z80_push_di](https://github.com/z88dk/z88dk/blob/master/libsrc/_DEVELOPMENT/z80/z80/asm_z80_push_di.asm) pushes the interrupt state onto the stack and [asm_z80_pop_ei](https://github.com/z88dk/z88dk/blob/master/libsrc/_DEVELOPMENT/z80/z80/asm_z80_pop_ei.asm) restores the interrupt state from the stack.  For the classic lib, choose the nmos version as it will work for all z80s.

I've started giving these additional names with "cpu" replacing "z80" because the code may have to be different for other z80 relatives.  sdcc will use three entry points to these functions as you will see in https://github.com/z88dk/z88dk/tree/master/libsrc/_DEVELOPMENT/l/sdcc .

When this is merged, I'll point the new sccz80 primitives toward these functions as is done for sdcc.

suborb:
For the z88, we need to call OZ_di and OZ_ei to get and set the appropriate status. For sccz80 we can have this copt rule:

```
        call    l_push_di
=
        call    $51
        push    af

        call    l_pop_ei
=
        pop     af
        call    $54
```

PR #224, opened 2017-05-02
Rearrange lib

suborb:
Almost done with make install

Rearrange lib a bit - creating arch directory to separate out def files.

Create a place for crt0 within arch file


suborb:
This moves the lib folder into separate arch folder (maybe should be target?) to disentangle things - it's confusing to figure out what file belongs to what port.

I'm going to cherrypick the build changes into another pr since they are fairly uncontentious.

suborb:
I'm gonna park this one for a while, I'm not certain it's totally the way to go.

PR #245, opened 2017-06-11
Getting started with zsdcc and ZX Spectrum

derekfountain:
First, this is the first time I've used github, so if I've done the pull request thing wrong, someone point me in the right direction. :)

Second, this is the first time I've offered anything to Z88DK, so if this isn't the way you do it, again, someone point for me.

Further to my pleas for help on WoS (where I'm "thricenightly") I wrote up the notes I made regarding getting zsdcc compiling a simple newlib Spectrum program. I wasn't sure if it should go in the wiki or in the docs directory, or quite what to do with it. Since it only applies to Spectrum development, and none of the other Z88DK targets, I wasn't actually sure if you'd want it. But hey, you can always say no. :)

It's a single markdown document which attempts to distil the compilation process without all the complex detail found in other Z88DK documentation, and which all rather confused me. I've put a 01 in the filename because I've already started the next instalment, 02. But I've no idea where that's going yet.

### DIFF
--- a/doc/ZXSpectrumZSDCC_01_GettingStarted.md
+++ b/doc/ZXSpectrumZSDCC_01_GettingStarted.md
@@ -1,0 +1,255 @@
+# ZX Spectrum Development with Z88dk - Getting Started
+
+This document describes how to get started writing ZX Spectrum programs using
+Z88DK. It concerns itself only with the newer, more standards compilant zsdcc C
+compiler. The older sccz80 compiler is not discussed.
+
+Further, this document only concerns itself with the 48K ZX Spectrum. Topics
+which came with the later machines, like memory bank switching, are not covered
+here.
+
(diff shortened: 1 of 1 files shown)


aralbrec:
I like this a lot - I think it's something ideal for a newcomer and is very clearly written.

Ideally I think this should be part of the wiki but the wiki's not ready for contributions yet so I am ok with giving it at least a temporary home in z88dk/docs.  With recent activity around the Next, there's been an up-tick in questions and people trying z88dk out and I think many people give up not understanding how things work.

Maybe we can extend the name to "ZXSpectrumZSDCCnewlib_01_GettingStarted.md" so it's clear from the title what compiler/library combination is discussed?  I'll do that and if it turns out the name is too long in listings it can be changed.

PR #249, opened 2017-06-16
Second offering for a getting started guide

derekfountain:
As before, I'm new to this. Be gentle! :)

This is another getting started sort of document, following on from the one I submitted last week. It covers the stdio library and related crts. It's ZSDCC/newlib only, and is a writeup of my notes, rather than an exhaustive guide.


derekfountain:
Hmmm, this pull request appears to list the commits I made to the previous document which has already been merged. I'm not sure I've done this right.

I shall blunder on and perhaps read the github docs a bit more. :)

aralbrec:
Thanks Derek, very readable.

derekfountain:
Did I do this pull request correctly? I've been reading the Github docs trying to work out why I can see commits for my previous document, all of which were merged a few days ago. I sort of assumed they'd all be "wiped off" in some way and I'd only see changes in the pull request which I'd made since I synced with upstream.

aralbrec:
I'm not a github expert either but this merge only contained one changed file so even though the commit comments are there for the other file, there are no associated changes.

I made some minor edits just now, mostly text errors but also a couple of small changes too:

* the brief description of drivers in part 2.  I thought mentioning a byte stream was more accurate language and simple enough so that no further explanation is necessary.

* the hexdump program I changed to use "unsigned char" instead of "int".  Hopefully that's a subliminal plant in the minds of people to prefer smaller unsigned types if possible.

* I added forward and backward links.  Actually if more parts come, maybe a table of contents should be on the front page.  Did you want to add your name someplace?

There's no mention of crt8/9 for proportional fonts in your summary at the end of part 2, not sure if that's intentional.  Also, if the startup values have 32 added to them, an if2 cartridge will be made if "-subtype=if2" is added to the compile line.  The +32 startup value corresponds to the same crt without 32 added but the memory map is changed for if2 cartridges.  The subtype option convinces appmake to make a 16k rom instead of a tap file.  I don't know if that has any place here.

There are several pragmas relevant at this point but these are a more advanced topic and maybe don't belong at the introductory stage.  But two that may be worth mentioning anyway are "#pragma printf = ..." and "#pragma scanf = ...".  These will control exactly which printf/scanf converters will be enabled for the compile.

For example, the rom hex dump program only uses %X and %c so:

#pragma printf = "%X %c"

at the top of the source file would reduce the program size by only enabling those converters.  In these enablers, case and length modifiers are important so:

#pragma printf = "%X %x %lx %llX"

is enabling different converters even though internally they are sharing the same code.

derekfountain:
OK, I'm going to try to clean up my fork of the z88dk repo. If I end up giving you a pull request for a pile of crap just hit reject, or whatever you have to do. :) github is _hard_! I might just re-clone, it'll be quicker.

I did look at the proportional fonts, but considered that to be something a newbie might venture into once they've got going, like I did. pragmas I noted, but for now memory consumption issues aren't my concern and pragmas seemed to be complex detail I can leave for another day. The if2 cartridge stuff isn't something I've looked at as yet.

I'd say that the bits I've left out of these guides are as considered as the bits I've put in. There's lots of docs about z88dk, and an awful lot of posts about it at WoS, etc., but from a newbie's point of view the overload of fine details makes life much harder, sometimes to the point where it's impenetrable. As a beginner I've had a lot of trouble working out what's relevant to get going and what's detail that might be of interest a long way down the line.

It's a judgement call each time, but that said, if you do think I've missed something that a newbie really should know about by all means point it out. :)

aralbrec:
I do agree with you - too much information can make things more complicated.  I think the crt8 and crt9s should at least be mentioned in your summary so that people know they exist.  The if2 cartridge thing could maybe be documented elsewhere.  I think a companion "advanced notes" produced at the end could maybe touch on these other things.

PR #250, opened 2017-06-17
zsdcc guide, part 3

derekfountain:
I tried building this pull request using a github branch and merge on my fork repo. The improvements to my process might be... noticeable? :)

Not sure what you'll make of this one. The plan was, after hello world and text IO, to go to simple graphics - draw, plot, etc. But as I found out, those routines haven't been ported to newlib as yet, so it can't be done. Thus what I've ended up with is a stub. You may decide a stub is unhelpful.

Thing is, it would have helped me. It still hadn't gelled in my head why I could see a header file full of declarations for the functions I wanted, but the compiler wouldn't find the header or build any program which used them. I worked it out in the end, and a post on the user's list confirmed it, but there was a period of bewilderment in there which could have been avoided if someone had spelled out for me the relationship between classic lib, new lib, and the Z88DK information found on the interwebs. It might also be useful to have an 03 stub document in there ready to be fleshed out when the graphics functions are ported over to newlib. Your call.

Next up I was going to look at the lower level input stuff in input_zx.h. :)

### DIFF
--- a/doc/ZXSpectrumZSDCCnewlib_02_HelloWorld.md
+++ b/doc/ZXSpectrumZSDCCnewlib_02_HelloWorld.md
@@ -318,3 +318,5 @@ don't use one of the crts which enable them. You'll save memory.
 #### -startup=31 (crt31)
 
 stdio not supported at all
+
+[... continue to Part 3: Simple Graphics](https://github.com/z88dk/z88dk/blob/master/doc/ZXSpectrumZSDCCnewlib_03_SimpleGraphics.md)
--- a/doc/ZXSpectrumZSDCCnewlib_03_SimpleGraphics.md
+++ b/doc/ZXSpectrumZSDCCnewlib_03_SimpleGraphics.md
@@ -1,0 +1,57 @@
+# ZX Spectrum Development with Z88DK - Simple Graphics
+
+This is the third document in the series which describes how to get started
+writing ZX Spectrum programs using Z88DK. As before, it concerns itself only
+with the newer, more standards compilant zsdcc C compiler. Neither the older
+sccz80 compiler nor the older "classic" libraries are covered.
+
+## Assumptions
+
+It is assumed the reader has worked through the earlier installments of this
+series and is continuing on from
+[installment 2](https://github.com/z88dk/z88dk/blob/master/doc/ZXSpectrumZSDCCnewlib_02_HelloWorld.md).
+
+## Simple Graphics
+
+This guide examines the Z88DK capabilities for handling low level graphics
+primitives like drawing lines and plotting points. But prepare for a
+disappointment.
+
+There's a header file full of graphics routines here:
+
+```
+ include/graphics.h
+```
+
+[This](https://github.com/z88dk/z88dk/blob/master/include/graphics.h) is what
+we're after: draw, plot, circle, all sorts of stuff. But there's a problem. The
+header file is in the Z88DK include/ directory, whereas the header files for all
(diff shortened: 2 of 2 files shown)


aralbrec:
If it helps to spell out the difference between classic and newlib it's still helpful.  There are lower level functions to help with screen address manipulations that can make it easy to do many things so maybe a small introduction of that would be helpful.  There are a couple of factual errors in there too... you can, in fact, use zsdcc with the classic library for example, so I will edit a bit in there to clear things up.  I'll add a short bit on the screen address manipulators and maybe include a line drawing routine from the internet to show how they can be useful.  Have a look at this later and edit where you see necessary.

For the next part, input.h is the correct header to use - it will include a target-specific header depending on which machine is being targeted so when the zx is target it will also include input_zx.h.

aralbrec:
Ok, I've made the changes:
https://github.com/z88dk/z88dk/blob/master/doc/ZXSpectrumZSDCCnewlib_03_SimpleGraphics.md
Edit as you see fit.

derekfountain:
I'm in two minds about this update.

The point of my efforts, from a personal perspective, was to learn enough about Z88DK in order that I might partake in the recently announced WoS game compo. It turns out that the Z88DK learning curve is remarkably steep, way steeper than it needs to be for someone well versed in C like myself. I concluded that it's that steep because there are lots of supported machines, 2 compilers, 2 libraries, and documentation which has grown organically over the years into something is which carries the information I require but has become egregiously tangled. My notebook reflects my efforts of untangle the documentation and to understand how I get from my bewildered starting point to the sprite library.

My ignorance is actually a benefit to this process. I'm deliberately trying not a learn anything without documenting if first. I haven't thought about proportional fonts or implementing algorithms like line drawing as yet. I was looking at the input library yesterday and worked out the basics, but haven't sorted out the purpose of scancodes yet. That's today's work, if I get time.

So I feel your addition is a big leap ahead of where I am and where the series is, an entanglement of keyboard reading and algorithm implementation information in a piece which spelled out what I thought was an important lesson: there are lots of stools in the Z88DK world, and falling between them is a pitfall to be avoided.

Having had chance to reflect as I've typed this, I think the best option might be to remove those keyboard reading functions and simplify those examples to just do 10 plots or lines. Then I can remove the forward reference to the input library. The explanation of the screen addressing functions is a borderline call, since I was hoping for a "call this library function to draw a line" sort of level. But that's not possible, and the lesson that it _is_ possible to solve these problems if you're prepared to get your coding fingers a bit dirty is a useful one at this stage, so it's probably worth keeping that information where it is. I might put the line drawing algorithm behind a "you don't have to understand this, just be aware you can solve problems by coding at this sort of level" statement. I wouldn't want a reader thinking they have to understand Bresenham in order to progress with learning to use Z88DK.

I think those changes will keep the flow through the documents reasonably linear and steadily educational. I'll finish piece 04 first though, since my head's wrapped up in that at the moment.

aralbrec:
Don't feel guilty about changing or deleting anything I might add.   I have a tendency to go a step too far and that can add to confusion.   It's your series after all :)  Mainly I'm making suggestions about things that I think you yourself may not be aware of and might fit in in the topic you're discussing.

Getting rid of the input is probably a good idea in retrospect to avoid forward references.  I do think this is a good spot to at least mention the display file manipulators because it will be something that people will need to use for, eg, text printing without stdio.  And the impression left without this section is "we're screwed on graphics", which is partly true, but all is not lost.

PR #251, opened 2017-06-19
rc2014 - fast ACIA ringbuffers

feilipu:
Modify the ACIA ring buffers to use the `section bss_align_256` capability for BSS alignment.
BOTH Rx and Tx buffers must be on the `0xnn00` boundary, as only the low byte of `l` of `hl` is incremented.

Not 100% certain whether these need to be declared as `0x0100` or `0x00FF +1`. But the alignment should take care of ensuring there's space assigned, even with the current `0x00FF` assignment.

Tx buffer is reduced to just 16 bytes, because the rc2014 doesn't need any more.

Tweaked the interrupt to check that it can transmit (`ACIA_TDRE` is set) before checking whether it has anything to send.

Added `ACIA_RX_EMPTYSIZE` for no particular reason.


aralbrec:
The size has to be 256 bytes and 16 bytes respectively.  Maybe the correct thing is to declare those sizes + 1 in the defs and leave the values as you had them.  I'll undo and do that,

feilipu:
I don't really mind. Either way works for me.

But the key code issue is that the sizes are used as `cp` and `and` masks, so they need to result in `0xFF` and `0x0F` respectively.

In hindsight it is probably clearer to declare them as `2^n` values, and then `-1` for the actual comparison and masking.

aralbrec:
I personally understand it better if the size constants accurately reflect the sizes.  Are you ok to keep it that way?  We just have to review that the -1s are applied properly.

feilipu:
Agreed.

For _Rx_ buffer, there is a `-1` needed for checking buffer fullness. Buffer rolls over automatically using `inc l` on the `hl` pair.

For _Tx_ buffer, there is a `and 2^n-1` for both the interrupt and the _Tx_ function.

aralbrec:
As you say the Tx buffer has to be in the 256-byte aligned section.  With a 16-byte buffer going in there, there will have to be 250 bytes of padding otherwise something else assigned to that section will not be 256-aligned.  I wonder if there is something that can be done to keep it in the 16-byte aligned section.

feilipu:
To be honest, changing the Tx buffer handling is a third order issue. The rc2014 is just too slow to fill more than a few bytes. I'd suggest to leave the buffer mgt generic as it is today.

That means just undo the two code pieces incrementing the Tx buffer pointers, and then range checking.

That's the easiest outcome. I think. 

For Rx it makes a huge difference to use the whole low byte with no range checking. Also can't control Rx rate so it stops overflow for HEXLOADR.

For z180 based YAZ180 it's another story, and I use 256 bytes for Tx too.

aralbrec:
This bit takes advantage of the 256-alignment:

```
        inc l                       ; move the Tx pointer, just low byte along
        ld a, ACIA_TX_SIZE - 1      ; load the buffer size, (n^2)-1
        and l                       ; range check
        ld l, a                     ; return the low byte to l
```

which is 19 cycles.

A change to this:

```
   ld a,l
   cp (txbase & 0xff) + ACIA_TX_SIZE - 1
   jr z, backtostart
   inc l
rejoin:
.....
.....
backtostart:
   ld l,txbase & 0xff
   jp rejoin
```

is 22 + 22/16 = 23.4 cycles.

So it's only a 4-5 cycle loss to avoid the padding.

I'd go for something like this to save 250 bytes if, as you say, the rc2014 is too slow to fill the buffer anyway.  I think this is what you mean by range checking?

feilipu:
Yes.
Thanks.

aralbrec:
Thanks again feilipu.  I think it's all correct now so I will commit.
Let me know if something doesn't work in your own tests and I will review this again tomorrow to make sure I didn't miss something.

feilipu:
Ok. Thanks.
Might not get to full testing tonight (other duties), but tomorrow.
Will update this comment with outcome.

PR #255, opened 2017-06-21
Slightly faster, smaller zx_py2saddr

gp48k:
Instead of setting the upper address bit using "or $40" I do that with a carefully placed "inc a" while shifting other address bits into place.

This saves 1 byte and 3 T-States.


aralbrec:
Looks correct, thanks.

PR #261, opened 2017-06-23
Tweaked graphics doc and new inputs doc

derekfountain:
More updates to my documentation.

I've updated the graphics doc you added to. Just some simplification of the examples to remove the forward references to the input library, and a bit of cleaning up of the text, making it match my other documents a bit better, and removing a few of the more technical concepts which I thought crept in unnecessarily.

I then added the inputs doc. There's more to this topic than one might think so knowing where to stop and let the reader take over wasn't obvious. Given that I originally wanted very basic game control for the keyboard and joystick, that's where I took it to.

I think the next one might look at compiling multiple source files and mixing assembly language in. After that I probably won't be able to put off looking at sp1 any longer. :)


aralbrec:
There is a subtle problem with this example code:

```
    in_wait_key();
    c = in_inkey();
    in_wait_nokey();
```

in_inkey() can in fact recognize shifted keypresses.  However, if you press "CAPS+a" (for example), normally you press the CAPS key first.  This is registered by in_wait_key() as a keypress, `c = in_inkey()` is run getting a result of 0 (CAPS doesn't count as an ascii code), then in_wait_nokey() is run.  CAPS holds the program there, then the user presses a and releases keys.  So you see ? mark printed and an ascii code of 0.  Your mention that in_inkey() can't register shifted keys is result of this subtle behaviour.

A fix would be:

```
  while( 1 )
  {
    while ((c = in_inkey()) == 0) ;

    printf("Key pressed is %c (0x%02X)\n", c, c);
    in_wait_nokey();
  }
```

where in_inkey spins until it registers a single ascii code.  in_wait_nokey is postponed until after the printf so there is immediate feedback.

The other thing is not everyone has kempston joystick capability.  I don't and I was hunting through my emulator (SpecEmu) to see if I could hook up cursor keys or something but it can't!  Most emulators can I think.  I think maybe adding a suggestion to use in_stick_sinclair1() as a substitute would be ok.  SInclair sticks map to the number keys so they will work with keyboards only.

I'll commit, make these two changes, let you know when I'm done and then you can review again.

aralbrec:
ok edits are in

aralbrec:
> I think the next one might look at compiling multiple source files and mixing assembly language in. After that I probably won't be able to put off looking at sp1 any longer. :)

Yep :)  If you are interested in them there are also beeper audio functions (simple beeps, a play string, sound effects, the tritone engine), direct fzx functions (proportional fonts), im 2 interrupts.

Then more advanced things like configuring the crt, setting the memory map, putting stuff into memory banks.

PR #262, opened 2017-06-24
inputs - another update

derekfountain:
All the changes you made to the inputs doc are fine except the in_key example. It's gone from clear and immediately understandable (I found it in the blackstar code where it was easy to spot) to being non-obvious due to what is, at this getting-started level, a technicality.

I think it's better to go back to the code as I had it and explain in the text that the in_inkey() function returns 0 under some circumstances. The reader can then experiment with key presses and see the 0s coming back. Although in a real program spinning to ignore the 0s might be the thing to do, I think for this guide showing the exact responses from the function improves the explanation of what's going on. The guide is supposed to be promoting understanding, not just providing examples to learn by rote.

I also thought that putting the printf() to show the result inside the in_wait_nokey(), although technically more correct for immediate feedback, broke up the point of the example which is wait for a key to go down, read it, wait for it to go back up, which is just what the backstar code does. So I kept the printf() outside, where it's less in the way.

I prefer it as it is now. If you have strong feelings, or think something is technically wrong, we can go round again. :)

BTW my pull requests are building up with lists of commits you've already merged. I still don't know why. If they get in your way let me know and I'll try to work out why it's happening.


aralbrec:
The main issue I have is that this test program will give the impression that in_inkey() cannot detect shifted keys like 'A', '$', "@" etc when in fact it can detect any ascii code that can be generated on the keyboard.

The example program has a problem because the moment a shift key is pressed, in_wait_key() will end and in_inkey() will read the state of the keyboard which will only be a shift key (thus returning 0).  Then the person, still holding shift, types another key, maybe trying to generate "A".  The screen is still blank as the program is held up on in_wait_nokey() and when he releases caps+a, a 0 gets printed.  So his impression is that caps+a is undetectable by in_inkey().

The immediate feedback would also help in clarifying that the output is generated by the CAPS keypress and not the CAPS+delayed A keypress?

If it's more clearly explained that these limitations are due to the structure of the example program then I think it's ok.

aralbrec:
The pull is ok here, no worries :)

derekfountain:
Yes, I do understand the problem. I think another example is required. The first one, which is currently in place, describes the typical case for defining a game key (like blackstar). To complete the picture and describe the CAPS interaction, a second example is required. I can add that.

But, correct me if I'm wrong, even checking for a 0 return then looping, there's no way to tell the difference between someone entering CAPS then 'a' and SYM then 'a'? The shift keys both return 0. In fact, the loop-while-zero example you provided wouldn't actually detect a CAPS-then-'a' combination. The loop just ignored the return code while it was 0 and would eventually return the 'a'.

If I've got that right I can explain the issue with another example.

aralbrec:
> But, correct me if I'm wrong, even checking for a 0 return then looping, there's no way to tell the difference between someone entering CAPS then 'a' and SYM then 'a'? The shift keys both return 0. In fact, the loop-while-zero example you provided wouldn't actually detect a CAPS-then-'a' combination. The loop just ignored the return code while it was 0 and would eventually return the 'a'.

No it will work properly if in_inkey() is in a spinning loop.  The shifts on their own are treated as no key pressed but when any other key is pressed, it is modified by the shift keys.  So a, caps+a, sym+a and caps+sym+a will all generate different ascii codes.

What Black Star is doing lacks a little responsiveness and will not always work (it is possible to hit caps+a at exactly the same time so that in_inkey() returns 'A' and if you use that as a scancode in your keyboard joystick, the user will have to press capital A to activate the direction.  Personally I would spin on in_inkey returning 0, convert that to a scancode, then strip off the two shift bits.  Or spin on inkey() returning 0 or a code outside [a-z0-9].

PR #276, opened 2017-07-04
yaz180 - 1st commit

feilipu:
This is a 1st commit for a new z180 based platform, called [YAZ180](https://feilipu.me/2017/02/26/yet-another-z180-v2/) (Yet Another...).

I think I've covered most issues, but the [configuration](https://github.com/z88dk/z88dk/compare/master...feilipu:master#diff-17c93d69a1dcf74ac515dddc7cba0e1a) stanzas are still a little bit of a mystery, and need to be properly resolved. I've tried to obtain a similar configuration result to the rc2014 configuration in the first instance.

The configuration for the z180 vector address tables is not completed, therefore the [`_asci0_interrupt`](https://github.com/z88dk/z88dk/pull/276/commits/08fc35e48db7b1b76eaae1499d4df024973a102c#diff-3700fd03273caba7358f29392b7b4164R13) interrupt is not connected. This is addressed in #277.

The YAZ180 can be either an embedded device with 256kB Flash, and 1MB of RAM with a parallel FTDI interface for direct programming, or it can be configured as a FUZIX, UZI180, or CP/M machine with an IDE hard drive (through a 8255 PIO).

It includes support for
- Am9511 APU for hardware 32bit floating point.
- Intel 8255 PIO for 44pin IDE, and general IO. TIL311s on Port B.
- PCA9665 I2C (x2) for video and other I2C peripherals.
- FTDI FT245 parallel I/O for programming.
- FTDI FT232 serial for ASCI0.

The hardware is configured by programming GAL16V8 devices. The [memory map and BOM](https://feilipu.me/2016/05/23/another-z80-project/) are documented.

There's only one of these boards now, but the intention is to make more (and to encourage others make more if they wish).

p1080760


aralbrec:
Let's stick it in since and fix it up later since there's no conflicts.

aralbrec:
Are you also considering display and internet options?

feilipu:
> Are you also considering display and internet options?

For the __display__ I have a concrete solution in mind. There are many options, but the one that I've [personally the most experience](https://feilipu.me/2014/01/31/gameduino-2-with-goldilocks-and-eve/) with is the [FTDI EVE](http://www.ftdichip.com/EVE.htm) family. Specifically from inspection, the [4D FT483 ](http://www.4dsystems.com.au/product/4DLCD_FT843/)looks like it can be easily converted to I2C (from SPI). There's an [App Note specific to I2C](http://www.ftdichip.com/Support/Documents/AppNotes/AN_308%20FT800%20Example%20with%208-bit%20MCU%20(I2C%20Interface).pdf) interfacing to EVE, and there will be other hardware options available. I haven't looked in detail yet.

I've put two I2C controllers on the YAZ180, to allow for the fact that one I2C interface might need to be running _high speed_ for display purposes, but the other can be run _standard mode_ over long distances for sensor applications.

The key advantage is that the EVE device generates all the video timing and the CPU just has to generate a "display list", essentially a high level script, to update the output. This doesn't have any specific timing constraints, as the "display list" is double buffered.

For __Internet__, I was (am) quite keen on integrating a [Wiznet W5300](http://www.wiznet.io/product-item/w5300/) device. It is perfectly matched to the Z80 platform. But as a 100 pin VLSI device, I was not certain my construction skills stretched to implementing it in the first board I've laid out. In hindsight, it was the right decision.

Therefore, I don't have a ready answer for Internet at this stage.

I've had [personal experience](https://feilipu.me/?s=Wiznet+W5500) with Wiznet W5x00, and I find their hardware implementation of TCP/IP a very effective way of working with small micro-controllers. Off loading the timing related issues and having external buffers is a great way of getting TCP/IP on 8 bit machines.

It is even possible to abandon their IP and MAC code, and [use uIP or similar](https://feilipu.me/2013/09/19/uip-on-wiznet-w5200-versus-w5100-on-goldilocks-1284p/) in MACRAW mode, if IPv6 is a mandatory requirement.

But, perhaps there are connectivity options via the 82C55 interface?
Or, via the ASCI1 interface, that is pinned out to a FTDI 6-pin connector (but not used normally).

Issue #277, opened 2017-07-05
Relocatable jump / vector tables for z80 & z180 ROM memory models - question / enhancement


feilipu:
Extending from #50 and #276

Looking at the memory model for the [z80](https://github.com/z88dk/z88dk/blob/master/libsrc/_DEVELOPMENT/target/crt_memory_model_z80.inc) and [z180](https://github.com/z88dk/z88dk/blob/master/libsrc/_DEVELOPMENT/target/crt_memory_model_z180.inc), both seem to assume that the jump destination for the `RSTnn` instructions is fixed into the ROM of the device, for the ROM model.

If the z80 [`__crt_org_vector_table`](https://github.com/z88dk/z88dk/blob/master/libsrc/_DEVELOPMENT/target/crt_memory_model_z80.inc#L135) is set positive, then the jump table location could be positioned in RAM, but there seems to be no mechanism to fill the jump table with the relevant vectors during the initialisation.

Also, the connection between the individual (eg [`RST08`](https://github.com/z88dk/z88dk/blob/master/libsrc/_DEVELOPMENT/target/crt_page_zero_z80.inc#L14)) jumps and the allowed jump table is not apparent. Is this jump table for the `RSTnn` instructions? Or does it have another purpose?

If the z180 [`__crt_org_vector_table`](https://github.com/z88dk/z88dk/blob/master/libsrc/_DEVELOPMENT/target/crt_memory_model_z80.inc#L138) is set positive, then an address vector table can be positioned in RAM, but initialisation of this table is [left to the user](https://github.com/z88dk/z88dk/issues/50#issuecomment-281574192). That is ok, but otherwise there is no mechanism to initialise them at boot.

__Enhancement Request__

Would it be possible to allow the z80/z180 RST vector jump table to be [prototyped](https://github.com/feilipu/NASCOM_BASIC_4.7/blob/master/includes/z80intr.asm#L80) within the `SECTION DATA` and have the [vectors initialised](https://github.com/feilipu/NASCOM_BASIC_4.7/blob/master/sbc_NascomBasic32k_HexLoadr/int32k.asm#L319) into RAM during the normal crt initialised data process?

This would allow z80 ROM builds to
* reassign the `RSTnn` vectors during run time,
* allow sharing (reassignment) of the `RST38` `INT0` and `NMI` vectors if hardware could share the signal.
* update systems if bugs are found post shipment of ROMs.

[...]

PR #278, opened 2017-07-06
RAM based RST vector jump table

feilipu:
This provides the option to build a RAM based `RSTnn` and `NMI` steerable vector jump table for ROM based z80 machines, as a partial resolution for #277. A separate pull request for z180 will be provided later.

As the `RSTnn` and `NMI` jump destinations are normally embedded into the ROM of the device, it is normally not possible to steer these to new destinations after the ROM is burned. If a jump table is installed in RAM, then the `RSTnn` instructions will go via this table, which can be reconfigured at run time.

This option is enabled by `__crt_org_vector_table = -2`, and only the `RSTnn` vectors enabled by `__crt_enable_rst` will be fed to the vector table. It could have been `-1`, but that option was taken.

An additional option `__crt_enable_rst_table` takes on the same function previously held by `__crt_enable_rst`, and the user provided `RST` routines must have an additional leading underscore.

The same applies for the `NMI` vector with `__crt_enable_nmi_table`

For example, below only the `RST38` is enabled, and the user supplied function is `__z80_rst_38h`.

```asm
   defc TAR__crt_org_vector_table      = -2
   defc TAR__crt_interrupt_mode        = -1

   defc TAR__crt_enable_rst            = 0xFE
   defc TAR__crt_enable_nmi            = 1
   defc TAR__crt_enable_rst_table      = 0x80
   defc TAR__crt_enable_nmi_table      = 0
```
The vector table is currently placed in `SECTION smc_driver`.
There may be a more appropriate place for it to live. Please adjust if needed.
```asm
IF (__crt_org_vector_table = -2)
SECTION smc_driver
   include "../crt_interrupt_table_z80.inc"
ENDIF
```
A location for the `RST00` vector is provided for future z180 based Trap support.

Minor adjustments to the RC2014 configuration have been made to test this option.


feilipu:
I forgot to add the defaults to `target/crt_defaults.inc`
```asm
defc DEF__crt_enable_rst_table = 0 ;; if non-zero and in some crts with code org = 0, set bits indicate which rst locations are implemented with user code, with __crt_org_vector_table = -2
defc DEF__crt_enable_nmi_table = 0 ;; if non-zero and in some crts with code org = 0, a jump to user code to service the nmi is inserted, with __crt_org_vector_table = -2
```

aralbrec:
I've been reviewing this and I agree an easy way to make a ram-based jp table for the restarts (which include im1 and nmi ints) is a good idea.

A couple of things:

 * I don't think renaming `_acia_interrupt` to `_z80_rst_38h` is a good idea since you can't undo that name once it's in the library.  If someone doesn't want the acia_interrupt on im1 then there's no way to access that subroutine since a user-supplied `_z80_rst_38h` will hide the library's version.  If a soft-default is needed, it's better to make it appear in a separate file as `defc _z80_rst_38h = _acia_interrupt`.

 * `crt_org_vector_table` is being conflated with the restarts.  `crt_org_vector_table` is supposed to be an address of the im2 vector table.  So using a value of -2 to indicate that a ram-based jump table for restarts should be made means nothing can be done with the im2 vectors.  I think maybe it should be renamed to clarify this.

 * Related to the last point, what about extending the meaning of `crt_enable_rst`.  Right now the bits in the lowest byte indicate which restarts the user will implement.  We could take the next higher byte to indicate if a jump table in ram should be made for corresponding restarts.  That way we could do away with `crt_org_vector_table=-2` (`crt_enable_rst & 0xff00 != 0` means the same thing) and `crt_enable_rst_table` as well since that info is in the second byte.

feilipu:
Thanks for the comments, and they'll help to improve this suggestion.

I think it is important to make this `RST` and `NMI` RAM table work seamlessly with the `IM2` RAM table, and the Z180 internal interrupt RAM table (when it is built). So any integration of configuration, or mechanism, or user experience we can get will hold for a long time.

When building this I didn't want to tug too hard at the z88dk ball of string, because I wasn't sure what threads would become unravelled, and I wouldn't know where to tie them in. So, I've only touched the minimum number of lines to make it work (for me).

> I don't think renaming _acia_interrupt to _z80_rst_38h is a good idea since you can't undo that name once it's in the library. 

Agreed. Nasty hack. But I wasn't sure of where the best location for the `defc` was. And I wanted to remove the special [`rc2014_crt_0.asm.m4`](https://github.com/z88dk/z88dk/pull/278/files#diff-fa8de3eb094fc81e7d9718da09fb4564) file that only had one line difference (the declaration of the ACIA interrupt) to the standard `z80_crt_0.asm.m4` file.

Please improve to let me know where the correct location for this definition is. I guess somewhere in an `m4` file?

> crt_org_vector_table is being conflated with the restarts.

Yes. I didn't want to invent a whole nomenclature, so I just stole one value.
Really there should be a set of options for precisely locating (in either ROM or RAM) of
* `RST` and `NMI` jump tables (optional),
* `IM2` address tables, and
* Z180 (0/L/S) `INT1`, `INT2`, and internal interrupt address tables.

And, the addresses of these should be available to both the ASM and C programmer.

An additional point that I omitted to fix is that the C programmer can't obtain the location of the `RST` or `NMI` address easily, because it is the jump label+1. Specifically `__z80_rst_38h` (double underscore) refers to the location of the `jp` instruction, not the address they need to modify. I don't know a neat way to automatically define an address in the middle of an instruction, (except for the nasty ways of creating yet another label, being the old label+1).

> Related to the last point, what about extending the meaning of `__crt_enable_rst`. 

Good idea. Better than yet another configuration label.

What I don't have a good idea about currently is how to create a similar nomenclature for the other address tables (above) that need to be configured. Ideally the user shouldn't have to understand something too weird.

aralbrec:
> Really there should be a consistent set of options for precisely locating (in either ROM or RAM) of
>    1. RST and NMI jump tables (optional),
>    2. IM2 address tables, and
>    3. Z180 (0/L/S) INT1, INT2, and internal interrupt address tables.
> And, the addresses of these should be available to both the ASM and C programmer.
> What I don't have a good idea about currently is how to create a similar nomenclature for the other address tables (above) that need to be configured. Ideally the user shouldn't have to understand something too weird.

The z80's im2 table (number 2 above) and the z180's interrupt address table (number 3 above) are the same thing.  `crt_org_vector_table` is intended to be the 256-byte aligned address of this table.  A section can be added to the crt that lets the user build the table in an .asm file and then the crt can be told to copy that to the final destination in ram at startup.

This im2 table is not the same as the jp vectors for the restarts :- im2 tables contain 2-byte addresses of isrs whereas the restart table consists of 3-byte jumps.

What I am thinking about is using special values for `crt_org_vector_table`:

 * -0x80 : set I=0, place a table at address 0x80.  This is a partial table (device ids >= 0x80 only) meant to be part of the crt at address 0.
 * -0x100 : set I=1, place table at address 0x100.  This table is meant to be part of the crt at address 0.
 * 0 : no table
 * > 0 : the address in ram of the vector table, initialized by the crt.
 * < 0 : the address in ram of the vector table, not initialized by the crt.

I may do this tonight for the z80 to see how it can work.

> But I wasn't sure of where the best location for the defc was. And I wanted to remove the special rc2014_crt_0.asm.m4 file that only had one line difference (the declaration of the ACIA interrupt) to the standard z80_crt_0.asm.m4 file.  Please improve to let me know where the correct location for this definition is. I guess somewhere in an m4 file?

We can change the config to set CRT_ENABLE_RST = 0x80 to get the acia_interrupt installed by default when building roms (that was the function of the separate rc2014 page zero code) and have _z80_rst_38h = _acia_interrupt in the library to get it as soft default.  It has to be on its own in an .asm file.  I'll try to find a place for it.

aralbrec:
Based on these ideas I committed a few things for creating a jump table for the restarts.

 * `section smc_jump_vectors` was created to hold jump vectors.  More jump vectors can be added by adding to this section from anywhere.
 * [crt_jump_vectors_z80.inc](https://github.com/z88dk/z88dk/blob/master/libsrc/_DEVELOPMENT/target/crt_jump_vectors_z80.inc) and [crt_jump_vectors_z180.inc](https://github.com/z88dk/z88dk/blob/master/libsrc/_DEVELOPMENT/target/crt_jump_vectors_z180.inc) are included into crts and set up the jump vectors for the restarts.  What's different here is it's not a table with an entry for every restart but rather a list of enabled jump vectors only.  The `defc` thing is unavoidable I think :)

You can add im2 jump vectors to this table and have the isr address in the im2 table itself point at the jumps but I think it will be easier to just put the im2 table in ram and modify it directly.

feilipu:
> crt_jump_vectors_z80.inc and crt_jump_vectors_z180.inc are included into crts and set up the jump vectors for the restarts. What's different here is it's not a table with an entry for every restart but rather a list of enabled jump vectors only.

I think it would be good to allow generation of null vector locations, if desired, even though the final destination may be unknown (or undesired) at compile or boot time.

I'm thinking of this for example.

```asm
IF (__crt_enable_rst & 0x0002)
  PUBLIC _z180_rst_08h
  PUBLIC _z180_rst_08h_vector
  IF (__crt_enable_rst & 0x0202)
    defc _z180_rst_08h_vector = _z180_rst_08h + 1
    EXTERN __z180_rst_08h
    _z180_rst_08h:
    jp __z180_rst_08h
  ELSE
    _z180_rst_08h:
    ret    #ei  / nop
    _z180_rst_08h_vector:
    ret    #reti / retn
    ret
ENDIF
```

aralbrec:
We can give weak defaults to __z180_rst_08h and company in the library so that they don't have to be defined at compile time.

I don't want to make an entry unless it's actually used as in `__crt_enable_rst & 0x0202` .  `__crt_enable_rst & 0x0002` may mean the user wants a jump directly to his subroutine rather than going through a jump vector in ram.

I'll put this in now and complete the im2 vector thing so another look can be made.

aralbrec:
Alright so an implementation is committed.  This is related to #277 as well.

The low byte of CRT_ENABLE_RST determines if the user will be implementation one of the restarts 08h through 38h.  The high byte indicates whether that restart should go through a jump vector in ram.

Using rst 08h as an example:

CRT_ENABLE_RST = 0x0000
 * The user does not implement rst 08h.
 * Address 0x08 contains a jump to a ret instruction.
 * `_z80_rst_08h_vector` = 0x09.
If page zero is in ram, the user will still be able to change the rst 08h vector at address 0x09.

CRT_ENABLE_RST = 0x0002
 * The user implements rst 08h but does not want to go through a vector.
 * Address 0x08 contains a jump to user-supplied routine `_z80_rst_08h`.
 * `_z80_rst_08h_vector` = 0x09.
If page zero is in ram, the user will still be able to change the rst 08h vector at address 0x09.

CRT_ENABLE_RST = 0x0202
 * The user implements rst 08h through a jump vector in ram.
 * Address 0x08 contains a jump to `_z80_rst_08h`.
 * The library creates a section `smc_jump_vectors`.
 * In that section, `_z80_rst_08h` jumps to a user-supplied routine `__z80_rst_08h` (extra underscore).
 * The library provides a weak default for `__z80_rst_08h` that jumps to a return instruction.
 * `_z80_rst_08h_vector` is the address of the jp destination at `_z80_rst_08h` so that it can be used to change the rst08h destination at runtime.
 * Section `smc_jump_vectors` can be added to at by the user at anytime but the library will automatically handle the restarts, nmi and z180 trap there.
 * Because the section holding the jump vectors are in a self-modifying code section, they will be stored in rom and copied to ram at startup by the crt.

The above applies to all the restarts except rst 00h, including the im1 interrupt (rst 38h).

The nmi is handled the same way but through the crt variable `CRT_ENABLE_NMI`.  0 corresponds to the first case above, 1 to the second case and anything larger to the third case.  The associated vector address is `_z80_nmi_vector`.

The z180 also has `CRT_ENABLE_TRAP` that works the same way.  0 corresponds to the first case, 1 to the second and anything larger to the third.  The associated vector address is `_z180_trap_vector`.

For the z180 all the names above have z180 substituted for z80.

=====

In addition to this a means to create interrupt vector tables for im2 (z80) and the z180 that is controlled by `CRT_ORG_VECTOR_TABLE`.

If `CRT_ORG_VECTOR_TABLE` = 0 no vector table will be created.

If `CRT_ORG_VECTOR_TABLE` = -0x80, a partial vector table will be created at address 0x80 in the crt, assuming the org of the compile is address 0.  This will have `I`=0 and device ids must begin at 0x80.  This is a bit unusual but I think more common on the z180.

If `CRT_ORG_VECTOR_TABLE` = -0x100, a vector table will be created at address 0x100 in the crt, assuming the org of the compiler is address 0.  This will have `I` = 0.

If `CRT_ORG_VECTOR_TABLE` > 0, the interrupt vector table will be located at that address and `I` = `CRT_ORG_VECTOR_TABLE`/256.  For positive values, the compile will store a copy of the table in rom and then automatically copy this to ram at address `CRT_ORG_VECTOR_TABLE` before main() is called.

If `CRT_ORG_VECTOR_TABLE` < 0, the interrupt vector table will be located at the absolute value of that address and `I` = -`CRT_ORG_VECTOR_TABLE`/256.  The compile will store a copy of the table in rom but the crt will not copy it to ram at startup.

A section `interrupt_vectors` is created to hold interrupt vectors.  Vectors can be created by adding to this section in device id order:

```
SECTION interrupt_vectors

_vector_0:
defw _isr_ctc

_vector_2:
defw _isr_pio

...
```

feilipu:
Playing with the different options and seeing what comes out of `x.lis` and `x.map`.
Looks good. :+1: 

If we (for example) wanted to define some defaults for the `RSTnn` vectors in the table form for RC2014 or other targets, where would these double underscore weak defaults best live?
```asm
defc __z80_rst_08h = rc_00_output_acia

defc __z80_rst_10h = rc_00_input_acia

defc __z80_rst_38h = _acia_interrupt
```
Could not all the double underscore defaults also just be weak defaulted to the single underscore versions, for simplicity?
```asm
PUBLIC __z80_rst_38h
EXTERN _z80_rst_38h

defc __z80_rst_38h = _z80_rst_38h
```

aralbrec:
Yes it looks like it is a mistake to choose a different name for ram vectors with the extra leading underscore.

So now we have the user implementing "_z80_rst_08h" sometimes and when going through jump vectors in ram, he's implementing "__z80_rst_08h".

I'll change it so that the user routine is always "_z80_rst_08h".

Then we get to the soft defaults.  At the moment for the jump table in ram, the library is providing soft defaults.  But the use case here is the rc2014 wants to supply different soft defaults.  We're putting all these soft defaults into the same library so the one that gets chosen by the linker depends on the order they are stored in the library.  It's not a good idea to depend on that.

How about we get rid of the soft defaults in the general portion of the libary entirely and then the target can optionally define them?

Then if the user creates jump vectors in ram, the linker will complain about any missing functions the user is supposed to define ("_z80_rst_08h", etc).  That's likely an error anyways.  Individual targets can define soft defaults though, and it would have to be done for individual targets only because these will be different target to target.

feilipu:
Ok makes sense. 

Probably a too general remark for here, but (off a git tree) I'm reticent to make modifications to the defaults, because it messes up the git status for merging.

Yet development requires constant tweaking.

Is there a preference for the method to conveniently hold a set of user TAR_crt_.  values, and defc values? Either for a project or for an individual user?

Just write a personal .inc file and add it to the project for example?

aralbrec:
> Is there a preference for the method to conveniently hold a set of user TAR_crt_. values, and defc values? > Either for a project or for an individual user?
> Just write a personal .inc file and add it to the project for example?

Yes the user overrides via pragams which can be collected in a single file, eg:

zpragma.inc
```
#pragma output CRT_ORG_CODE = 32768
#pragma printf = "%c %u"
#pragma CRT_ENABLE_RST = 0x8080
...
```

This file is parsed like a C file so you can put in '//` or `/* */` style comments.

Then on the compile line, add "-pragma-include:zpragma.inc"

There is another method that includes a file from the compile directory instead of the library config file but if you use that you have to define all the pragmas instead of just overriding the ones of interest.

aralbrec:
Thanks feilipu.  I think with this latest commit, your ideas are incorporated and we have a decent model for the restarts and im2 vectors.  I'll close this pull request and any further issues can be brought up in #277.

PR #280, opened 2017-07-11
BiFrost doc

derekfountain:
Pull request for my BiFrost document. This turned into a bit of an epic, but since things are getting more complex maybe that's inevitable. I wanted to add a couple of images and a ctile resource and couldn't find a precedent to follow. I took a guess on what might be a suitable way to add those binaries.

I ended up splitting the details of the memory map, heap and stack issues out into a separate document which I've generically titled "some details". Other stuff will need to go in here at some point - interrupts probably being the next thing. For now it just covers the mechanics of moving the stack and heap, as needed for the BiFrost examples.

This one could probably benefit from a careful proof read. It certainly stretched my grey matter. :)


aralbrec:
Ok thanks derek :)   I'll merge it in now and do a proof-read.

aralbrec:
I think you did a decent job on introducing more complicated subject matter.  I only made a few changes which you can review if you like:  https://github.com/z88dk/z88dk/commit/5c58d425a87cad27c5dd9a02ea19617b057f5023

PR #282, opened 2017-07-12
yaz180 z88dk 2nd commit

feilipu:


libsrc/_DEVELOPMENT/target/crt_set_interrupt_mode.inc
```diff
@@ -27,6 +27,17 @@
       ENDIF
       
       ld i,a
+
+      INCLUDE "config_private.inc"
```
feilipu: There is a better way of doing this.
Just, I don't know what it is.
Sorry.

aralbrec: Setting IL is in the [z180 page zero code](https://github.com/z88dk/z88dk/blob/master/libsrc/_DEVELOPMENT/target/crt_page_zero_z180.inc#L743) but I think you're right that it has to be in the crt if the user is changing the vector table location.

aralbrec:
I'll have to take a look at what we're doing with IL a little later but it's ok to merge nevertheless.

PR #285, opened 2017-07-14
rc2014 heap+stack location & yaz180 3rd commit

feilipu:
Taking advantage of the case if [CLIB_MALLOC_HEAP_SIZE < -1](https://github.com/z88dk/z88dk/commit/fae0af10453508fd64dc1a20785f6a9133b87d79), to adjust the default configuration of the rc2014 Basic subtypes. Previously, it was not clear how the heap would be defined, if the stack was below the bss assignment.

origin = `0x9000`
heap = `0xFFFF`
stack = `0x9000` (decrement before push).

The RC2014 NASCOM Basic [`hexload.bas`](https://github.com/RC2014Z80/RC2014/blob/master/ROMs/hexload/hexload.bas) program needs quite a few bytes to load itself first, and I found it needs up to about `0x88FF` to work.

The active `hexload` asm code itself only needs a few bytes, and doesn't extend beyond `0x8A00`.

It might be reasonable to have default the C stack start at `0x9000` and grow down. If it grows too far it will only overwrite the `hexload` program.


feilipu:
Second independent commit for yaz180.

Please review separately.

aralbrec:
These are both in-progress targets and there are no build problems so they're ok to commit.
I'll do a closer inspection tomorrow sometime.

feilipu:
> For the yaz180, there may be a problem with the `crt_preamble.asm` code.

There are a few extra configuration steps required to get the z180 to be happy, and I am not sure of the best place to put them. I thought `crt_preamble.asm` might be the right place.

The main issue is getting the MMU set up as desired. I'd prefer some method to have a simple default as is currently configured.

Later users will have other ideas, for sure.

I'll [describe what the yaz180 has available](https://feilipu.me/2016/05/23/another-z80-project/).

The physical address space is configured by GALs to have Flash from 0x00000 to 0x3FFFF, and RAM from 0x40000 to 0xFFFFF.  There is also a RAM window inserted from 0x02000 to 0x03FFF.
The Flash can be turned off entirely, simply by reprogramming the GALs, leaving 1MB of RAM.

The point of this arrangement is to allow a ROM/RAM Global space from 0x0000 to 0x3FFF in Common 0 logical space. With a BIOS (quasi) in 0x0000 to 0x1FFF, and the global buffers / stack from 0x2000 to 0x3FFF.

The Banked and Common 1 logical spaces can be arranged within the remaining physical address space as either multiple 16kB Banks, with another 32kB of Common 1 in high RAM, or any other arrangement that suits the user application. Setting this up should be a user configuration, beyond a simple default included to get started.

PR #286, opened 2017-07-15
rc2014 bss config & yaz180 4th commit

feilipu:
Minor change to rc2014 bss default config, based on @aralbrec comment to #285.

Added z180 internal interrupts configuration.

Feels very "clunky" to have the origin of the z180 internal interrupts defined in two places in two separate files. `__crt_org_vector_table + __IO_VECTOR_BASE`

And also ugly to need to [insert 0x80](https://github.com/z88dk/z88dk/compare/master...feilipu:master#diff-44d287292022d9d2d4fdeef70cb0bb63R23) bytes, for example, because the `SECTION interrupt_vectors` doesn't start at the origin of the z180 internal vector table.

There must be a neater way to achieve this outcome. But, it is beyond me.


feilipu:
On further thought, since the `__IO_VECTOR_BASE` is a z180 specific configuration, would it be possible to have a key (eg -1) to allow it to take its value from the low byte of `__crt_org_vector_table`?

[Something like this](https://github.com/z88dk/z88dk/blob/master/libsrc/_DEVELOPMENT/target/crt_set_interrupt_mode.inc#L35).
```asm
      IF (__Z180)
      
        ; Set the z180 internal interrupt vector low byte  
        IF (__IO_VECTOR_BASE = -1)
            ld a,__crt_org_vector_table%256
        ELSE
            ld a,__IO_VECTOR_BASE
        ENDIF

        out0 (IL),a
        
      ENDIF
```
In this case there would be no need to [insert padding bytes](https://github.com/z88dk/z88dk/blob/master/libsrc/_DEVELOPMENT/target/crt_set_interrupt_mode.inc#L35) into the z180 interrupt table, because the table would be copied to the right place.

And the location of the table would be configured from just one place by `#pragma`.

Any drawbacks to this idea?

aralbrec:
> On further thought, since the __IO_VECTOR_BASE is a z180 specific configuration, would it be possible to have a key (eg -1) to allow it to take its value from the low byte of __crt_org_vector_table?

Yes that may be a good idea.  Right now the __IO_VECTOR_BASE is a library build-time constant.  The user is allowed to change it by editing `config_cpu.m4` but then he has to recompile the library to get the effect.

It may be possible to change it to a CRT variable, then it can be modified at compile time via pragma and we can do things for special values.

I'll just have to take a look to see if __IO_VECTOR_BASE needs to be library build-time constant or not.

aralbrec:
__IO_VECTOR_BASE is now a crt variable controlled by "CRT_IO_VECTOR_BASE".

Set to -1, __IO_VECTOR_BASE will be set so that the z180 vectors appear at the start of the vector table at `__crt_org_vector_table`.

Being a crt variable, "CRT_IO_VECTOR_BASE" can be set via pragma at compile time.  __IO_VECTOR_BASE reflects the value to be written to the IL register and is made PUBLIC for the compile.

PR #296, opened 2017-07-25
yaz180 - IDE via PIO 82C55 - infrastructure

feilipu:
Extends the config_target to include PIO and APU definitions, for driver development.

Initial commit containing the underlying IDE infrastructure for integrating the [ChaN FatFS](http://elm-chan.org/fsw/ff/00index_e.html) file management system with the yaz180 IDE interface.

Subsequent commits will build the Media Access Interface (diskio) functions, upon these underlying routines.

Please check that I've got this commit organised in the correct z88dk way and, if necessary, please adjust to suit the correct style. Hopefully, these can be integrated with the stdio functions appropriately, at a later stage.

Thanks in advance.


feilipu:
I've extended the [`config_target.m4`](https://github.com/feilipu/z88dk/blob/2fbbdad6d61321433edfcc44f804c3a376f6e8b0/libsrc/_DEVELOPMENT/target/yaz180/config/config_target.m4) with a number of definitions relating to to the hardware integrated into the board. But I note that the `config_cpu.m4` definitions are proceeded by double underscore.
Would it be better style to do the same, and use double underscore everywhere?

Where is the right directory to put a `C` header file, to allow access to assembly functions?
Is this typically done by `m4` as well?
Or, is the header file maintained separately?

Example header file. [diskio.h](https://github.com/z88dk/z88dk/files/1173734/diskio.h.txt)

Specifically, I need to be able to reference the `disk_initialize()`, `disk_status()`, `disk_read()`, `disk_write()`, and `disk_ioctl()` functions, which I'll add in the next commit, from the FatFS code in `C`.
Is it appropriate just to define them with `extern` and keep the `C` code separate from z88dk?

Is a zsdcc `enum` defined as 8 bits, or 16 bits?
Relevant as it is a return from a function.

aralbrec:
> I've extended the config_target.m4 with a number of definitions relating to to the hardware integrated into the board. But I note that the config_cpu.m4 definitions are proceeded by double underscore.
> Would it be better style to do the same, and use double underscore everywhere?

Yes unless the names are somehow defined by the manufacturer.  The leading double underscore is in the implementation's namespace so that names defined this way will not conflict with user names.  Everything in the config files will be exported as PUBLIC to asm and will be available to C through arch.h.  Then the idea is specific C headers will selectively expose these names by removing the two leading underscores as in [this header](https://github.com/z88dk/z88dk/blob/master/include/_DEVELOPMENT/sdcc/arch/zxn.h#L96).

For io, I think [this formula](https://github.com/z88dk/z88dk/blob/master/libsrc/_DEVELOPMENT/target/zxn/config/config_target.m4#L36) is going to be followed.  The io register is given a descriptive name and then bit fields are given names with the full io register name they are associated with abbreviated like this:

```
define(`__IO_REGISTRY_REG_MACHINE_TYPE', 3)
define(`__IO_RRMT_LOCK_TIMING', 0x80)
define(`__IO_RRMT_TIMING_48', 0x10)
define(`__IO_RRMT_TIMING_128', 0x20)
define(`__IO_RRMT_TIMING_P3', 0x30)
define(`__IO_RRMT_TIMING_PENTAGON', 0x40)
define(`__IO_RRMT_CONFIG', 0x00)
define(`__IO_RRMT_48', 0x01)
define(`__IO_RRMT_128', 0x02)
define(`__IO_RRMT_P3', 0x03)
define(`__IO_RRMT_PENTAGON', 0x04)
```

The io port is `__IO_REGISTRY_REG_MACHINE_TYPE` and the bitfields all begin with `__IO_RRMT_` where "RRMT" = **R**egistry **R**eg **M**achine **T**ype.

> Where is the right directory to put a C header file, to allow access to assembly functions?
> Is this typically done by m4 as well?
> Or, is the header file maintained separately?

The headers are maintained separately.  The appropriate place for target-specific headers is in [include/_DEVELOPMENT/sdcc/arch](https://github.com/z88dk/z88dk/tree/master/include/_DEVELOPMENT/sdcc/arch) where a general "yaz180.h" may sit and then more specific headers for the yaz180 can be in the subdir arch/yaz180.

Headers are defined in one place:  [include/_DEVELOPMENT/proto](https://github.com/z88dk/z88dk/tree/master/include/_DEVELOPMENT/proto).  This tree holds the prototype headers from which m4 generates all the headers for clang, sccz80 and zsdcc.  After a header is defined here, Winmake.bat is executed from [include/_DEVELOPMENT](https://github.com/z88dk/z88dk/tree/master/include/_DEVELOPMENT) to get m4 to create the headers for the clang,sccz80,zsdcc directories.  It's only a windows batch file at the moment and I execute it, then commit changes.  It is not yet part of the z88dk build process.

The definitions of the headers are using m4 macros to define functions.  `__DPROTO` defines  function that will be fastcall or callee.  `__OPROTO` will do ordinary functions and will not apply fastcall or callee; this should be used for functions taking no arguments or for functions that are vararg.

You can take a look at some of the headers in the proto subdir to see how functions are defined.  The format is similar to this:

```
__DPROTO(`b,c,d,e',`b,c,d,e',int,,ffs,int i)

The first parameter `b,c,d,e' is for zsdcc and indicates which registers are preserved in fastcall or callee invokes.

The second parameter `b,c,d,e' is for zsdcc and indicates which registers are preserved for ordinary function calls.  This applies to calls through function pointers mainly.

The third and fourth parameters "int,," are the return type with the fourth parameter either empty or solely for asterisks.

The fifth parameter "ffs" is the function name.

Any parameters that follow are the function parameters.
```

The corresponding asm and c interface code is found in libsrc/_DEVELOPMENT/arch or if you deem it driver related, it can be in libsrc/_DEVELOPMENT/target/yaz180/...

> Specifically, I need to be able to reference the disk_initialize(), disk_status(), disk_read(), disk_write(), and disk_ioctl() functions, which I'll add in the next commit, from the FatFS code in C.
> Is it appropriate just to define them with extern and keep the C code separate from z88dk?
> 

It can be treated like a 3rd party library.

3rd party libraries can be incorporated into z88dk by placing headers in include/_DEVELOPMENT/proto/arch/yaz180 .  Then they can be built by the user running Winmake in include/_DEVELOPMENT (there needs to be a makefile there too for non-windows) and they are included with `#include <arch/yaz180/libname.h>`

After the 3rd party library is built, its lib file can be put in [libsrc/_DEVELOPMENT/lib/sdcc_iy/yaz180](https://github.com/z88dk/z88dk/tree/master/libsrc/_DEVELOPMENT/lib/sdcc_iy/yaz180).  Actually three versions of the library are normally built, one for sccz80, one for sdcc_ix and one for sdcc_iy.

Building a library is very simple - it's like compiling but you add the "-x" option.  The linker works on a file granularity so you should be separating each individual function into its own file and make a list of files in the library in a project .lst file to simplify the build line.   Quite likely 3rd party code does not separate each function into its own file.

aralbrec:
> Is a zsdcc enum defined as 8 bits, or 16 bits?
> Relevant as it is a return from a function.
> 

I think it chooses the smallest possible int type from char, int, long, longlong.  I'll have to check that tomorrow.

feilipu:
Thanks.
Not sure why Travis failed. Looks like it timed out.

Would it be best to go back with this #296 issue to revise all of the constant names?
I really should do that for the rc2014 ACIA names as well.
Or, commit this issue as is, and let me start another issue for the name revision?

aralbrec:
> Would it be best to go back with this #296 issue to revise all of the constant names?
> I really should do that for the rc2014 ACIA names as well.
> Or, commit this issue as is, and let me start another issue for the name revision?

Let's commit this first and then make changes to make things easier.  I'll try to do a check today.

aralbrec:
I don't know what the best way is to organize peripheral drivers yet.  Commit it and figure it out.

aralbrec:
> I think it chooses the smallest possible int type from char, int, long, longlong. I'll have to check that tomorrow.

Yes it chooses the smallest int type, taking into consideration unsigned and signed.

This is an unsigned char:
`enum test { start = 0, lower, middle = 10, upper, end = 129 };`

Whereas this is a signed int:
`enum test { start = -1, lower, middle = 10, upper, end = 129 };`

feilipu:
Converted all `config_target.m4` definitions to double underscore in #297.

Definitions is `config_cpu.m4` are not converted, at this stage.

PR #311, opened 2017-08-10
yaz180 - c interface for diskio

feilipu:
work in progress for the diskio functions, this time a c interface has been added.


aralbrec:
That's a lot of painful work :)

I'll help to fix a few of the problems I see here in the near term - I'm in the zx next libs at the moment.   For sccz80 there is no distinction between fastcall and standard linkage because in both cases the compiler deliberately  has the parameter in hl.

For sdcc_ix, you must preserve both ix and iy.  For sdcc_iy, just iy must be preserved.

I'll commit these things and the problems can be cleared up later.

feilipu:
Thanks. I'm testing with `sdcc_iy` using the [`diskio_check.c`](https://github.com/z88dk/z88dk/files/1218073/diskio_check.c.txt) code provided by ChaN. Hopefully should have the asm code working in a week.

So there's still quite a bit of work to get the diskio functions right, but when they do work, then FatFS functions should also "just work".

PR #314, opened 2017-08-11
Wrong sequence

polluks:
See above.

### DIFF
--- a/include/conio.h
+++ b/include/conio.h
@@ -63,7 +63,7 @@ static int PCDOS_COLORS[]={0,4,2,6,1,5,1,7,4,6,2,6,1,5,3,7};
 #define highvideo()	printf("\033[1m")
 #define lowvideo()	printf("\033[2m")
 #define normvideo()	printf("\033[m")
-#define delline()	printf("\033[m")
+#define delline()	printf("\033[M")
 //#define clrscr() textattr(7);fputc_cons(12)
 /* In this way we can do some sort of color and setcolor functions redefinition */
 #define clrscr() printf("\033[%um\033[%um%c",30,47,12)


aralbrec:
That code looks like a scroll.

Should it be esc[2K ?

polluks:
delline requires scrolling
http://www.delorie.com/djgpp/doc/libc/libc_129.html

aralbrec:
I see two or three different documentation for that vt100 escape code.

I think the most trustworthy is this:
http://www.csie.ntu.edu.tw/~r92094/c++/VT100.html

index IND             Move/scroll window up one line         ^[D
revindex RI           Move/scroll window down one line       ^[M

So what is needed is a scroll up from the current row.  Maybe ^[D is the right code if it's not full screen?
@zx70 can the ascii driver even do this?

polluks:
Just look at the other definitions of delline in conio.h, it's a simple typo: m → M

aralbrec:
I agree with you but I think the intended M is the wrong escape sequence.

The link you gave says delline deletes the line and scrolls the area underneath upward one row.  ^[M scrolls downward which is the wrong direction.  And I'm not sure if the vt100 driver in classic z88dk can scroll a region.  However fixing the control code to what it should be can be done now even if the driver won't interpret it correctly atm.

polluks:
Well, I'm talking about ^[[M not ^[M
http://www.real-world-systems.com/docs/ANSIcode.html

aralbrec:
Ah ok, that's a vt102 escape sequence.

I can see in the classic ansi driver:
https://github.com/z88dk/z88dk/blob/master/libsrc/stdio/ansi/f_ansi.asm
that ^[[M is not implemented.

It'll be ignored and not crash so it's ok to commit.

PR #317, opened 2017-08-18
yaz180 - C interface for 8255 IDE interface - WIP

feilipu:
Some updates to the asm and C interface for the yaz180 8255 PIO IDE interface.

Not sure that the C interface is being handled correctly, but it seems to be working.
Would appreciate having it corrected, so that I know the correct way to do it.


aralbrec:
ok will have a look at the c interface soon

PR #334, opened 2017-08-28
yaz180 - Am9511A APU operation declarations

feilipu:
Add the operation declarations for the Am9511A APU.
[asm driver](https://github.com/feilipu/LLL-Floating-Point/blob/master/am9511a.asm), and c interface following some further testing.


feilipu:
Sometimes, the difference between success and failure is just 42ns.
e wr 42ns
:+1:

PR #347, opened 2017-09-05
yaz180 - driver C header prototype files for apu & diskio

feilipu:
I've attempted to add these C header prototype files to a `~/include/_DEVELOPMENT/proto/arch/yaz180` directory. I'm not 100% certain that I've done it correctly, or that they're in the right place. But, a starting point for later correction.

The associated asm and C code is in the `~/libsrc/_DEVELOPMENT/target/yaz180/driver/` directories.


aralbrec:
```
/* This type MUST be 64-bit (Remove this for ANSI C (C89) compatibility) */
typedef uint64_t        QWORD;
```

Is this being passed around in functions (as opposed to pointers to it) or used to do arithmetic?  64-bit numbers have special argument passing with zsdcc (and are unsupported in sccz80).  If it's being treated as an 8-byte container maybe sccz80 can be given a struct of two 32-bit numbers to be compatible in most circumstances?

Things look correct to me, one function:

```
__OPROTO('a,f,b,c,d,e,ixh,ixl,iyh,iyl','a,f,b,c,d,e,ixh,ixl,iyh,iyl',void,,apu_reset,void *int_addr)
```

maybe should be DPROTO?   DPROTO is going to do fastcall and callee declarations for functions with 1 or more arguments.  OPROTO is used if you only want standard linkage and must be used for 0-argument functions or var arg functions.

The preserves registers list only pays attention to `a,b,c,d,e,h,l,iyl,iyh` and I'm not sure what happens if `f,ixl,ixh` are in there but it could be they are just ignored, in which case this is kind of future proofing.

feilipu:
```c
/* This type MUST be 64-bit (Remove this for ANSI C (C89) compatibility) */
typedef uint64_t        QWORD;
```
This isn't used at all in the diskio functions. The declaration comes from its use in the FatFS functions, which use the same `ffinteger.h`, provided by ChaN, which I just copied over. `QWORD` can be safely removed.

The `apu_reset` function is only called once (where the CRT0 is not included) and doesn't need to be optimised for speed, hence I haven't done a `_callee` or `_fastcall` version. It is ok as standard linkage.

The only function that needs to be optimised is `apu_cmd_ld` which has a `_callee` asm function linkage provided. But not a `_fastcall` linkage. Does it matter if only one of the two options is provided?

Some of my asm functions preserve registers, to ensure they only clobber the registers that they're actually using. Therefore they push/pop on entry/exit.

Is it better form or style to build "naked" asm functions for C interfacing, that don't try to preserve the unused registers, and have an additional asm function (that will be used by users) to do the push/pop before it calls the naked function?

aralbrec:
> QWORD can be safely removed if exFAT is not being used. QWORD is being used in just two functions in the ff.c, wrapped by a #if. I'm sure there's another way to achieve this outcome using a structure.

`ld_qword` and `st_qword` are just doing a load and store of a 64-bit number in little endian form.  That's the natural byte order for the z80 so these could be replaced by simple assignment in zsdcc.  How those QWORDs are used will have to be looked into (what calls `ld_qword` and `st_qword`) to figure out the best way to work around things.  I think for now we can just conditionally typedef QWORD to an array of bytes for sccz80 and figure this out later.

> The only function that needs to be optimised is apu_cmd_ld which has a _callee asm function linkage provided. But not a _fastcall linkage. Does it matter if only one of the two options is provided?

Functions can only be one or the other.  With > 1 arguments, it's callee and for exactly one argument it's fastcall.  So apu_cmd_ld can only be callee so that's fine.  DPROTO is doing a prototype for standard linkage in addition to fastcall/callee for function pointers.

> The additional registers have to be removed, otherwise this happens. The build works for me when I pull the commits around this back, so I don't know what the problem is.

Ok so you cannot add `f`, `ixl`, `ixy` to the preserves list.  It looks like the fail was just a timeout.

> Is it better form or style to build "naked" asm functions for C interfacing, that don't try to preserve the unused registers, and have an additional asm function (that will be used by users) to do the push/pop before it calls the naked function?

You'd end up with extra calls in there and that has cost too.  I always just do what's best for asm in the asm.  That's also going to be good for c because the execution time is in the asm code.   For preserving registers, that's up to you.  I actually don't do it that often as I tend to write in caller save style.

PR #348, opened 2017-09-06
Md5sum optimisations

suborb:
Some work on the test program from the ZX next forum: http://specnext.com/forum/viewtopic.php?f=13&t=736&sid=ec0ddd0f1027796507bd49453348e587

Basically an MD5 checksum routine that makes heavy use of 32 bit shifts. Previously some work had be done on << shifting, but this program also uses a lot of >> 

There's now optimised cases for most shift values, most of these come out longer, but naturally faster. There's the start of a config option within sccz80 to favour speed over size.

ticks has been updated to calculate hotspots/frequency of instructions. It's indicative since it doesn't take account of how long an instruction takes but is still useful.

I've also inlined (always) the constant l_long_add, this turns out to the be the same size, but we save us the call/ret so if used a lot it's fairly significant.

So the figures, this running on a smallish file (actually a 13668 binary of a previous build of md5sum):

Master:

zcc +test md5sum.c

-clib=default
Size: 13669
Ticks: 66084762

This branch:

-clib=default -O2
Size: 14084
Ticks: 50368504

-clib=default -O1
Size: 14136
Ticks: 49473751

-clib=default -sccz80-inline-ints -O2
Size: 14207
Ticks: 48762706

-clib=default -sccz80-inline-ints -O1
Size: 14262
Ticks: 47867851

-clib=zxn -O2
Size: 13995
Ticks: 49814290

-clib=zxn -O1
Size: 14028
Ticks: 48919474

-clib=zxn -sccz80-inline-ints -O2
Size: 14118
Ticks: 48208492

-clib=zxn -sccz80-inline-ints -O1
Size: 14154
Ticks: 47313568

It's a classic space/speed trade off. I think it's probably worthwhile, but thoughts welcome.

It also shows that something is wrong in z80rules.2 - it may well be clobbering a later optimisation.


suborb:
@aralbrec have you got any thoughts on this one?

aralbrec:
Looks ok.  You only have to add l_longhlp to the sccz80 lst file and the source directory to get it added to newlib as you've done.

Enabling z80-zxn cpu for sccz80 in zcc.c may still be a problem.  Today's release of ZEsarUX still doesn't have any new opcodes in it and that's the only real emulator available.

suborb:
Numbers have changed after the division fix earlier. We pull in some extra errno stuff which makes the binary a bit larger, but it's a bit faster again.

>Enabling z80-zxn cpu for sccz80 in zcc.c may still be a problem. Today's release of ZEsarUX still doesn't have any new opcodes in it and that's the only real emulator available.

Hmmm, maybe we should remove -mz80-zxn in zxn.cfg. A quick search doesn't show any uses of the z80asm macro.

aralbrec:
> Hmmm, maybe we should remove -mz80-zxn in zxn.cfg. A quick search doesn't show any uses of the z80asm macro.

That enables it for user asm code only, which makes it up to the user.  Some are using CSpect which does support these opcodes but it can't do esxdos dot commands and some features from esxdos, which others are using with ZEsarUX.

suborb:
Where we are at the moment.

-clib=default -O2
Size: 13706
Ticks: 49121317

-clib=default -O2 -sccz80-inline-ints
Size: 13829
Ticks: 47515519

-clib=zxn -O2
Size: 13615
Ticks: 48566814

-clib=zxn -sccz80-inline-ints -O2
Size: 13738
Ticks: 46961016

I'm a lot happier with the size coming back down. The speed increase is nice as well. The -O1 anomaly has gone as well.

Issue #342, opened 2017-09-02
z80asm: new section model supporting overlays


pauloscustodio:
There is a need to define a section model that allows the definition of banked-switched memory models, where the assembler can generate code from different sections to the same address range. There is a need for overlays, where the compiler inserts bank-switching code before calling a function. There is also a need to pass more detailed information to downstream tools (memory loader/relocator, debugger, disassembler) and to get more information from upstream tools (source file and line information from the C compilers). 

Several issues have been opened in the past on this issue. The goal of this issue is to settle on the requirements and design.

References: #13, #19, #42, #71, #156, #243, #273, #333, #341

pauloscustodio:
Requirements:
- Assembly code is be grouped in sections, sorted at link time.
- Minimum entity handled is the set of code lines from one source file in one section.
- Sections are grouped in section groups.
- Each section group:
    - may be allocated at a physical memory address in the 64k space, defined in the code, or overridden in the link command line;
    - may be limited to a maximum size;
    - may flow around "holes" in the memory map;
    - may be squashed with a previous section group, building one logical section group in the output;
    - produces one output binary file for each continuous area of memory;
    - may be compressed at compile time in order to be expanded in the start-up code;

PR #349, opened 2017-09-07
yaz180 - diskio and apu updates

feilipu:
Finally...
Working diskio (in preparation for ChaN FatFS).

```bash
**** Test cycle 3 of 3 start ****
 disk_initalize(0) - ok.
**** Get drive serial number ****
 disk_ioctl(0, ATA_GET_SN, 0xA828) - ok.
Serial number of the drive 0 is          HR0RRQG3912.
**** Get drive size ****
 disk_ioctl(0, GET_SECTOR_COUNT, 0x817F) - ok.
Number of sectors on the drive 0 is 12685680.
**** Get sector size ****
 disk_ioctl(0, GET_SECTOR_SIZE, 0x8185) - ok.
 Size of sector is 512 bytes.
**** Get block size ****
 disk_ioctl(0, GET_BLOCK_SIZE, 0x8183) - ok.
 Size of the erase block is 1 sectors.
**** Single sector write test ****
 disk_write(0, 0xA828, 0, 1) - ok.
 disk_ioctl(0, CTRL_SYNC, NULL) - ok.
 disk_read(0, 0xA828, 0, 1) - ok.
 Data matched.
**** Multiple sector write test ****
 disk_write(0, 0xA828, 1, 4) - ok.
 disk_ioctl(0, CTRL_SYNC, NULL) - ok.
 disk_read(0, 0xA828, 1, 4) - ok.
 Data matched.
**** Single sector write test (misaligned address) ****
 disk_write(0, 0xA82B, 5, 1) - ok.
 disk_ioctl(0, CTRL_SYNC, NULL) - ok.
 disk_read(0, 0xA82D, 5, 1) - ok.
 Data matched.
**** 4GB barrier test ****
 disk_write(0, 0xA828, 6, 1) - ok.
 disk_write(0, 0xAA28, 8388614, 1) - ok.
 disk_ioctl(0, CTRL_SYNC, NULL) - ok.
 disk_read(0, 0xA828, 6, 1) - ok.
 disk_read(0, 0xAA28, 8388614, 1) - ok.
 Data matched.
**** Test cycle 3 of 3 completed ****

Congratulations! The disk driver works well.
```
And adding the APU Service Request flag (for DMA)


feilipu:
I'm not sure I can say how good it feels to finally... compile this code... load it onto my Z180... run it...
```cpp
/*----------------------------------------------------------------------*/
/* Foolproof FatFs sample project for Z80                               */
/*----------------------------------------------------------------------*/

// zcc +yaz180 -subtype=basic_dcio -v --list -m -SO3 -clib=sdcc_iy --max-allocs-per-node200000 @ffmain.lst -o ffmain -create-app

#include <stdio.h>
#include <string.h>

#include "ff.h"		/* Declarations of FatFs API */

FATFS FatFs;		/* FatFs work area needed for each volume */
FIL Fil;			/* File object needed for each open file */

int main (void)
{
	UINT bw;
	f_mount(&FatFs, "", 0);    /* Give a work area to the default drive */
	if (f_open(&Fil, "newfile.txt", FA_WRITE | FA_CREATE_ALWAYS) == FR_OK)
	{	/* Create a file */
		f_write(&Fil, "It works!\r\n", 11, &bw);    /* Write data to the file */
		f_close(&Fil);    /* Close the file */
		printf("It works!\r\n");
	}
	return 0;
}
```
Then plug the ATA IDE drive into my normal PC,
and find a simple file `NEWFILE.TXT` containing the words "It works!".

That is the beauty of [ELM ChaN's service](http://elm-chan.org/fsw/ff/00index_e.html).
Get the diskio working, and the FatFS layer comes for free.

:smiley: :smiley: :bouquet: :champagne:

aralbrec:
Nice!  How big is it when compiled?  Have you tried --opt-code-size as well?

feilipu:
The binary size is variable, depending on which functions are enabled in the `ffconfig.h` file at compile time. I've not done any testing on size for z80, nor on the other functions as  yet. (Netflix and chill night). Soon.

However, I think function optimisation is one of the issues for z88dk, in that putting it in as a compiled library, will invariably hit the wrong optimisation point (in terms of functions available). Also, under SDCC, it takes an age to compile. It will probably be best just to include the source somewhere, and advise the user to configure and compile it according to their needs.

Using LFN consumes quite a bit of space, because ChaN has the `ff_wtoupper()` functions and friends in the same file as a large Unicode blob. GCC has two optimisation settings `-ffunction-sections` and `-fdata-sections` that take a single file and divide it up into individual link sections, where the unused functions or data can be discarded. Certainly discarding the unused UTF data section would be useful in this case.

Also, the most additional space would be taken by using the UTF capability. I have never used it, as I've stuck to code page 437 (USA), which disables all of the UTF data. In avr-gcc this data simply is not linked in.

More to come. I might open a discussion issue on this, if you think it is something that FatFS interests z88dk users more broadly. Thoughts?

pauloscustodio:
I have a feature in the never-ending-todo of z80asm that might help in the splitting of a source file in separate linking blocks to allow the linker to choose just the used functions - it is the definitions of multiple MODULEs inside a source file, each module originating a separate linking entity - see #342.

feilipu:
Using the `ffconfig.h` file in [Version 13 Patch 2](https://github.com/z88dk/z88dk/files/1288178/ff13p2.zip) code, I can get some example simple numbers for code size.

Normal compile using zsdcc, without any stdio included:
```
__code_driver_size   = $0355
__code_compiler_size = $5B4B
```
Using `--opt-code-size`:
```
__code_driver_size   = $0355
__code_compiler_size = $5485
```
So the `--opt_code_size` option saves about 1,734 bytes in the `ff.c` library and small `ffmain.c`.

Total size is about 21kB. This seems like quite a lot, because the other platforms referenced in the [Application Note](http://elm-chan.org/fsw/ff/doc/appnote.html) require typically less than 10kB for a similar configuration. So, I'm not sure I'm counting this correctly.

I can't get the zcc80 sizes, because there's an error with compilation. I'll flag it in another issue.

aralbrec:
> Total size is about 21kB. This seems like quite a lot, because the other platforms referenced in the Application Note require typically less than 10kB for a similar configuration. 

I'll take a look at this.   zsdcc generates very verbose code for 32-bits and above.  There is some relief with `--opt-code-size` but there is a lot more we can do and I've been working a bit on that with another example that came up.  If the problem is 32-bitness then there should be another size reduction coming once I've done some more fiddling.

Issue #352, opened 2017-09-07
szcc80 - math issue


feilipu:
Related to issue #351, szcc80 causes calculation error.
```cpp
#include <stdio.h>
#include <string.h>

int main (void)
{
    uint32_t lba, lba2;
    uint16_t  sz_sect;
    
    sz_sect = 512;
    lba = 6;
    lba2 = lba + 0x80000000 / (sz_sect / 2);
    printf("%lu\n", lba2);
    return 0;
}
```
With SDCC the correct result.
```bash
zcc +yaz180 -subtype=basic_dcio -v --list -m -SO3 -clib=sdcc_iy --max-allocs-per-node200000 calculation_check.c -o calculation_check -create-app

Ok
print usr(0)
8388614
 0 
Ok
```
With SZCC80 an incorrect result.
```bash
zcc +yaz180 -subtype=basic_dcio -v --list -m -SO3 -clib=new calculation_check.c -o calculation_check -create-app

Ok
print usr(0)
6
 0 
Ok
```

suborb:
The operands for the long division are crossed. So we end up calculator divisor / dividend. I think this only occurs in the case where the constant is on the lhs of the expression.

I think #116 introduced this one, I'll fix this by the end of the weekend I hope.

suborb:
As usual a rabbit hole opened. But should be good now.

PR #356, opened 2017-09-08
Issue #352: Sort some problems with int32 division.

suborb:
If a constant was on the right then the division was incorrect (quotient/dividend).

The classic library signed division was incorrect if dividend was -1. Use newlib
versions of division to cut down code duplication.


suborb:
Mac builds are down on travis, this passed linux. 

Merging

PR #360, opened 2017-09-10
Add a -trace flag to ticks

pauloscustodio:
Add a -trace flag to run ticks in trace mode, showing the registers, flags and disassembly on stdout.

### DIFF
--- a/src/ticks/debugger.c
+++ b/src/ticks/debugger.c
@@ -69,9 +69,9 @@ static command commands[] = {
 
 static breakpoint *breakpoints;
 
-static int debugger_active = 1;
+       int debugger_active = 0;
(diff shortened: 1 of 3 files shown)


suborb:
I already added a trace command to the debugger. It's probably nicer to fold the two features together rather than have two different implementations.

pauloscustodio:
My need was to run the ticks with a -trace flag and no manual intervention, and to analyze the output for the unwanted changes in the flags. You can ditch my changes if this is done some other way.

suborb:
I'll fold it in so that -trace uses the debugger version - and still runs without intervention.

pauloscustodio:
Works perfectly. Thanks!

PR #367, opened 2017-09-13
Interrupt management in C

feilipu:
Expose the `cpu_push_di()` and `cpu_pop_ei()` functions for use in C.
Requires C programs provide `#include cpu.h`

Implemented for z80 and z180 cpus.
Not sure how to do it for rabbit. Sorry.


feilipu:
Compiler `__critical` does this better.

aralbrec:
The c equivalents are actually z180_get_int_state() -- does not di -- and z180_set_int_state() which store the interrupt state in a communicated byte.  There probably needs to be a define for cross-cpu compatible versions as in cpu_get_int_state().

The cpu_push_di and cpu_pop_ei functions modify the stack so they can't be used in a c context because there is no way to tell the compilers that the stack pointer has moved.

PR #373, opened 2017-09-13
yaz180 - initialise single step hardware

feilipu:
Provide C mechanism (macro for inline asm) for initialising the single step hardware.
Writing to `__IO_BREAK` causes the WAIT signal to halt the CPU.

It would be good to know if there is a short cut to refer to these kind of header files,
as they are very deep in the hierarchy.

Below example is very long form and unwieldy.
`#include "~/z88dk/include/_DEVELOPMENT/sdcc/arch/yaz180.h"`


aralbrec:
> Below example is very long form and unwieldy.
> #include "~/z88dk/include/_DEVELOPMENT/sdcc/arch/yaz180.h"

`#include <arch/yaz180.h>` should work?

The macro inlines assembler so it interferes with the peephole optimizer in sdcc.  We have a kludge that will do the inlining without affecting the optimizer.  I'll add that for the yaz180 in #374 and then show how the BREAK macro can be implemented this way there.

Issue #374, opened 2017-09-13
newlib +yaz180:  add copt rules file to allow inlining of asm functions


aralbrec:
From #373.
Add mechanism to yaz180 to inline asm instructions without affecting optimizer.

aralbrec:
I can see this becoming a thing where each target may get additional post-processing steps.  We probably should have a better place to store the various peephole rules for copt, sccz80 and sdcc.

aralbrec:
For __BREAK, I've changed the yaz180.h proto header to this:

```
// Halt the YAZ180 with single step hardware.

#define __BREAK  __BREAK_HELPER()
__OPROTO(`a,d,e,h,l,iy,iyh',`a,d,e,h,l,iyl,iyh',void,,__BREAK_HELPER,void)
```

The define to `__BREAK_HELPER()` allows `__BREAK;` to be used as is inside c code.

`__BREAK_HELPER` is being defined as a function that preserves all registers except bc.  This way inlining the asm won't affect any registers zsdcc is using.

In a new file yaz180_rule.1, a copt rule is added to replace the function call the compilers are generating with inlined asm:

```
	call	___BREAK_HELPER
=
	IFNDEF __IO_BREAK
		EXTERN	__IO_BREAK
	ENDIF
	ld	bc,__IO_BREAK
	out	(c),c

	call	__BREAK_HELPER
=
	IFNDEF __IO_BREAK
		EXTERN	__IO_BREAK
	ENDIF
	ld	bc,__IO_BREAK
	out	(c),c
```

zsdcc adds another leading underscore to the name, sccz80 does not.  I'm checking if `__IO_BREAK` is defined in case someone does something bizarre like `include "config_private.inc"` somehow as that will then include a file defining that label; I think an EXTERN will cause an error out of z80asm in that case.  The EXTERN will work because the crt exports those labels as PUBLIC.

suborb:
> I can see this becoming a thing where each target may get additional post-processing steps. We probably should have a better place to store the various peephole rules for copt, sccz80 and sdcc.

Yes, that was my thinking behind: https://github.com/z88dk/z88dk/tree/feature/more_build_changes/lib/arch/z88 Something didn't feel quite right about what I was doing though so I halted work on it.

feilipu:
I'm afraid that I've got a case where this doesn't quite work.

Where there is a simple assignment of function, the `___BREAK_HELPER` doesn't work.
Probably this won't occur very often (as it creates a useless program), but still.
[...]

PR #379, opened 2017-09-15
yaz180 - handle tail call optimisation for __BREAK

feilipu:
A simple fix for #374 and #373 for tail call optimisation.

I spent a while looking at pragmas to disable optimisation,
but sdcc doesn't allow `#pragma` within a function.

I think this achieves the desired outcome, without much further thought investment.

Just to note that this solution is whitespace sensitive.
`call  __BREAK_HELPER` contains a tab character
`jp  __BREAK_HELPER` contains two spaces
If the peep-hole optimisation whitespace usage changes, this solution may break.

### DIFF
--- a/include/_DEVELOPMENT/clang/arch/yaz180.h
+++ b/include/_DEVELOPMENT/clang/arch/yaz180.h
@@ -9,7 +9,8 @@
 
 // Halt the YAZ180 with single step hardware.
 
-#define __BREAK  __BREAK_HELPER()
+#define __BREAK __BREAK_HELPER()
+
 extern void __BREAK_HELPER(void);
 
 
--- a/include/_DEVELOPMENT/proto/arch/yaz180.h
+++ b/include/_DEVELOPMENT/proto/arch/yaz180.h
@@ -7,7 +7,8 @@ include(__link__.m4)
 
 // Halt the YAZ180 with single step hardware.
 
-#define __BREAK  __BREAK_HELPER()
+#define __BREAK __BREAK_HELPER()
+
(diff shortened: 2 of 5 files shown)


aralbrec:
I have a solution here which I think I will commit shortly:

```
%"[ \t]*"0call%"[ \t]*"1___BREAK_HELPER
=
	IFNDEF __IO_BREAK
	EXTERN	__IO_BREAK
	ENDIF
	ld	bc,__IO_BREAK
	out	(c),c

%"[ \t]*"0call%"[ \t]*"1__BREAK_HELPER
=
	IFNDEF __IO_BREAK
	EXTERN	__IO_BREAK
	ENDIF
	ld	bc,__IO_BREAK
	out	(c),c

%"[ \t]*"0jp%"[ \t]*"1___BREAK_HELPER
=
	IFNDEF __IO_BREAK
	EXTERN	__IO_BREAK
	ENDIF
	ld	bc,__IO_BREAK
	out	(c),c
	ret

%"[ \t]*"0jp%"[ \t]*"1__BREAK_HELPER
=
	IFNDEF __IO_BREAK
	EXTERN	__IO_BREAK
	ENDIF
	ld	bc,__IO_BREAK
	out	(c),c
	ret

%"[ \t]*"0defc%"[ \t]*"1%"[A-Za-z0-9_]*"4%"[ \t]*"2=%"[ \t]*"3___BREAK_HELPER
=
%4:
	IFNDEF __IO_BREAK
	EXTERN	__IO_BREAK
	ENDIF
	ld	bc,__IO_BREAK
	out	(c),c
	ret
```

Were it not for the nature of "BREAK" another solution may have been to add an actual function to the library to catch these cases.  It may be more efficient to do that, particularly for the defc case.

aralbrec:
#374 addresses the issue and borrows from here

PR #383, opened 2017-09-16
Feature/classic reorg

suborb:
Fold a while bunch of stuff into the core platform libraries.

Use newlib string code and fix a couple of issues spotted.


suborb:
@aralbrec I've stuck a few IF CLASSIC into newlib so that ix can be saved - do you have a preference for a better name?

aralbrec:
Maybe make sure the name is in the system namespace ie "__CLASSIC" with two leading underscores.

suborb:
Good idea. Thanks. 

I'll merge this later this evening.

Just for your interest, it looks like classic is now using around about 1300 newlib labels out of just over 6000 in classic.

As a note, it appears that newlib is missing strrstr - I've kept the classic version around.

PR #397, opened 2017-09-19
Inverted check for NULL

pauloscustodio:
Test for NULL was inverted.


aralbrec:
Thanks Paulo, I've incorporated into the main branch.

PR #398, opened 2017-09-19
Fix 396: build lstmanip in Visual Studio

pauloscustodio:
Had to pull-in the getopt.c implementation.


suborb:
Thanks for doing this. My windows8 tablet can barely cope with notepad, I daren't try Visual studio!

In the makefile I just referred to the version in z80nm. Maybe we should hoist them to src/win32 and reference that way though?

pauloscustodio:
I agree. I will do it.

suborb:
Thanks for doing this @pauloscustodio

suborb:
Travis is struggling with mac builds again, but this passed linux so merging.

PR #403, opened 2017-09-24
rc2014 - add 82C55 ide and diskio support

feilipu:
A 82C55 PIO based IDE interface is being constructed to support CP/M and FatFs file access.

The eventual port address for the hardware is still to be defined. But that is [one value](https://github.com/z88dk/z88dk/pull/403/files#diff-43f197db3d12426aab9ef215b2d8ed68R14).


aralbrec:
I'm quite a bit behind here - not enough time :)

The defines you are putting into *.asm files may not be necessary (eg libsrc/_DEVELOPMENT/target/rc2014/device/8255/8255_defines.asm)?  If they are part of a config file, they will be exported as PUBLIC by the crt or could be included via config_rc2014_private.inc for no-crt compiles.

feilipu:
> The defines you are putting into *.asm files may not be necessary (eg libsrc/_DEVELOPMENT/target/rc2014/device/8255/8255_defines.asm?
> If they are part of a config file, they will be exported as PUBLIC by the crt or could be included via config_rc2014_private.inc for no-crt compiles.

I actually just copied the solution mechanism over from what I did with yaz180, which at the time was done just for that specific target. So now almost the same source exists twice! See below for minor differences

If other "home built" targets want to use the IDE solution using 82C55, then perhaps the solution could / should be generalised? Perhaps (when a bit more time is at hand), it might make sense to generalise the solution a little to share the device and driver code between the two platforms. I'm not sure where it would be appropriate to put it then.

The particular defines you mention are simply to specify [how the ports of the 82C55 are configured](https://github.com/z88dk/z88dk/blob/master/libsrc/_DEVELOPMENT/target/rc2014/device/8255/8255_defines.asm#L20) for the IDE interface, and then how the [pins are assigned to IDE control lines](https://github.com/z88dk/z88dk/blob/master/libsrc/_DEVELOPMENT/target/rc2014/device/8255/8255_defines.asm#L33) on Port C for both builds.

The standard solution is to use Port C for the IDE control lines. It is only in the ordering of the Port C pins that there is a difference. The rc2014 solution (Ed is building the hardware), is to go with the RomWBW configuration, that was copied two prior iterations in my knowledge, from the original invention in 1990s. This ordering is "standard". I used a different [pin ordering in yaz180](https://github.com/z88dk/z88dk/blob/master/libsrc/_DEVELOPMENT/target/rc2014/device/8255/8255_defines.asm#L33), because it supports Mode 1 and Mode 2 usage for the 82C55 by not putting an inverter on the 82C55 input pins.

The other difference between the two builds is the [base address for the 82C55](url). I have left it the rc2014 version the same as yaz180, until we get better info. It will need to be adjusted in the config file.

Perhaps the right solution is to build a new config file for rc2014, [config_ide.m4](https://github.com/z88dk/z88dk/tree/master/libsrc/_DEVELOPMENT/target/rc2014/config), which just covers the IDE related things, basically the content of the [82C55_defines.asm](https://github.com/z88dk/z88dk/blob/master/libsrc/_DEVELOPMENT/target/rc2014/device/8255/8255_defines.asm) file you mentioned? Thoughts?


OK, and for the record, the only other differences between the source versions are use of z180 `tst nn` specific opcodes in two places, in [ide_wait_ready](https://github.com/z88dk/z88dk/blob/master/libsrc/_DEVELOPMENT/target/yaz180/driver/ide/ide_wait_ready.asm#L30) and [ide_wait_drq](https://github.com/z88dk/z88dk/blob/master/libsrc/_DEVELOPMENT/target/yaz180/driver/ide/ide_wait_drq.asm#L27).  They've been changed to `push af`, `and nn`, `pop af` for rc2014 z80. Conditional compile is an option?

If you think something needs to be done, let me know please.

feilipu:
I had a go at writing a `config_ide.m4` for the yaz180 target. This would incorporate both [`8255_defines.asm`](https://github.com/z88dk/z88dk/blob/master/libsrc/_DEVELOPMENT/target/yaz180/device/8255/8255_defines.asm) and [`ide_defines.asm`](https://github.com/z88dk/z88dk/blob/master/libsrc/_DEVELOPMENT/target/yaz180/driver/ide/ide_defines.asm) into one file.

But, since the initial hardware pin definitions in the notional `config_ide.m4` that depend on information sourced from [`config_target.m4`](https://github.com/z88dk/z88dk/blob/master/libsrc/_DEVELOPMENT/target/yaz180/config/config_target.m4#L44), I couldn't find a way to make it work.
Simple assignment didn't work. `eval()` didn't work.

I could repeat the information contained in `config_target.m4`, but then there wouldn't be a "single source of truth".

Don't know what to do next. Sorry.

Also, there is no need for any of the contents of `config_ide.m4` to appear in C space. There is no C interface requiring their availability. The [`diskio.h`](https://github.com/z88dk/z88dk/blob/master/include/_DEVELOPMENT/proto/arch/yaz180/diskio.h) file makes the C interface available, and the [IDE asm](https://github.com/z88dk/z88dk/tree/master/libsrc/_DEVELOPMENT/target/yaz180/driver/ide) commands (for other file systems, potentially) also doesn't require the user to have any of the defines.

aralbrec:
The configs are consolidated in one place - target/rc2014/config.m4.  As long as the new configs are included after target's you will have access to the symbols.

Even though they're not likely to be needed, I still think it's a good idea to make all symbols reachable via c.  As they are, with leading underscores, they conceptually stay in system space but you never know if someone finds a reason to look at the values in the future.

feilipu:
I have adjusted the peripheral `config_*.m4` files for yaz180 target first, because it has a lower impact, #414.

But there is an issue that I haven't been able to resolve with the new `config_ide.m4` file, where it doesn't convert the defines to their related values #415. Somehow m4 is not processing the file correctly.

Once this is resolved, then I'll do the same process here.

PR #414, opened 2017-09-27
yaz180 - generalise peripheral config_*.m4 files

feilipu:
Extract the generic configuration of hardware from the `config_target.m4` file and move to device specific files, leaving the platform specific base addresses as they were.
 - `config_am9511a.m4`
 - `config_82c55.m4`
 - `config_pca9665.m4`

Adding a [second level of abstraction](https://github.com/feilipu/z88dk/commit/75b861044023915bb853011628096074ce00d3d0#diff-bfc77c503460639198de07074c09d141R49) via [`config_ide.m4`](https://github.com/feilipu/z88dk/blob/75b861044023915bb853011628096074ce00d3d0/libsrc/_DEVELOPMENT/target/yaz180/config/config_ide.m4)
 - doesn't correctly associate the m4 defines, and
 - creates an core dump, raised as issue #415.


aralbrec:
Quoting issues should be sorted but let's put it in to keep the train rolling.

PR #428, opened 2017-09-30
yaz180 - i2c  wip1 & basic interrupt defines

feilipu:
Adding the drivers for the PCA9665 I2C interface devices.
Probably 3 stages in total.

Attaching interrupts is easier if the Nascom Basic interrupt table locations are available to the `basic` and `basic_dcio` subtypes, where the page zero interrupt definitions are not used.

Adjusted the rc2014 8255 base address. The standard location will be 0x20.


aralbrec:
Is there a missing config_basic.m4 in the yaz180 target?

feilipu:
Yes. Found it, thanks.

PR #432, opened 2017-10-02
Fix #409: library make failing on mingw

### LINKED ISSUE #409: library make failing on mingw
Recent changes make my (very personal and bugged) mingw installation fail:

--- Building Z88 Library ---

C:/MinGW/bin/make gfxdeps
make[1]: Entering directory 'c:/z88dk/libsrc'
make[1]: Leaving directory 'c:/z88dk/libsrc'
C:/MinGW/bin/make -C ctype
make[1]: Entering directory 'c:/z88dk/libsrc/ctype'
process_begin: CreateProcess(NULL, z88dk-lstmanip -i c:/z88dk/libsrc//_DEVELOPMENT//ctype/ctype_sccz80.lst -p c:/z88dk/libsrc//_DEVELOPMENT/ -t, ...) failed.
make[1]: Makefile:14: pipe: No error
mkdir: missing operand
Try `mkdir --help' for more information.
make[1]: *** [Makefile:72: dirs] Error 1
make[1]: Leaving directory 'c:/z88dk/libsrc/ctype'
make: *** [Makefile:198: z88_clib.lib] Error 2
bash-3.1$

pauloscustodio:
Problems found and fixed:

1) MinGW transforms the input command line arguments, inserting the full path name of the mingw home
before the given command line argument, i.e. the argument /home/paulo/git/z88dk-msys/libsrc/_DEVELOPMENT/
to lstmanip is received in argv[] as c:/MinGW/msys/1.0/home/paulo/git/z88dk-msys/libsrc/_DEVELOPMENT/

The result is that the $(subts...) in the makefiles that replace the library root NEWLIB_DIRECTORY by obj/z80/
fail to find the library root and the result paths are wrong.

Solved by a very hughly hack in lstmanip - filtering the output and removing what seems to be a MinGW root directory
prefix.

2) In the MinGW build sometimes a double slash is inserted between directories, causing the same $(subst...) as
above to fail to find the library root and the result paths are wrong.

Fixed by filtering the output of lstmanip, removing duplicated slashes, and the same in Make.config.

3) Some makefines did not have the trailing slash in obj/z80, causing in MinGW paths like obj/z80ctype, which then
failed to assemble. Fixed.


suborb:
Hmmm, directories are determined as follows:

OUTPUT_DIRECTORY :=  $(dir $(abspath $(lastword $(MAKEFILE_LIST))))
NEWLIB_DIRECTORY := $(OUTPUT_DIRECTORY)/_DEVELOPMENT/

Maybe $(abspath should be replaced with $(realpath ?

Failing that I'd prefer an option passed in to lstmanip which defines a prefix to strip rather than hardcoding it within the binary.

The lstmanip was a quick solution to sharing the source files, maybe I need to come up with another way of doing it to avoid all these windows problems.

pauloscustodio:
> Maybe $(abspath should be replaced with $(realpath ?

Does not work. The first returns /home/paulo/git/z88dk-msys/libsrc/, the second  /usr/home/paulo/git/z88dk-msys/libsrc/ and they are still different from the one lstmanip receives.

I'll try the command line option to lstmanip

pauloscustodio:
> Failing that I'd prefer an option passed in to lstmanip which defines a prefix to strip rather than hardcoding it within the binary.

That works. I've added a ```-r /``` to all command line invocations of z88dk-lstmanip, and the optarg received is ```c:/MinGW/msys/1.0```. I then use this to remove the prefix.

PR #454, opened 2017-10-12
Feature/z80asm glob

pauloscustodio:


suborb:
Sorry, I've only just got a chance to try this out now.

I've still got a problem:

```
z80asm -xx.lib "obj/r2k/**/*"
Error: file 'obj/r2k/string/c' is not a regular file
Error: file 'obj/r2k/string/z80' is not a regular file
Error: file 'obj/r2k/string/c/sccz80' is not a regular file
3 errors occurred during assembly
```

suborb:
Before this is merged I'll need to update the nightly build script to pull the submodules.

Also, do we need to anything with the Travis config to handle submodules? - Forking uthash is great, be nice to have fewer copies of it about the place!

pauloscustodio:
> Also, do we need to anything with the Travis config to handle submodules?

No, Travis is building with submodules just fine.

> Forking uthash is great, be nice to have fewer copies of it about the place!

Yes, I'm just getting rid of my personal copy in z80asm. It will live in z88dk/ext/uthash.

pauloscustodio:
> Ah, that's by design! Think it might be best to do this - we're in a glob so I think I'd expect it to skip over files it can't deal with.
I'll remove the check.

pauloscustodio:
@suborb, the error_not_regular_file() is gone, please let me know when to merge.
Thanks.

suborb:
That's working as expected now, thanks. Feel free to merge - I updated the build script to pull submodules.

Just have to remember to quote the argument if globs are used on the command line or the shell expands it for you and you end up with the error again. However, this is expected.

PR #458, opened 2017-10-16
yaz180 - initiate subtype=yabios

feilipu:
First steps to building the yabios subtype for yaz180 (converted the current asci0 subtype).

The yabios specific code will be added to a ["third party library"](https://github.com/feilipu/yaz180/tree/master/yabios), at least until it is mature, but the device driver and stdio drivers should exist (as they already do) in z88dk.

I've found two problems with the current ROM (yabios) model subtype.
1. when using DEFS, memory is not allocated. Specifically the [Rx Buffer](https://github.com/z88dk/z88dk/compare/master...feilipu:master#diff-e179f15e091174a770a8c8420b1fc074R8) and the [Tx Buffer](https://github.com/z88dk/z88dk/compare/master...feilipu:master#diff-c066f0b9db35c991024c7ad1af6652d0R8) allocate 0 bytes. UNLESS, they are defined with DEFM.
2. it needs to have the dcio terminal drivers applied. I could copy/paste from the `basic_dcio` subtype, but I'm not sure that's the whole answer.

Tested using the `password.c` test file with this command line, the `.map` files shows 0 bytes provided for the Rx and Tx buffers.
```
zcc +yaz180 -v -subtype=yabios -SO3 --list -m -clib=sdcc_iy --max-allocs-per-node200000 password.c -o password -create-app
```


aralbrec:
```
SECTION data_align_256

PUBLIC asci0RxBuffer

EXTERN __ASCI0_RX_SIZE

asci0RxBuffer:   defs __ASCI0_RX_SIZE   ; Space for the Rx Buffer
;asci0RxBuffer:   defm __ASCI0_RX_SIZE   ; Space for the Rx Buffer
```

This is a assemble time versus link time issue.  `__ASCI0_RX_SIZE ` is only known at link time but the assembler needs to fill in the `defs` before link time so it uses the assemble time value of `__ASCI0_RX_SIZE` which is 0.

To use that symbol in this way you'll have to include the private config file which defines the symbol at assemble time instead:

```
INCLUDE "config_private.inc"   ;; includes "config_yaz180_private.inc"

SECTION data_align_256

PUBLIC asci0RxBuffer

asci0RxBuffer:   defs __ASCI0_RX_SIZE   ; Space for the Rx Buffer
```

feilipu:
Thanks.
I knew it had to be simple.
But, I couldn't see the answer after staring at it for far too long.

feilipu:
I am having quite a lot of trouble understanding the terminal driver structure.

I understand the terminal driver is necessary to get "line editing", which is useful for a bios CLI.

Is there a reference page for how it works, and how to modify it?
The [terminal flags](https://github.com/z88dk/z88dk/blob/master/libsrc/_DEVELOPMENT/target/yaz180/yaz180_rules.inc#L10) are a mystery too.

Is there a defined (recommended) way to cut underneath the terminal driver system to get fast and direct access to raw characters (for a `load_hex` or `load_bin` program for example) when required?

Thanks in advance.

aralbrec:
It's an easy one to miss!

> it needs to have the dcio terminal drivers applied. I could copy/paste from the basic_dcio subtype, but I'm not sure that's the whole answer.

Do you mean for the asci0 drivers in startup=0?  Any common code can be shared; I think it's only `rc_01_output_basic_dcio_oterm_msg_putc` that might be different.  You could probably write a driver that only checks for OTERM_MSG_PUTC and forwards the rest to the basic_dcio driver.  Or you could write a base driver that sends a message you make up instead of `rst 0x08` at the end of `rc_01_output_basic_dcio_oterm_msg_putc` so that everything that needs that functionality can subclass it and deal with the special message to send the char to the right place and forward everything else to the base class.

If you have a lot of peripherals that might want to behave like a terminal you could do that to reduce code size.



 You could even subclass one level further by sending a message you make up instead of `rst 0x08` at the end of that function then you can share all the code but there is a level in front that understands that message only and sends the char to the right place.

aralbrec:
Input terminal flags:

```
   IFNDEF CRT_ITERM_TERMINAL_FLAGS
      defc CRT_ITERM_TERMINAL_FLAGS = 0x03b0
   ENDIF

bits 15..10  reserved (includes some state for driver base class)
bit 9 = 1 to enable printed cursor
bit 8 = 1 to enable crlf conversion (will be filtered to generate lf only)
bit 7 = 1 to echo typed input characters
bit 6 = 1 to enable password mode (input printed as *)
bit 5 = 1 to enable line mode (input is edited before being delivered)
bit 4 = 1 to enable cook mode (input characters are interpretted in some way)
bit 3 = 1 for caps lock
bits 2..0  reserved (device state)
```

Output terminal flags:

```
   IFNDEF CRT_OTERM_TERMINAL_FLAGS
      defc CRT_OTERM_TERMINAL_FLAGS = 0x2370
   ENDIF

bit 15..14 reserved (one bit indicates input terminal is reading a line)
bit 13 = 1 to clear or 0 to wrap when output terminal is in page mode
(when screen is full, clear screen or not before moving cursor to top left)
bit 9 = 1 to enable signal bell (this is generated by the input terminal)
bit 8 = 1 to enable the ctrl-g bell
bit 7 = 1 to choose page mode rather than scroll mode
bit 6 = 1 to pause when terminal is filled with text
bit 5 = 1 to enable cook mode (the terminal will interpret some ascii codes)
bit 4 = 1 to enable crlf conversion (z88dk will send \n, driver decides what this gets turned into)
bits 2..0 = reserved
```

feilipu:
Thanks, that makes it clearer on the terminal flags usage.

Hmm. TBH, I'm not sure what to think next. My understanding of the terminal driver code is limited.

There are two (ASCI0 and ASCI1) devices that are independent. I'm intending that their interrupt service routines lives in CA1 memory. And the lowest level of their `_msg_getc` routines have to be in CA1 too, otherwise other CA1 colocated code like `load_hex` and `load_bin` need to be buffer and DMAC driven (difficult), rather than just simple straight write to BANKn destination.

Aside from that, the rest of the character/terminal driver code can live in BANK0, and be reachable from a system `RST+DEFW` call.

So I guess my question is whether the character I/O drivers are sufficient for [`_getline`](https://github.com/z88dk/z88dk/blob/f93b5ddc7141fb14b107df82ed7285439400092c/libsrc/_DEVELOPMENT/stdio/z80/asm_getdelim_unlocked.asm) for a simple CLI implementation?

And if so, then I can leave the terminal driver out of scope, which I think is preferable (because I don't understand them fully).

aralbrec:
> I am having quite a lot of trouble understanding the terminal driver structure.
> Is there a reference page for how it works, and how to modify it?

There is some information in this thread:
https://www.z88dk.org/forum/viewtopic.php?pid=15549#p15549

What happens is that stdio sends a small set of messages to the driver as documented here:
https://github.com/z88dk/z88dk/blob/master/libsrc/_DEVELOPMENT/stdio/readme.txt#L90

So output terminals would see PUTC and WRIT while input terminals would see GETC, EATC and READ.  Plus a few others as documented.  Messages are just A=message id plus the other registers set to parameter values.

When the driver's function runs (eg [zx_01_output_char_32.asm](https://github.com/z88dk/z88dk/blob/master/libsrc/_DEVELOPMENT/target/zx/driver/terminal/zx_01_output_char_32.asm#L144) which I am using as example because it has a more complicated data structure) it has `ix` pointing at its driver data structure as documented there.  Then the state of the driver can be seen by looking at values stored with respect to `ix`.  `ix+0, ix+1, ix+2` also point at a z80 `jp zx_01_output_char_32` so that the driver and its base classes can send messages to itself using `call l_jpix` ([l_jpix](https://github.com/z88dk/z88dk/blob/master/libsrc/_DEVELOPMENT/l/z80/l_jpix.asm)).

So a bare driver would implement the few messages that stdio sends as detailed above.  However implementing editing facilities and all that requires a lot more work so subclassing happens.  What this means is the library provides code that consumes stdio messages and invents a whole new set that make it easier to do things like terminals.

If you take a closer look at one of the yaz180 terminals [rc_01_output_basic_dcio](https://github.com/z88dk/z88dk/blob/master/libsrc/_DEVELOPMENT/target/yaz180/driver/terminal/rc_01_output_basic_dcio.asm#L10) you will see that it subclasses a library driver [CONSOLE_01_OUTPUT_TERMINAL](https://github.com/z88dk/z88dk/blob/master/libsrc/_DEVELOPMENT/drivers/terminal/console_01/output/console_01_output_terminal.asm) and if you look at that code you will see the library is dealing with all the messages sent by stdio.  You can see how it implements two important messages [WRIT and PUTC](https://github.com/z88dk/z88dk/tree/master/libsrc/_DEVELOPMENT/drivers/terminal/console_01/output/console_01_output_terminal).  WRIT and PUTC can send a lot of chars in one message so the library driver is breaking that down to one char at a time via a new message it invents `OTERM_MSG_PUTC` and it's this message your code must intercept to send to an attached device.  This driver is very simple so there's not much too it but if you look at the input driver for terminals with line editing, there are a lot of messages it can generate.  Your driver gets first crack at the messages and can intercept them to change behaviour.

Anyway I hope that's good for a first look.

> Is there a defined (recommended) way to cut underneath the terminal driver system to get fast and direct access to raw characters (for a load_hex or load_bin program for example) when required?

You can `read` and `write` via file descriptors to avoid the passage through stdio but both stdio and file descriptors generate the same set of messages: PUTC, WRIT, etc.  These messages are intended to read or write entire buffers at a time but the library drivers are breaking them down to character at a time for ease of use.  If you want entire buffers, you'll have to implement PUTC / WRIT yourself.  The drivers also have a concept of binary mode that you can use to avoid interpretation of characters but it will require you to write a little more code.  You're also free to invent your own messages to implement special behaviour.  You could invent new functions that take FILE* or file descriptors as arguments and generate these special messages to do special tasks.  Drivers that don't understand these messages (nothing intercepts the message) will usually have the message ignored and the FILE* put in an error state.

aralbrec:
> So I guess my question is whether the character I/O drivers are sufficient for _getline for a simple CLI implementation?

The entire library is available if stdio's messages are implemented, which they are for anything subclassing the library drivers.

getline depends on malloc and realloc to resize its buffer.  The heap that malloc/realloc uses can be changed at any time so you could have a bit of system memory set aside for heap and change to that whenever the system/user interface is crossed (the "heap" is just a pointer to a block of ram set aside).  So there are a few details to keep in mind for sure.

getline may not be necessary for text terminal input though -- the text terminal is limited to an edit buffer of certain size so you can never get an input line longer than that and the line will not be returned until the user presses enter so you will always have an entire line of known max size read from an input terminal.  getline is more useful for disk io or reading from raw character devices.

aralbrec:
> And if so, then I can leave the terminal driver out of scope, which I think is preferable (because I don't understand them fully).

If I undestand right, you want to fill / empty buffers independently of the drivers.  There's no problem with that and yes the stdio code can be out of scope while that is happening.   The drivers will have to be in scope and be able to access those buffers when stdio code is running.

Issue #425, opened 2017-09-29
appmake: banked memory issues


aralbrec:
Some things that have come up:

1. Sections "BANK_00" through "BANK_07" (numbers with leading zeroes) cannot be used anymore.

The bank numbers used to be two-digit hex and now they are decimal as in "BANK_0" through "BANK_7". I thought compatibility was kept because the section names are scanned as "BANK_%d" to identify number but the incompatibility happens at link time where the section names have no meaning and are treated as strings.  So BANK_00 is placed in a separate binary outside the memory map because it is not the same as BANK_0.

Section names without the leading zeroes must be used from now on.

2. Stuff placed in sections not part of the memory map and without org gets appended to the last defined section.

From the first point, someone was using a section "BANK_05" which was not being placed in BANK_5 but rather in the last defined section in the memory map "DIV_31".  This was a mysterious error and there is no easy way to find out what was placed in DIV_31 to know where the data was coming from.  The map file showed nothing assigned to section "DIV_31" (that data was in BANK_05) so you couldn't find out what was in there.

**Is it possible to add to the map file information on what the parent section is**, as in the section holding the org?  This way the map file could have shown that the data was in BANK_05 but its parent was DIV_31 so it would be known BANK_05 would be part of the DIV_31 binary.

I'm also adding section "UNASSIGNED" to the end of the memory map for all targets so these things are made part of an invalid binary rather than part of a legit binary in the wrong place.  Appmake's banking code will remove section UNASSIGNED from the memory model and emit a warning.

3. It used to be that a user defined section with its own org was output as a separate binary.

This is still the case.  But appmake's banked model treats sections not part of a bank space as belonging to the main bank.  So that means any processing by appmake involving banked memory will put that user defined section into the main bank too.

If that's not what is wanted, the section should be removed from appmake's memory model by adding `--exclude-sections "space_separated_section_names_here"` to the appmake invoke.  If invoking from zcc, this would be added with `-Cz'--exclude-sections "..."'`.   By excluding particular sections, those binaries will be ignored by appmake and will be left in the user directory.

4. There's a new requirement - being able to assemble to a specific address but then being able to place the resulting binary someplace else.

The specific use-case is someone assembling for org 0x2000 but wanting to store the binary someplace else so that it can be memcopied to the right address on demand.

The construction of output binaries and knowledge of memory maps is in appmake so this is a requirement for appmake.

[...]

PR #462, opened 2017-10-17
yaz180 - yabios wip2

feilipu:
Work in progress for the yabios for yaz180.

I have encountered the `PHASE` and `DEPHASE` issue #425, in creating stubs for COMMON AREA 1 system functions. The system functions are stored in Flash low memory on boot, but must be copied to high memory for use.

I need [system functions](https://github.com/z88dk/z88dk/compare/master...feilipu:master#diff-014bc81d1980d1c333a2f849cc321eef) to [reside in high memory from 0xF700](https://github.com/z88dk/z88dk/compare/master...feilipu:master#diff-17c93d69a1dcf74ac515dddc7cba0e1aR27), and have added `PHASE` and `DEPHASE` statements to that effect. But, I note that they are all assembled and linked to the __same phase__ or origin.

Also of note is that it doesn't seem possible to [add a reference to `PHASE` in the usual way](https://github.com/z88dk/z88dk/compare/master...feilipu:master#diff-014bc81d1980d1c333a2f849cc321eefR7), as using `EXTERN` is needed or there is a compile error, but this leaves the `PHASE` with no value. A concrete value seems to be required. Am I missing something here?


aralbrec:
> I need system functions to reside in high memory from 0xF700, and have added PHASE and DEPHASE statements to that effect. But, I note that they are all assembled and linked to the same phase or origin.

Yes this only really works painlessly if the PHASEd block exists entirely in a single file.  Once phase can be applied to a section so that independent pieces can be packed into the section by the linker, the problems will go away.

> Also of note is that it doesn't seem possible to add a reference to PHASE in the usual way, as using EXTERN is needed or there is a compile error, but this leaves the PHASE with no value. A concrete value seems to be required. Am I missing something here?

PHASE is an assemble time quantity because of the way it's implemented.  It's simply temporarily changing the current offset into the section to another value (the assemble-time ASMPC value).  So PHASE can't be set by a link time EXTERN value.  You'd have to use an assemble-time value using the INCLUDE thing mentioned for the previous `defs` issue.

feilipu:
Ok, in the interim (until `PHASE` is enhanced) I'll pack all the system data and system functions into one "system driver" file.  Noting that it is not the usual style, and quite a bit ugly.

but, that will also make it simpler to copy them up into `section COMMON_AREA_1` somewhere in the preamble code.

Issue #395, opened 2017-09-18
Latest snapshot breaks CPC: blank screen at startup


Fabrizio-Caruso:
The code works but nothing is shown on the screen.

I  have tested it with both color and no color version of my game for the CPC.

zx70:
It it resolved now?

zx70:
I'm keeping the issue open because the snakes.c game is not visible.. colour problems.?

zx70:
First small portion of cpcrslib is now merged in the current classic library.  Using the standard library edition is still possible,  the only conflict is in cpc.h, which had very few functions useful in conjunction with cpcrslib.  The high resolution graphics is now working at a full 640x200 monochrome resolution with an almost full coverage of the function list (it was 256x200 sometimes, 320x200 for other rom based stuff with limited coverage of the whole library).  Cp/m variant and greylib will require a serious reassessment,  but the new stuff worths the changes.

zx70:
Fonts and non-tile-based sprites are now in.   In the latest commit I'm also correcting a bug on the font color settings, which was also compromising the code stability.

zx70:
640x200 monochrome mode is now fully operational and respectably fast, I think *all* the graphics demos will be in working order by tomorrow morning.   
The CPCRSLIB is not totally included yet, the tile mode graphics have been improved in the years and there are two different, slightly incompatible versions around.  The code structure for 'tiles' could be kicked approaching it like we did with the ANSI VT options, I'm taking some time to think at it.
By the way, the current CPCRSLIB imported portion is now in a brand new, shiny shape.

zx70:
lots of open points already,  the original claim is solved and the library is now reasonably extended and cleaned..     I still need to check the status of the cp/m version of the libraries, so I keep the issue open

zx70:
I recently removed the 'startup=2' option  from the default cpc configuration. I couldn't succeed in running the graphics programs otherwise.    please advise if you find problems after this

Issue #38, opened 2017-01-22
zsdcc: delayed error reporting for function prototypes missing terminating semicolon (K&R issue)


aralbrec:
This is from a bug report in the forums before the switch to git.

The complaint was in code like this:

```
#ifndef __MYFILE_H__
#define __MYFILE_H__

void just_some_function(int some_param)   // <=== forgot ;

#endif    //__MYFILE_H__
```
and

```
#include "MyFile.h"

void some_other_function(int some_param)
{                                                               // <=== syntax error report on '{'
   // blabla
}
```

The missing semicolon is not detected in "MyFile.h" but shows up as an error inside "some_other_function()" in another file.  This caused a great deal of extra debugging time for what is a trivial mistake.

The compiler saw this after preprocessing:

```
void just_some_function(int some_param)   // <=== forgot ;

void some_other_function(int some_param)
{                                                               // <=== syntax error report on '{'
   // blabla
}
```

The reason why sdcc does not complain about the missing semicolon in the right place is that I believe it is a reflection of the halfway support sdcc has for K&R function syntax.  sdcc does not support K&R currently but that is something on its to-do list.  I think the scanner portion of sdcc may recognize it but the parser / code generator portion does not so the recognition of an error is delayed.

K&R syntax allows types of function parameters to be declared after the function header itself like so:

```
void func(a,b,c)
int a;
char b;
long c;
{
   ...
}
```

and if you compare this to the original problem file you will see the compiler seems unable to determine there was an error until the declaration following the prototype with missing semicolon is completely parsed.

The situation is a little worse than that.  In this case:

```
void just_some_function(int some_param)   // <=== forgot ;

int a;

void some_other_function(int some_param)
{                                                               // <=== syntax error report on '{'
   // blabla
}
```

sdcc is unable to determine there is an error until it's inside some_other_function().  It should be able to report an error at "int a;".  This is an error under K&R because "a" does not match the name of any parameter given to just_some_function().

I believe the reason it gets this far is a partial recognition of K&R syntax inside sdcc.

This is not an error but may cause grief in debugging and is something that must be dealt with by the sdcc team.

PR #466, opened 2017-10-19
HRG and cpcsrlib import in progress

zx70:


zx70:
#395

cpcitor:
# cpcrslib import?

Hello. I come here on trying to understand your comment https://github.com/cpcitor/cpc-dev-tool-chain/issues/10#issuecomment-336743628

No trace of cpcrslib revealed either in your fork or in original z88dk by `git log --all -Scpcrslib -p` or `git log --all -Gcpcrslib -p` .

I'm not sure about what you mean about "cpcrslib import".

# cpcrslib location

Anyway, the original cpcrslib project by Raúl Simarro is on https://sourceforge.net/projects/cpcrslib/ .
Since latest cpcrslib release was incomplete (no portable build structure, missing important source files), I cleaned it up and put the cleaned up version on https://github.com/cpcitor/cpcrslib .

# cpcrslib imported in cpc-dev-tool-chain

[Cleaned up cpcrslib](https://github.com/cpcitor/cpcrslib) works very well in https://github.com/cpcitor/cpc-dev-tool-chain/

On a Linux machine having `wget` `make` `patch` `gcc` `bzip2` `g++` `bison` `flex` `libboost-dev`
the command line below will produce disk images ready to run.

```
git clone https://github.com/cpcitor/cpc-dev-tool-chain/
cd cpc-dev-tool-chain/
make cpcrslib-all
```

You can do `make -j8 cpcrslib-all` for extra speed boost on a multi-core machine.

```
ls -1b cpclib/cpcrslib/cpcrslib-master/examples/*/*dsk

cpclib/cpcrslib/cpcrslib-master/examples/001_-_8x8_Chars___Small_Scroll/0018x8Ch.dsk
cpclib/cpcrslib/cpcrslib-master/examples/003_-_Small_Sprite_Demo__Tile_Map_/003Small.dsk
cpclib/cpcrslib/cpcrslib-master/examples/004_-_Music,_SFX___Keyboard/004Music.dsk
cpclib/cpcrslib/cpcrslib-master/examples/006_-_UnExo/006UnExo.dsk
cpclib/cpcrslib/cpcrslib-master/examples/007_-_Uncrunch/007Uncru.dsk
cpclib/cpcrslib/cpcrslib-master/examples/009_-_Keyboard/009Keybo.dsk
cpclib/cpcrslib/cpcrslib-master/examples/010_-_Sprites/010Sprit.dsk
cpclib/cpcrslib/cpcrslib-master/examples/011_-_Small_Sprite_Demo__Tile_Map__-_SCROLL/011Small.dsk

```

# What now?

If you have questions about cpcrslib and its possible adaptation, I'd be happy to help.

Please open an issue on Raúl's original project https://sourceforge.net/projects/cpcrslib/ or the portable, cleaned up version on https://github.com/cpcitor/cpcrslib .

zx70:
It is great to have your news!
I got the request to look into cpcrslib from a z88dk user complaining about most of the library being incompatible with the current z88dk distribution, so I'm cleaning it and progressively testing the demos and importing the evolved code directly into z88dk.
To discover you already have something is fantastic, I'm downloading your tool and I'll try to understand how to correct my current approach.

cpcitor:
Great! :-)

## How I see z88dk (might be nonsense, please tell)

What I understood so far is that z88dk aims at providing a common programming environment allowing to write some "portable" code that can run on many Z80-based targets.

I see it like a "Qt for 8-bit", in that it offers a lot of code, from low-level up to high-level stuff (like [3D vector computation](https://www.z88dk.org/wiki/doku.php?id=library:lib3d)  even a [A* Search Algorithm](https://www.z88dk.org/wiki/doku.php?id=library:algorithm)  8-).

I'm under the impression that in this spirit of a full environment, z88dk environment "wants" to control the Z80 interrupts, which causes problems on any platform where an OS already handles (and needs) interrupts.

## Problem with current implementation

Reading again https://github.com/z88dk/z88dk/blob/master/lib/cpc_crt0.asm I feel like z88dk changes the interrupt handler in RAM, but since calling firmware enables ROM at that address, z88dk ISR is no longer called (or worse), until control is put back to RAM.

Let me read https://www.z88dk.org/wiki/doku.php?id=library:interrupts

I see there `M_BEGIN_ISR` and siblings and `Generic ISR`.

What worries me reading `cpc_crt0.asm` is that this ROM/firmware interposer adds more burden on each firmware call, as if it was not slow enough.

Plus it change the calling convention for **all code** already written on the CPC, which is kind of the ultimate breaking change. ;-)

## Alternative proposition for the CPC

Perhaps a "hands-off" approach would be lighter.  Practically, instead of overwriting the 16-bit word at address at 0x39, the CPC-with-full-firmware target would register z88dk ISR using CPC firmware's "KL NEW FRAME FLY" (50HZ), "KL NEW FAST TICKER" (300Hz), or "KL NEW TICKER" (300/n Hz). These are very similar to the concept of "Generic ISR" (though z88dk does not need to know that, it comes downstream in the interrupt processing).  By the way, Amstrad CPC uses interrupt mode 1, not 0 or 2.

In this case, the part of the z88dk ISR that saves registers is not useful so it could be shaved away to save some cycles.

## Actually useful for more platforms?

I would be surprised if only the Amstrad CPC had interrupt handling in its OS design.

Thus, this CPC-specific change can be turned into a general feature.  Instead of unconditionally handling interrupts, z88dk could offer differing implementations of `M_BEGIN_ISR` and `im2_InstallISR` depending on platform.

On bare metal platforms, current implementation would take full control of the Z80 interrupts (like current code does).  This could still be the preferred implementation for some projects that target the CPC while fully ditching firmware.

On CPC-with-full-firmware platform:

* `M_BEGIN_ISR` would just expand to a regular C function declaration.
* `im2_InstallISR` would prepare parameters and call "KL NEW FAST TICKER" (or some sibling)
* `im2_CreateGenericISR` would probably need minor adjustment (only shave the reg saving/restoring code), the hook mechanism would run unmodified
* similar for other calls (cleanup, etc).

## Conclusion

Then :

* no heavy rom interposer code fiddling with interrupt vectors at each firmware call.
* no need to change firmware calling convention for the whole world.
* actually simpler z88dk implementation on that platform (less code to do because target firmware already does the heavy lifting)
* not a platform-specific hack but a cleaner design applicable to all z88dk target.

Perhaps I'm totally misguided?
What do you think?

suborb:
> I'm under the impression that in this spirit of a full environment, z88dk environment "wants" to control the Z80 interrupts, which causes problems on any platform where an OS already handles (and needs) interrupts.

Actually, no we don't want to do anything interrupts by default in z88dk. It's just that we use the alternate set and one of the index registers (ix). Unfortunately I think the CPC firmware reserves use of them.

So we end up having to stick an interstitial interrupt handler in there so the interrupt doesn't crash, and interpose the firmware call to restore the affected registers before calling in.

I don't think the im2 library has been used on the CPC (it's mainly for the ZX) but if it can be adjusted to work with the CPC firmware then that's great.

aralbrec:
The im2 library is for any target.  But as dom says, the issue with the cpc is its firmware does not allow modification of the exx registers.  z88dk's libraries can use all of them so this means if the firmware is to be called, either by call or interrupt, the register state expected by the firmware has to be restored before entry and then on return the register state expected by z88dk has to be restored.  This is the purpose of the interposer in the interrupt routine and in front of firmware calls.  Without this the machine will crash.

You don't have to use the interrupt routine you are looking at - you can replace it with your own.  But then you can't jump into the firmware without doing the same thing first yourself.

cpcitor:
Thanks a lot @suborb for this clarification.

## Traditional wisdom

While IX is okay for use on the CPC, the alternate register set is reserved by the firmware.
So, the traditional wisdom on the CPC is "_either don't use alternate registers, or give up on using the firmware_".

I understand better the benefit of the ROM interposer in the case one absolutely wants ability to use alternate registers freely, whatever the cost in constraints and performance.

## Needed?

I see that `noaltreg` option was introduced recently https://github.com/z88dk/z88dk/blob/master/src/sccz80/main.c#L113

I had no alternate registers issue when I wrote cpc-dev-tool-chain, with two possible backends:

* z88dk (1.10 at the time, now 1.99B) with makefile rule `zcc +cpc -lndos -create-app $< -o $@.cpc`.
* sdcc with rule `sdcc -mz80 --allow-unsafe-read --std-sdcc99` and link rule `sdcc -mz80 --no-std-crt0 -Wl-u --code-loc 0xsomevalue --data-loc 0`

I just updated https://github.com/cpcitor/cpc-dev-tool-chain to use z88dk 1.99B.

## Desiderata

* 1 freedom to use alternate registers (else only use libraries that have noaltreg version)
* 2 compatibility with existing code
* 3 "direct" access to firmware

## Enumerating combinations

It seems you can't have all three but can pick any two.

* pure 2+3 is traditional wisdom.

What you offered provided 1, but not complete 2 nor complete 3.

* pure 1+3 would be like requiring *all* code to conform to new pattern, putting the burden of saving registers etc. More than what you requested.

One combination remains:

* pure 1+2 isn't exactly what you offered, you did not push the logic to the conclusion.

## Another solution?

More explicitly, on the CPC, calls to firmware are implemented with RST pointers.

Address range BB00-BC76 and BC9E-DB61 for example use the byte pattern CF xx xx where CF encode Z80 instruction RST 08 (KL_LO_JUMP).
BC77-BC9D use DF( RST 18, KL_FAR_CALL_SP)
BD61-BDC1 use EF (RST 28, KL_FIRM_JUMP).

To make the ROM interposer compatible with existing code (2) while offering freedom to use alternate registers (1), have you considered replacing the corresponding addresses (09-0A, 19-1A, 29-24) in RST table in low memory? Then user code can still to `CALL #BD19` or other address and work without modification.

You get compatibility with all libraries, all existing code, at the cost of changing more RST table entries.

What do you think?

aralbrec:
> I see that noaltreg option was introduced recently https://github.com/z88dk/z88dk/blob/master/src/sccz80/main.c#L113
> 
> I had no alternate registers issue when I wrote cpc-dev-tool-chain, with two possible backends:
> 
>     z88dk (1.10 at the time, now 1.99B) with makefile rule zcc +cpc -lndos -create-app $< -o $@.cpc.
>     sdcc with rule sdcc -mz80 --allow-unsafe-read --std-sdcc99 and link rule sdcc -mz80 --no-std-crt0 -Wl-u --code-loc 0xsomevalue --data-loc 0

The `noaltreg` option for sccz80 will affect the c compiler only.  The c library is written in assembly language and makes use of the exx set in some areas because it leads to much faster and smaller code; this part cannot be changed without providing alternatives for those functions.  An example from the classic library is the [32x32->32 multiply](https://github.com/z88dk/z88dk/blob/master/libsrc/z80_crt0s/crt0_long/l_long_mult.asm).  Most functions in classic will not be using the exx set and if you confine yourself to a subset of the standard library you can rely on the exx set not being used but this is not really a solution.  I suspect the test programs you wrote didn't use any of the functions using the exx set.

sdcc is not aware of the exx set and only uses the 8080 registers plus index registers so it has no compatibility issue with the cpc.  Because its c library is written in c, and because the compiler doesn't use the exx set, nothing it produces will conflict with the cpc firmware.

You will get much better code out of zsdcc (z88dk's modified sdcc) than sdcc itself btw.

> I just updated https://github.com/cpcitor/cpc-dev-tool-chain to use z88dk 1.99B.

Good to hear :)  We are several thousand commits beyond 1.99B though so there are significant differences from that version now.

>     1 freedom to use alternate registers (else only use libraries that have noaltreg version)
>     2 compatibility with existing code
>     3 "direct" access to firmware

z88dk distinguishes between compile types with a startup value.  So it's possible to have different compiles using interposer / not using interposer / having interrupt interposer / disabling interrupts, etc.  The user can choose on the compile line what sort of compile he wants so it's possible to meet every scenario.  Each compile type just needs to be enumerated.

> To make the ROM interposer compatible with existing code (2) while offering freedom to use alternate registers (1), have you considered replacing the corresponding addresses (09-0A, 19-1A, 29-24) in RST table in low memory? Then user code can still to CALL #BD19 or other address and work without modification.
> 
> You get compatibility with all libraries, all existing code, at the cost of changing more RST table entries.
> 

That's what should be done for the interposer.  The last time I looked at this, the problem was lack of documentation.  The Amstrad technical reference was available online but it was missing one or two chapters, which happened to be the two most important ones.  So there was only enough information to do the interrupt interposer and understand what had to be done for direct firmware calls.

It's stefano who's currently working on this.

cpcitor:
Thanks @aralbrec your precisions which confirms and clarifies the situation and makes a case for deciding sdcc or zsdcc depending on projects. sdcc already provides a solid foundation (fairly complete standard C support) and generates decent (even often good) code (was much better than old zcc when I compared a few years ago). I'll certainly consider zsdcc for future projects, especially those that intended to work without firmware.

Looks like the interposer trick is actually fine for programs where firmware is seldom used but still necessary. In any case, a solution that works widely even if not optimal is a good safety net and default configuration.

Regarding CPC documentation, I know that a number of resources were written in French (possible observation bias, Google won't tell me if there are good resources in other languages), also many are from old magazines and while they do convey some information they also sometimes are incomplete sometime contains mistakes. 

There are a number of highly knowledgeable people on http://www.cpcwiki.eu/forum/ . They'll be glad to help. 

It might happen that you encounter some specific difficulty and/or you find some resource written in French that seems difficult to figure out (though, certainly a number of CPCWiki forum members can read french), open a public request or [contact me via PM](http://www.cpcwiki.eu/forum/profile/?u=531) and if I can devote some time (sometimes with a delay) I'd be happy to help!

@zx70 Stefano keep going!

zx70:
Thankyou @cpcitor 
as said my first goal is simply to recover (at least) part of the original work by Raúl and provide a working variant for the recent z88dk's classiclib versions on sccz80.


I had a first look at your very clean adaption of the library and I see the following differences:

 - The assembler syntax (z80asm).  Not to be confused with the legacy z80asm versions, it permits to skip many of the original workarounds (undocumented instructions, many "global" declarations not requiring a main module to be forcibly linked, the new support for sdcc).

 - The 'callee' mode for parameter passing.  I'm trying to take benefit of this trick (an invention by @aralbrec) which reduces the code size when a function is used repeatedly in a program.

- The 'tiles' related functions highly optimized and evolved.    I was halfway in the library conversion (introducing callee, etc..), but now this is an open point for me:  should I keep supporting the classic variant or embrace the new one ?   I'm obviously for the second option, but the conditional code and the required parameter tables need to be arranged to z80asm.   I should probably take benefit of the z88dk capabilities and provide a simpler way to set up the tilemap (e.g. introducing something similar to the "ansifont, ansirows, ansicolumns" options to alter a preset default configuration).

Last but not least, I know I should take care of the sdcc version (the parameter passing probably will need to be tuned), but it is a big amount of work already, and I really need to proceed by steps.

cpcitor:
> The last time I looked at this, the problem was lack of documentation. The Amstrad technical reference was available online but it was missing one or two chapters, which happened to be the two most important ones. So there was only enough information to do the interrupt interposer and understand what had to be done for direct firmware calls.

I draw Kevin Thacker's attention on the missing chapters of the official firmware documentation from the Amstrad CPC and, as a result, yesterday he posted several missing chapters, see [SOFT968/SOFT158 PDFs](http://www.cpcwiki.eu/forum/programming/soft968soft158-pdfs/msg151621/?topicseen#msg151621).

Chapter 2 looks interesting. At the end it also mentions patching RST jumpblock to stay compatible with firmware (although for different reasons).

> In particular, if organization 2 is selected it is necessary to patch a program into RAM
at #0038 to catch interrupts and to bank switch back to a more normal organization
(e.g. organization 1) to run the standard interrupt code.

Soft968 section "14.4 The Low Kernel Jumpblock." seems also relevant 

The other chapters are already mentioned on e.g. http://www.cpcwiki.eu/index.php/Soft968:_CPC_464/664/6128_Firmware

Cheers!

zx70:
Wow, the firmware is really nasty in the way it alters the colors of the display and alters a program stability !  I'm strongly tempted to exclude it furtherly.  I thought it was nice to have some flexibility related to the current display mode,  but it is definitely better to have some more graphics tricks firmware independent.

cpcitor:
> Wow, the firmware is really nasty in the way it alters the colors of the display and alters a program stability ! I'm strongly tempted to exclude it furtherly. I thought it was nice to have some flexibility related to the current display mode, but it is definitely better to have some more graphics tricks firmware independent.

It's probably tripped by something not properly restored. Perhaps BC', see below.

**Good news!** I just found the official documentation about using alternate register set while being compatible with firmware. Probably, issue you encountered can be explained by something not properly restored.  Again thanks to Kevin Thacker for all the conversion work: http://cpctech.cpc-live.com/docs/manual/s968ap11.pdf

> Table of contents:
> Appendix XI The Alternate Register Set.
> a. The firmware’s use of the alternate register set.
> b. Simple use of the alternate register set.
> c. Use of the alternate register set with interrupts enabled.

These instructions seem of paramount importance to me:

> When an   interrupt   occurs   the   user’s   alternate   registers   are   saved,   the firmware’s  BC’  and  carry’  are  restored  and  INTERRUPT  ENTRY  is  patched back  to  the  firmware’s  interrupt  routine.  The  latter  is  done  in  case  a  second interrupt occurs whilst processing the events kicked from the interrupt path of the  first  interrupt  (remember  that  the  event  processing  is  performed  with interrupts enabled).

> Note that when INTERRUPT ENTRY is patched it is vital to ensure that the lower ROM is disabled and remains disabled. It is impossible to patch the ROM version of   INTERRUPT   ENTRY!   If   an   interrupt  occurred  whilst  the  lower  ROM  was enabled  then  the  firmware  would  jump  straight  into  its  interrupt  routine  without restoring its alternate registers first.

This means that z88dk must make developers aware not to manually enable ROM in this mode. Some do enable lower ROM to access e.g. ROM-based font.

## Z88dk compile types

> z88dk distinguishes between compile types with a startup value. So it's possible to have different compiles using interposer / not using interposer / having interrupt interposer / disabling interrupts, etc. The user can choose on the compile line what sort of compile he wants so it's possible to meet every scenario. Each compile type just needs to be enumerated.

This seems a good framework! Brings flexibility, not impose choice on developers.

IMHO, there are 3 program styles on CPC:

* Lots of firmware calls (the approach I took on https://github.com/cpcitor/color-flood-for-amstrad-cpc ). Best for non-time-critical stuff. All firmware text, graphics, I/O (keyboard, disc access) primitives can be used and bring the "classic CPC style" with them (font, etc). Generated code size is probably a good optimization metric. The best use case for the interposer.
* Programs that would perform best temporarily disabling firmware, while keeping ability to return to it on exit. Perhaps for libraries extending BASIC with high-performance routines?
* Programs that fully take over the machine, wipe firmware-reserved area in RAM, performing all I/O with custom routines. Looks like demos. 

The first option is the one that benefits from most combinations of interposer/interrupts you mention.

Anything not leveraging the CPC firmware will either:
* imply a lot of CPC-specific reimplementation (and will come with its own style which may not suit every taste), 
* or suffer a "race to the bottom" style typical of highly portable programming environments.

Either case sounds not appealing to people specifically targetting the CPC. 

## Conclusion

* All in all, supporting firmware in your options is indeed a very nice goal.
* It will become a major selling point for z88dk, like: "z88dk produces (1) shorter and (2) faster code than plain SDCC, while (3) supporting many workflows (enable/disable interrupts, etc)."

(I don't know if I'm clear, and might even be wrong. Please complain if needed. )

Side note: the C-firmware-interface I write as part of [cpc-dev-tool-chain](https://github.com/cpcitor/cpc-dev-tool-chain "cpc-dev-tool-chain") might become in the future an interesting piece of that "z88dk supports the CPC very well" selling point. See [coverage statistics](https://cdn.rawgit.com/cpcitor/cpc-dev-tool-chain/master/cpclib/cfwi/coverage.html "coverage statistics").

Keep up the good work!

zx70:
I'm being diverted by other issues (and by the Christmas period)  but I'm very interested in making the mixed bios stuff and direct access work together,  it is something deserving extra study.

cpcitor:
Thanks @zx70 for your follow-up. Of course, no one demands you anything, I'm just expressing support and provide help as I can (while I progress on https://github.com/cpcitor/cpc-dev-tool-chain with https://github.com/cpcitor/color-flood-for-amstrad-cpc in the role of the C-based third party project).

It's really nice that the official firmware manual gives so much detail about how to make a firmware interposer. It is a very strong sign that it can be done reliably. If z88dk's SDCC can indeed leverage the second set of registers to produce smaller and faster Z80 code, then it's a huge win!

Cheers and happy new year!

Issue #452, opened 2017-10-10
New decl parser for sccz80


suborb:
This is in progress in an obviously named branch, list of things to do so I don't lose track:

* [x] Initialisation for static
* [x] Initialisation for auto structs/arrays
* [x] Initialisation for auto
* [x] Anonymous structs
* [x] Far code validation
* [x] BSS output
* [x] Restore typedef
* [x] Validation of callee/fastcall function pointers
* [x] Warnings for pointer conversion for function calls
* [x] K&R functions
* [x] Function pointer calling
* [x] Array parameter passing
* [ ] Pointer to int conversion for function calls (warning)
* [x] Reject names for cast expressions
* [x] Restore sizeof
* [x] Restore offsetof
* [x] Cleanup unsigned handling
* [x] Validation that function definition matches declaration
* [ ] Function decorator to support sdcc-style single byte passing for char (NTH)
* [x] sizeof/offsetof in static initialisation
* [ ] Decide on CARRY/INT as result of conditions
* [x] Verify auto array access
* [x] Implicit int functions
* [x] Repair auto structs
* [x] const data should end up rodata section

suborb:
I'm still working on this, but this now passes all the ticks tests. (Hurrah!) testsuite tests is another matter of course, but it's nice to finally see some working code generation.

Test snippets welcome (those that work with sccz80 at present especially so).

PR #471, opened 2017-10-22
Redo sccz80 declaration parsing

suborb:
This is a chunky change. Inspired by code in 8cc I've rewritten the decl parsing of sccz80. This has affected pretty much the entire codebase of the compiler.

For new features we now get:
- multi-dimensional arrays
- function pointers with parameters.
- some stricter type checking

So far we've lost:
- Validation of CALLEE qualified function pointers
- ~~far support (temporarily broken, I want to restore it)~~

I want to add:
- Support for the sdcc calling convention for chars
- The other items in #452 (mostly non-critical)

At present this is passing all test suites and building the classic library and the classic examples.

*However* I'm bound to have missed some use cases so I'd appreciate scenarios that cause problems.


aralbrec:
There are a number of programs in the newlib examples that can't be compiled with sccz80 that maybe can be compiled now.  I will give those a try.

suborb:
I think the time has come to take the plunge and merge this one.

There will obviously be some cases which I've not covered, but as far as I can tell all of our existing examples work, some extra test programs I use are also correct.

PR #473, opened 2017-10-24
yaz180 - yabios wip4 - instantiate asci1 stdio & add DMAC bit definitions

feilipu:
More housekeeping:

- Added the stdio drivers for asci1 (second serial port on z180). This was done by simple replication of the asci0 stdio drivers. If there is a neater / smarter way to do this, please let me know.

- Added some more DMAC control register bit defines to the z180 CPU crt file. These are useful for any z180 CPU, and could/should be replicated more generally.

- Converted `crt_preamble.asm` to use DMAC0 for moving `COMMON_AREA_1` data and drivers.

Z180 general change:

- moved the `SECTION code_crt_start` out of the vector table setup in `crt_memory_model_z180.inc`, and added a `SECTION code_crt_start` reference to the `crt_page_zero_z180.inc`.

```diff
IF (__crt_org_vector_table = -0x80) || (__crt_org_vector_table = -0x100)
section interrupt_vectors
-section code_crt_start
ENDIF

+section code_crt_start
```
If there are vector tables required, but the Z180 table is just not at `0x0080` or `0x0100`, then the page0 code doesn't get referenced until AFTER the preamble code in `code_crt_init`, if the `section code_crt_start` is not called out.

If the `section code_crt_start` is not called out at the start of page0, then the start code also doesn't get put in the right order.

I think this is a left-over from when vector tables could only be at one of the two above locations.

__FIXME:__ I think this should be fixed for the __Z80__, but I'll leave that for someone else to pull.


aralbrec:
There may be some issues with this and the last commit.  Moving to crt variable instead of config variable for __IO_VECTOR_BASE, eg, means its value is no longer available to the program and it's not possible to move vectors without modifying a bunch of values in the config file.

This change in the page zero code:

```
-   out0 (0x3f),a
+   out0 (ICR),a
```

May not be ok.  The fixed value is deliberate because the user can change ICR in the config but this absolute value is always correct.  I have to look at that again to make sure that's right.

The preamble is meant as either an easy way for the user to replace the page zero code while still having an org 0 or to put a header on code not at address 0 (and therefore the page zero code is not incorporated). 

I'll see if I can have a more thorough look tomorrow.

feilipu:
Thanks.
> Moving to crt variable instead of config variable for __IO_VECTOR_BASE, eg, means its value is no longer available to the program.

This `__IO_VECTOR_BASE` was commented out in `config_cpu.m4` some time ago, and I was just cleaning up. Perhaps it should just stay as a comment?

Sorry, thought that below was just an oversight.
```diff
-   out0 (0x3f),a
+   out0 (ICR),a
```
My tidiness OCD and mistake.

Issue #485, opened 2017-10-30
Create function decorator to specify sdcc calling spec


suborb:
Mainly thinking of the single byte char here. 

Potentially could do something with floats as well.

PR #488, opened 2017-11-01
__z88dk_sdccdecl + definition in for loop initial clause

suborb:
Add support for __z88dk_sdccdecl for sccz80. Issue #485 

This passes chars as a single byte on the stack. Both calling functions and compiling functions with this flag are supported.

Support defining the loop variable inside the for: Issue #489


suborb:
Still need to force __stdc and handle printing out the flags

Issue #493, opened 2017-11-03
sccz80: Support __func__


suborb:
Because it's useful

Issue #491, opened 2017-11-03
sccz80: problems with enum


aralbrec:
zzz.c
```
#include <stdint.h>

typedef enum eNoteEdit
{
   EDIT_NONE,
   EDIT_START_TIME,
   EDIT_POS,
   EDIT_DURATION,
} ENoteEdit;

ENoteEdit g_noteEditMode = EDIT_NONE;

typedef struct
{
   uint8_t     note;    // 0-127
   uint8_t     dur;     // 0-32
   uint8_t     time;    // 0-32
   uint8_t     vel;     // 0-15
} EditDataNoteStep;

typedef struct
{
   uint8_t        index;
   EditDataNoteStep  notestep;
} SelectedEditDataNoteStep;

extern SelectedEditDataNoteStep* editedNoteStep;

void main(void)
{
   EditDataNoteStep* editDataNoteStep;
   
   g_noteEditMode = EDIT_DURATION;
   editDataNoteStep = &editedNoteStep->notestep;
}
```

zcc +zx -vn -a -clib=new zzz.c

```
sccz80:"zzz.c" L:33 Error:Must be lvalue
sccz80:"zzz.c" L:33 Warning:Expected ';'
sccz80:"zzz.c" L:34 Warning:Assigning 'editDataNoteStep', type: struct 0__anonst
ruct_0 EditDataNoteStep*editDataNoteStep from struct 0__anonstruct_0 notestep
```

suborb:
Thanks again, I understand this one too, instantiated enums are marked as KIND_ENUM which is wrong.

Incidentally, I was trying to compile the newlib sms examples and appmake failed on all of them. (Getting to the point), CandyKidDemo redefines enum_direction_type, which results in a sccz80 and clang error, but sdcc accepts.

aralbrec:
Not all of them were compilable with sccz80 before.  I just tried the moggy one and it does compile but the game doesn't work correctly.  See what happens after the identified problems are fixed.

suborb:
I can't get any newlib SMS examples to create a .sms file - it looks like appmake just exits.

However, I've looked at the asm for moggy and it's now functionally the same as before I broke everything.

aralbrec:
>I can't get any newlib SMS examples to create a .sms file - it looks like appmake just exits.

Which ones?  They are all working for me.

[...]

Issue #490, opened 2017-11-03
sccz80: errors in some example programs


aralbrec:
The rewrite is not compiling some example programs correctly.  I'll list them in this thread as I find them.

[sudoku.c](https://github.com/z88dk/z88dk/blob/master/libsrc/_DEVELOPMENT/EXAMPLES/sudoku.c)
```
zcc +zx -vn -startup=1 -O3 -clib=new sudoku.c -o sudoku -create-app

sccz80:"sudoku.c" L:91 Warning:Pointer arithmetic with non-matching types: struc
t sudoku_cell *sc-struct sudoku_cell grid[81]
sccz80:"sudoku.c" L:92 Warning:Pointer arithmetic with non-matching types: struc
t sudoku_cell *sc-struct sudoku_cell grid[81]
sccz80:"sudoku.c" L:153 Warning:Pointer arithmetic with non-matching types: stru
ct sudoku_cell *sc1-struct sudoku_cell grid[81]
sccz80:"sudoku.c" L:154 Warning:Pointer arithmetic with non-matching types: stru
ct sudoku_cell *sc1-struct sudoku_cell grid[81]
sccz80:"sudoku.c" L:154 Warning:Pointer arithmetic with non-matching types: stru
ct sudoku_cell *sc2-struct sudoku_cell grid[81]
sccz80:"sudoku.c" L:155 Warning:Pointer arithmetic with non-matching types: stru
ct sudoku_cell *sc2-struct sudoku_cell grid[81]
sccz80:"sudoku.c" L:338 Warning:Control reaches end of non-void function
```

The warnings are incorrect - the array is same type for pointer arithmetic.  The program compiles but produces incorrect results.  You can try with this sudoku:

```
790000300
000006900
800030076
000005002
005418700
400700000
610090008
002300000
009000054
```

aralbrec:
[open_memstream.c](https://github.com/z88dk/z88dk/blob/master/libsrc/_DEVELOPMENT/EXAMPLES/open_memstream.c)

```
zcc +zx -vn -O3 -clib=new open_memstream.c -o openmem -create-app --list

sccz80:"open_memstream.c" L:30 Warning:Loss of precision, converting int  to lon
g offset
```

This isn't a loss of precision since you're going from 16-bit to 32-bit.  But anyway the code generated connected to this warning is not right:

```
71    005F  21 04 00    	ld	hl,4	;const
72    0062  CD 00 00    	call	l_gintspsp	;
[...]

PR #494, opened 2017-11-04
Issue #490, #491, #493

suborb:
Fix up various type problems.

Add support for __func__


suborb:
* Reverse warning for int -> long conversion. It needs to be the otherway round!
* Make the type an explicit part of addglb() this way we can check for repeated definitions
* Change enum to KIND_INT, the values are still KIND_ENUM. 
* Allow forward definitions of  enums
* With & operator, make a pointer of lvalue 
* For an array set lval->ptr_type correctly to ->ptr type

PR #500, opened 2017-11-07
yaz180 - yabios wip6 - initial functional state

feilipu:
There is so much to do, but these yabios drivers now in a functional working state, that can be used to test from.
- load and store from IDE hard drive working
- load Intel Hex from asci0 working
- bank initialisation, copy, remove working
- memory dump (based on left-right & right-left DMAC memcpy_far) working.
- and finally a `jp_far` works too, so programs can be started in any bank.

The next thing to do is to work out a simple method of compiling applications. The `system_call` can discern whether the called function is "above the line" in COMMON AREA 1, and make a normal jump, or "below the line" where a bank switch has to be done. In the short term, I think there's enough to do with just getting the normal short jump access working.

Currently, I'm cutting a `-gp` file down into the COMMON AREA 1 functions, and then building that as an asm file along with the test program. Not very clean, but at least it works. :sweat_smile: 

[YAZ180 yabios repo](https://github.com/feilipu/yaz180/tree/master/yabios)

screenshot from 2017-11-07 14-53-31
<br>
screenshot from 2017-11-07 15-33-47


aralbrec:
> Currently, I'm cutting a -gp file down into the COMMON AREA 1 functions, and then building that as an asm file along with the test program. Not very clean, but at least it works. 😅

Manually doing that is probably quite painful :D 

The option "-gpf regexfile" will do it automatically.  Here "regexfile" is a file containing regular expressions, one per line with a preceding "-" (reject) or "+" (accept).  It will compare the symbol name with this ordered sequence of regular expressions and when a match is found, it will either reject or accept the symbol in the output .def file.  This way you can automatically pull out only the symbols you're interested in.

PR #511, opened 2017-11-11
Fix #441: add command line option for specifying the output directory

### LINKED ISSUE #441: z80asm: add command line option for specifying the output directory
See #440

pauloscustodio:
New option -O=outdir creates the output files in the given outdir. If the source file has a path,
that path is created under outdir. e.g. z80asm -O=obj src/mod1/file1.asm creates obj/src/mod1/file1.o


suborb:
Nice, I've converted libsrc/string to use both -O and glob, if that works (on windows as well) then I'll convert over the others that use z88dk-lstmanip which should hopefully speed up the build a lot on windows and a little on *nix

PR #513, opened 2017-11-12
Fix #312: Update z80asm zxn opcodes

### LINKED ISSUE #312: z80asm: implement zx next opcodes (was: serving fpga z80 variants)

pauloscustodio:
Updated parser to the recent changes in the zxn-specific opcodes.
Added macros for the copper unit of the zxn.


suborb:
I'll need to update sccz80, ticks and the rules to reflect recent changes in the opcodes.

I'm minded to wait for it to be official before doing anything though.

aralbrec:
Maybe we can split out the copper portion so we can merge that part in?

Issue #564, opened 2018-01-01
Pointer to Struct conversions not working


Fabrizio-Caruso:
While trying to do a little OOP in ANSI C through pointer to struct does not work.

Something like:
struct ParentStruct
{
  ...
};
typedef struct ParentStruct Parent;

struct ChildStruct
{
   Parent _parent;
... 
};
typedef struct ChildStruct Child;

Convert a poinyer to Child to a pointer to Parent, won't work.
It does work on other ANSI compilers, ie., you get erros anso Unknown member when trying to access
Parent's members.

This feature is very important because it allows OOP in ANSI C.

aralbrec:
I think you're just missing `struct`s in your typedefs.  C doesn't allow structs to be referred to using their names only (that's a C++ thing).

```
struct ParentStruct
{
   int a;
};
typedef struct ParentStruct Parent;

struct ChildStruct
{
   Parent _parent;
	int b;
};
typedef struct ChildStruct Child;

Child c;

void main(void)
{
	Parent *p = (Parent *)(&c._parent);
	p->a = 6;
}
```

Fabrizio-Caruso:
@aralbrec 
The missing struct was a typo (now edited). Sorry.

The problem is that sub-type polymorphism should work in ANSI C:
   Parent *p = (Parent *) &c

No need to access the c._parent explicitly.

When I use this, it does not always work as expected.

suborb:
> The problem is that sub-type polymorphism should work in ANSI C:
> When I use this, it does not always work as expected.

Have you got an example?

Fabrizio-Caruso:
I am getting an error if I cast to the "parent" struct and pass the cast object as an argument.
The line of my code (line 274)
		if(areCharctersAtSamePosition(&player, (Character *) itemPtr))
produces the following errors 
"
sccz80:"src/main.c" L:274 Error:Unknown member: _x
sccz80:"src/main.c" L:274 Error:Missing token, expecting ) got C
sccz80:"src/main.c" L:274 Warning:Too few arguments to call to function 'isCharacterAtLocation'
sccz80:"src/main.c" L:275 Error:Unknown member: _x
sccz80:"src/main.c" L:276 Error:Unexpected end of file
sccz80:"src/main.c" L:276 Error:Unexpected end of file
sccz80:"src/main.c" L:276 Warning:Converting pointer to integer without cast
sccz80:"src/main.c" L:274 Error:Unexpected end of file
[...]

PR #565, opened 2018-01-03
Issue #564: Casts and structs

suborb:
Fix an issue around casts and structs - the cast wasn't being applied
before doing the lookup.

However, doing the cast early breaks the lookup, so use the cast
type when doing the member lookup.


suborb:
I want to do some more tests on this to make sure I've not broken anything.

PR #568, opened 2018-01-06
Allow zx target IM2 to return to BASIC

derekfountain:
First stab at changing the CRT code to allow the Spectrum to return to BASIC with a Z88DK ISR in place.

It turns out there's a lot I still don't know. :) I changed the CRTs which appear to return with IM1 set, following CRT31 as a template. I don't know about the others so I left them alone. I also have no clue about other targets, so I didn't (deliberately) change those either.

I used the pragma as suggested in the discussion forum. I tested with my coin spinning animation nonsense:

https://spectrumcomputing.co.uk/forums/viewtopic.php?f=6&t=289&start=10#p3366

I built it using each modified CRT, both with and without the new pragma, and ensured it only worked when expected to. I spot checked the resultant code with a disassembler in an emulator and it looked correct each time.

I found the prototype for the im2.h file. I not entirely sure how its build works, or how it relates to other targets. I made the obvious change and hoped for the best. :) I wasn't entirely happy with the name of the new ISR defining macro: _WITH_BASIC. But I couldn't think what else to call it. I won't be offended if you change it to something more suitable.

I haven't updated any documentation in this pull request. I can do that once the code updates are complete.


aralbrec:
It should probably be made general for all targets.  There is a [general pragma](https://www.z88dk.org/wiki/doku.php?id=libnew:target_embedded#crt_configuration) now `CRT_INTERRUPT_MODE` that causes the crt to set interrupt mode before main is called.  A similar one should probably be added to optionally set interrupt mode on exit.  That's similar to this patch but applied everywhere.

The special macro for zx basic maybe needs to be qualified for the zx and zxn targets only.

I will have a look later on, thanks!

derekfountain:
Yes, I was very aware I was making a generic change for a zx-specific reason. I don't know anything about the other targets, and don't know how to test on them. I'd like to learn, so let me know what you'd like me to do with the patch.

aralbrec:
To keep the pragma naming consistent, the pragma controlling the exit im mode should probably be called `CRT_INTERRUPT_MODE_EXIT` and it can work the same way as `CRT_INTERRUPT_MODE`.  Ie values of 0,1,2 set the im mode but otherwise nothing is done.

The crt handles program startup and exit so the changes to accommodate the pragma are in the crts.  Every target has its own set of crts in its startup directory.

The most basic one is the [embedded z80 target](https://github.com/z88dk/z88dk/tree/master/libsrc/_DEVELOPMENT/target/z80/startup) and everything else is a augmentation of that.  Anyway the standard exit code in the embedded z80's (one) crt is this:

derekfountain:
I'm working through this and need a clarification. Or perhaps just a different explanation for my overheating brain. :)

The current behaviour for the zx target with the crts I looked at is to force "IM 1" in the exit code. That behaviour clearly needs maintaining, so although you're saying the new int-mode-on-exit default value will be -1 in crt_defaults.inc (meaning don't change mode):

`defc DEF__crt_interrupt_mode_exit   = -1`

in the zx target the target default should override that to ensure IM1 is generated at exit, as before. So zx/crt_config.inc needs:

`defc TAR__crt_interrupt_mode_exit   = 1`

I then looked at the target config file z180/crt_config.inc. That one has a TAR__crt_interrupt_mode of 1, which I think means that for those targets the z88dk code needs to run with IM1. OK, but what mode should it default to as the program exits?

Looking at z180_crt_0.asm.m4, I don't see any code being generated to restore the interrupt mode. Which strikes me as odd. The target default for those machines is to force mode 1 on z88dk program startup, but then leave it in mode 1 when the program exits. So the machine continues running with the interrupt mode set to whatever the z88dk program wanted, despite that program having exited.

Do I read that correctly? If so I can just leave the target default for that target at -1 in order to maintain current behaviour. But... am I reading it correctly?

aralbrec:
> The current behaviour for the zx target with the crts I looked at is to force "IM 1" in the exit code. That behaviour clearly needs maintaining, so although you're saying the new int-mode-on-exit default value will be -1 in crt_defaults.inc (meaning don't change mode):
> 
> `defc DEF__crt_interrupt_mode_exit = -1`
> 
> in the zx target the target default should override that to ensure IM1 is generated at exit, as before. So zx/crt_config.inc needs:
> 
> `defc TAR__crt_interrupt_mode_exit = 1`

Yes that's right.  There are three different configs in zx/config.inc.  As long as the default is to exit to basic (see crt_on_exit), then the interrupt mode should also set im1 mode on exit.

> I then looked at the target config file z180/crt_config.inc. That one has a TAR__crt_interrupt_mode of 1, which I think means that for those targets the z88dk code needs to run with IM1. OK, but what mode should it default to as the program exits?

Still -1 (ie do nothing).  The z180 powers up in im1 mode so if the user does nothing about that then the crt will also set im1 mode.  This is because the z180 probably uses a "restart" scenario when exiting.  When the program exits, there's nowhere to go so the crt just restarts and that restart will ensure the interrupt mode is set back to 1 in case the program itself had changed it.

derekfountain:
I've been working through all the targets and CRTs and I think I understand them all and have modified them accordingly. With one exception...

In the ZX Next code there are 3 cases where __CRTCFG=3: CRTs 776, 798 and 799. Their return to BASIC exit code is the same as the others for zxn, except they've had this added for interrupt mode control on exit:

```
   IF (__crt_interrupt_mode = 0) || (__crt_interrupt_mode = 2)
         im 1
   ENDIF
```

This seems to be saying that if the interrupt mode at the start of the program was set to 0 or 2, set it back to 1, otherwise leave it alone. I think it's probably a bit of a shortcut which is assuming that if the z88dk program changed mode at the start it should change it back at the end. It doesn't take into account what the program itself might have done. I'm guessing there's something special about these 3 CRTs which has necessitated this conditional step.

I'm not quite sure how to maintain this behaviour while introducing the __crt_interrupt_mode_exit constant. Can I just put:

```
      IF (__crt_interrupt_mode_exit >= 0) && (__crt_interrupt_mode_exit <= 2)
         im __crt_interrupt_mode_exit
      ENDIF
```

after the condition which is already there? That might result in an "im 1" instruction immediately followed by, say, an "im 2" instruction if __crt_interrupt_mode_exit is set accordingly. Not ideal.

aralbrec:
You can just replace that code with the new exit thing and set the default to do im1.

dot commands are commands run from basic by typing ".command_name".  These are loaded from disk, run and always return to basic.  You're in the same situation as the one you had - you have to go back to basic and you probably want im1 mode but on that rare occasion you may want to do something different.  So the new exit option will serve for that.

derekfountain:
OK, I've just pushed my second attempt. It turned out to be quite hard work, but I learnt an awful lot on the way. :)

I had a bit of a mix up with git merges; hopefully I got them all straightened out OK.

I wasn't sure about how to test anything. Is there a regression suite? The only thing I could actually test was the Spectrum changes using my little animation demo. At least that appears to work.

What follows is the contents of my notebook. It's a bit messy, but will give you a commentary on what I changed and why.

======
I've implemented this as you described in the previous post on this pull
request. There's now a __crt_interrupt_mode_exit constant which is populated by
a new block in crt_rules.inc from the default and target specific values in
DEF__crt_interrupt_mode_exit and TAR__crt_interrupt_mode_exit respectively.

I modified the crt_exit_eidi.inc file as you suggested, to carry the code to use
the __crt_interrupt_mode_exit constant to decide whether to add an "IM x"
instruction to the program exit point. The default value in
DEF__crt_interrupt_mode_exit is -1 which causes the eidi code to add nothing, so
by default there are no behavour changes.

I updated the zx targets first, since those are the ones I understand best. The
48K zx target used to add "IM 1" in the exit code, so I defaulted the
TAR__crt_interrupt_mode_exit value for that target to 1, hence no change of
behaviour. The other 2 targets (IF2 cartridge and esxdos dot command) did not
change the interrupt mode in their exit code, so I defaulted the
TAR__crt_interrupt_mode_exit value for those targets to -1.

I created the new include file you suggested, crt_exit_basic.inc, which
contains:

```
      ; returning to basic
      
      pop bc
      
      ld sp,(__sp_or_ret)
      
      exx
      pop hl
      exx
      pop iy
      
      IF (__crt_interrupt_mode_exit >= 0) && (__crt_interrupt_mode_exit <= 2)

         im __crt_interrupt_mode_exit

      ENDIF

      ei
      ret

```
That now includes into the zx 0, 1, 4, 5, 8, 9, 30 and 31 CRTs.

That code obviously doesn't match the exit points for the IF2 and esxdos dot
command targets, all of which use crt_exit_eidi.inc, and which defaults to no
change of behaviour for those targets. I checked all the CRTs for IF2 and they
all do exactly the same thing: include crt_exit_eidi.inc without touching
anything else interrupt related. I then checked all the esxdos targets and they
all do the same thing too. It's closer to the 48K model behaviour, but not the
same so the new include file can't be used. They all exit via the
crt_exit_eidi.inc file contents, and none of them change the interrupt mode, so
the target default of -1 is correct for all of them.

There's scope to use a common file for all IF2 cartridge exit points, and
another for the esxdos target exit points, but I'm probably not the person to do
that. :)

I then looked at the other targets. I'm not at all familiar with these, and
actually can't even tell you what most of them are, but looking at all their
startup CRTs and looking for interrupt related things:

z180: only 1 CRT for the 2 targets, and it exits via crt_exit_eidi.inc without
touching interrupts, so TAR__crt_interrupt_mode_exit set to -1.

rc2014: CRTs 0 and 256 exit via crt_exit_eidi.inc without touching
interrupts. I left CRTs 0 and 256 alone. CRTs 16 and 17 both force IM 1. In CRTs
16 and 17 I replaced the "im 1" with a check of the __crt_interrupt_mode_exit
value and a conditional "im" instruction. So to make this work the
TAR__crt_interrupt_mode_exit value needs to default to -1 for CRTs 0 and 256
and to 1 in CRTs 16 and 17. That will produce the same behaviour.

Looking at rc2014_crt.asm.m4 I see that __CRTCFG is set to 0 in the section which includes
crt0 (refered to in the comments as "acia"). Therefore in the
rc2014/crt_config.inc file, in the __CRTCFG=0 section, I set
TAR__crt_interrupt_mode_exit to -1, ensuring that for CRT0 the IM mode doesn't
get changed. Ditto for the __CRTCFG=2 section which includes the CRT256.

Back in rc2014_crt.asm I see that CRTs 16 and 17 are both included when
__CRTCFG=1, and those are the ones with require "IM 1" to retain current
behaviour. So for CRTCFG=1 I set TAR__crt_interrupt_mode_exit to 1.

zxn: 3 targets, lots of CRTs. For __CRTCFG=0 it needs to work the same as the zx
target. So I created a zxn/crt_exit_basic.inc file and put the standard exit
code in it. It's identical to the zx/ version and uses the same switch on
__crt_interrupt_mode_exit to reset the interrupt mode. That gets used for CRTs
0, 1, 4, 5, 8, 9, 16, 20, 24, 30 and 31. They all return with "IM 1" at the
moment so that's the default for TAR__crt_interrupt_mode_exit for __CRTCFG=0.

For __CRTCFG=2 the CRTs exit via crt_exit_eidi.inc, which means they have to
have their TAR__crt_interrupt_mode_exit set to -1. Hence no behaviour
change. This applies to CRTs 256, 257, 260, 261, 264, 265, 286, 287, 512, 513,
516, 517, 520, 521, 542 and 543. All those CRT files are unchanged.

For __CRTCFG=3 the CRTs are 776, 798 and 799. These had code in them which
inserted an "im 1" instruction if the __crt_interrupt_mode was 0 or 2, which I
took to mean "if the z88dk program changed interrupt mode on the way it, put it
back on the way out". As advised, I changed this to use the new
__crt_interrupt_mode_exit constant instead and defaulted
TAR__crt_interrupt_mode_exit to 1 for CRTCFG=3.

vgl: 2 targets with 4 CRTs. CRTs 0, 2001 and 4001 all exit via crt_exit_eidi.inc
without touching the interrupt. CRT 1 exits without doing very much at all. All
the interesting stuff is commented out. TAR__crt_interrupt_mode_exit set
to -1 for both targets so no change of behaviour for anything.

yaz180: 3 targets with 4 CRTs. This is the same as for rc2014: CRTs 0 and 256
are used when __CRTCFG is 0 and 2, respectively. Those shouldn't be touched
since they exit via crt_exit_eidi.inc without touching the interrupt, so
TAR__crt_interrupt_mode_exit is set to -1 for both. __CRTCFG=1 will pull in
either crt16 or crt17, and both force IM 1. So I replaced the "im
1" with a check of the __crt_interrupt_mode_exit value, and made it default to 1
for those CRTs.

z80: 3 targets with one CRT which exits via crt_exit_eidi.inc without touching
the interrupt. Exits via crt_exit_eidi.inc, so TAR__crt_interrupt_mode_exit set
to -1.

sms: One target with 6 CRT files, none of which touch the interrupt
mode. They all exit via crt_exit_eidi.inc, so TAR__crt_interrupt_mode_exit set
to -1. There's no TAR__crt_interrupt_mode in this one.

cpm: One target with two CRT files, neither of which touch the interrupt
mode. They both exit via crt_exit_eidi.inc, so TAR__crt_interrupt_mode_exit set
to -1. There's no TAR__crt_interrupt_mode in this one.

aralbrec:
Thanks Derek.  Yes, it a is a royal pain to make a global change like this.  There is no regression testing for this so care must be taken.  I'll doublecheck in the next day or so.

aralbrec:
I believe it's ok but I will wait to merge it tomorrow when I have a couple hours to do some testing and make changes where necessary.

aralbrec:
nightly build is made, let's merge and give it a go.

PR #597, opened 2018-02-06
RC2014 & Z80 - fix config

feilipu:
* Fix the RC2014 default config, which had omitted the ACIA interrupt by default.

* Adjust the `SECTION code_crt_start` to always come before `SECTION code_crt_init` for Z80. This matches the Z180 example. Without these changes the `code_crt_start` code, which contains the page 0 code and needs to begin at `0x0000`, could be supplanted (bumped downwards) by the inclusion of `code_crt_init` code, potentially contained in a `crt_preamble.asm`. Since the `code_crt_start` section contains the `RST` jump vector information, this should never be desirable.


aralbrec:
Section code_crt_start does not appear first in the output binary.  Section CODE is first and that's where the page_zero code is included to in the crts.  The optional preamble code deliberately supplants page zero as it's intended to allow the user to replace the page zero code with his own.  It's also used for specific targets to insert signatures at the front of binaries (org not zero) that are sometimes required by system software.

Section code_crt_start is used when the interrupt vector table is inserted into the page zero code so that code can be sequenced before the vector table (section CODE) and code can be sequenced after the vector table (code_crt_start).

This will probably be a bug in the z180 side.

feilipu:
> This will probably be a bug in the z180 side.

It is a [bug introduced ](https://github.com/z88dk/z88dk/commit/90b35ad216f447324141aead0708226c2c2466e7#diff-0f6228a6d7e9d76402670d598c8851e7)to resolve a bug that I was experiencing.

My issue was that when supplying [`crt_preamble.asm`](https://github.com/z88dk/z88dk/blob/master/libsrc/_DEVELOPMENT/target/yaz180/crt_preamble.asm) code in `SECTION code_crt_init`, that did not replace the standard interrupt vector code in [`crt_page_zero_z180.inc`](https://github.com/z88dk/z88dk/blob/master/libsrc/_DEVELOPMENT/target/crt_page_zero_z180.inc), but IS required to configure the Z180 processor BEFORE any other code, the preamble code was incorrectly appearing BEFORE the interrupt vector code.

Adding the `SECTION crt_page_zero` at the start of [`crt_page_zero_z180.inc`](https://github.com/z88dk/z88dk/blob/master/libsrc/_DEVELOPMENT/target/crt_page_zero_z180.inc#L2) is necessary otherwise this code (down to [here](https://github.com/z88dk/z88dk/blob/master/libsrc/_DEVELOPMENT/target/crt_page_zero_z180.inc#L432)) just ended up in the general code location (not right at the start) where it needed to be. It wasn't automatically ending up in `SECTION CODE`. Or maybe it was, but it was placed after all the other sections were positioned.

And, as I had [`__crt_org_vector_table`](https://github.com/z88dk/z88dk/blob/master/libsrc/_DEVELOPMENT/target/crt_memory_model_z180.inc#L12) defined as [neither of the two options](https://github.com/z88dk/z88dk/blob/master/libsrc/_DEVELOPMENT/target/yaz180/crt_config.inc#L35) noted, then the standard vectors in `SECTION code_crt_start` were never included in the right location, as they should have been as the first item. Moving `SECTION code_crt_start` out of the IF statement resolved that issue.

There will be another way to solve for what I needed, but the fix noted above (and this pull) only involved moving one line and adding one other. I'm not sure how to do it to support the other issues you mention above.

aralbrec:
ok will work on the terminals this weekend and then have another look at this.

feilipu:
@aralbrec you commit back in [July](https://github.com/z88dk/z88dk/commit/8aeaa582f5c32c538f58c1652b272dabb3b2adee#diff-78e49f687a6bb297ac3e943ff4ea4c05 ) should have reversed the previous custom Page 0 code, and added back the default configuration of `RST 38` for the ACIA. But, when looking at the `crt_config.inc` file it isn't there. Very odd.

aralbrec:
I'll have a pass at all the terminals but I'd like to get the basic_dcio on the rc2014 working as the prototype first.

feilipu:
I'll close this and resubmit two separate pulls. It was a mistake to conflate these two things.

The RC2014 default config is an easy repair, which somehow didn't work on [July 7th.](https://github.com/z88dk/z88dk/commit/8aeaa582f5c32c538f58c1652b272dabb3b2adee#diff-78e49f687a6bb297ac3e943ff4ea4c05)

The Z80 initialisation and section definitions is a more comparative thing. Ideally, an outcome that enables placing a `crt_preamble.asm` without disturbing the default `page_zero_z80.inc`, or the vectors associated with the default page zero, is desired.

A simple method of doing this is in z180 currently, but this solution has probably broken other things.

PR #612, opened 2018-02-20
Makefile: make git hashes/version overridable

Mic92:
This allows to set them without building z80dk from source without git
repository and therefore helps with to make builds more reproducible (on
NixOS).


suborb:
Thanks - have you got any other makefile changes that you're patching for?

Mic92:
No. Now the build works as it is: https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/compilers/z88dk/default.nix

bignaux:
it looks like my patch :D

PR #622, opened 2018-03-01
Crap graphics for Kaypro '83 models

zx70:


suborb:
With a title like that I want to see a screenshot!

zx70:
The '84 models (most of the existing Kaypro computers) have a better resolution (160x100) but the z88dk support level is lower because I used the terminal ESCape codes, I strongly suspect that the compatibility is guaranteed at BIOS level, I'll probably try to get something similar for the Otrona Attachè.


Here I'm attaching only the output for the earlier, less compatible, model.


Please ask as much screenshots you like  ;)


Stefano


________________________________
Da: suborb <notifications@github.com>
Inviato: giovedì 1 marzo 2018 10:06
A: z88dk/z88dk
Cc: Stefano; State change
Oggetto: Re: [z88dk/z88dk] Crap graphics for Kaypro '83 models (#622)


With a title like that I want to see a screenshot!

zx70:
The pictures currently in the Wiki show a bug which now is solved, I'll replace them sooner or later.

PR #626, opened 2018-03-03
Intdoc

derekfountain:
This is the next installment in the getting started guide. It looks at interrupt mode 2 on the Spectrum.

I've been sitting on this for weeks and always felt I should delve deeper, but I've rather moved on so it's probably better to submit what I've done rather than wait indefinitely for more. I've tidied and proof read it.

Over the weeks I've done git-ish things on the branch I'm not hugely confident about. Rebase and stuff. I think it should be OK to merge but if there are any problems let me know. It's only 3 files so I'm sure we can work something out if I've messed up the branch. :)


aralbrec:
Welcome back derek.  Looks good - there may be a few things to correct but I will merge now.  Eg, if the heap is used, you have to tell it what the last byte in memory is or it will continue to assume it can use up to the bottom of the stack.

Issue #596, opened 2018-02-06
YAZ180 - support yabios inline RST system calls


feilipu:
Related to #595 the current mechanism for in-line assembly RST instructions as[ macros in a header](https://github.com/z88dk/z88dk/blob/master/include/_DEVELOPMENT/proto/arch/yaz180.h#L69) file doesn't work. It won't compile.

```c
#define call_error(code)        \
    do{                         \
        __asm                   \
        rst 8H                  \
        defb code               \
        __endasm;               \
    }while(0)

#define call_far(addr, bank)    \
    do{                         \
        __asm                   \
        rst 10H                 \
        defw addr               \
        defb bank               \
        __endasm;               \
    }while(0)

#define call_sys(addr)          \
    do{                         \
        __asm                   \
        rst 20H                 \
        defw addr               \
        __endasm;               \
    }while(0)

#define call_apu(cmd)           \
    do{                         \
        __asm                   \
        rst 28H                 \
        defb cmd                \
        __endasm;               \
    }while(0)
```

A complete solution was developed for [`break()`](https://github.com/z88dk/z88dk/blob/master/include/_DEVELOPMENT/proto/arch/yaz180.h#L104) in #374 but that was [quite complicated](https://github.com/z88dk/z88dk/blob/master/lib/yaz180_rules.1#L2) and fragile because of the various SDCC optimisation techniques.

Ideally, a mechanism of inserting a non-optimising but byte counting mechanism for in-line assembly is needed. Building this the same way that `break()` was built is probably not the answer. Probably the solution to this and #595 would be similar.

suborb:
Just picking this up (a bit late).

This was something that I was experimenting with a long time with sccz80 - I was thinking about implementing shared libraries on the z88 and could hijack a rst to call via.

As a result,  `__SHARED__` and `__SHAREDC__` function decorators were introduced.`__SHARED__` just did a call [function]_sl (which would be a stub to a rst), and `__SHAREDC__` did a rst 8/defw [funcname]

Of course, calling via a rst introduced a stack offset, such that arguments weren't at sp+2 on entry to the function which is where the -shareoffset came in which added a displacement when compiling them.

Of course this is all a bit dusty and irrelevant since I don't think sdcc has similar concepts and keeping source compatibility is important these days.

suborb:
The attached diff simplement __z88dk_shortcall(rst_number, value) for both compilers. It's a rough cut and doesn't consider a call through a function pointer.

[...]

Issue #595, opened 2018-02-06
zsdcc: move code inlining for +zxn to peepholer and make it aware of z80-zxn instructions


aralbrec:
From the bug reported in the [forum](https://www.z88dk.org/forum/viewtopic.php?pid=15818#p15818).

Some code inlining is being done in post-processing which increases code size.  In the case reported in the forum, the code size is increased enough to put some JR out of range.   The problem happens because zsdcc is doing JP->JR substitution in its peephole phase before the code is inlined and where it exists as a 3-byte CALL.  A following copt phase replaces the CALL with a longer sequence of code.  This inlining must be moved into zsdcc's peepholer instead.

Related to this is #55 where it's proposed that zsdcc's active peephole set is customized depending on architecture and user-supplied rule sets.

zsdcc also has to be made aware of the new z80-zxn instructions so that it knows both size and cycle count.

PR #637, opened 2018-03-10
yaz180 - revise RST functions and __sfr

feilipu:
- move `__IO_BREAK` and other IO ports to be `__sfr`
- rework the `RST` functions as macros

The `RST` macros seem to work as expected.
There is probably a better way to do this, but I'm afraid I don't it.


aralbrec:
There is a way to do the rst if the params are passed using the usual linkages on stack.  So a single param rst could be fastcall and have the value passed in DEHL and multiple param rst could use callee linkage.  But there is no way to do the common `rst n` with bytes defined following yet.  I think that will come with #595 as noted in #596 .

PR #638, opened 2018-03-13
Fix message when undetected Currah uSpeech

garrafonsoft:


suborb:
Makes sense :) Thanks!

PR #641, opened 2018-03-13
__far and other calling convention odds, and misc other

suborb:
Bit of an odd sorts PR, opened for comments and thoughts.

* Add the proposed __z88dk_shortcall() for calling via a rst, probably needs a better name?
* Provide a _Static_assert()
* Support (and extend) the sdcc #pragmas for setting sections
* Tag BSS variables with the current BSS section and dump accordingly
* Introduce __far (left associative) to indicate that a pointer is far.
* Allow calling __far function pointers - through an (unimplemented) l_farcall library routine

Will need to add a few code generation tests before merge


suborb:
There's an updated sdcc patch in #596 to support __z88dk_shortcall().

I'm additionally going to add a `__z88dk_arguments_offset([bytes])` to both compilers to define the offset to the start of the parameters.

In sccz80 this will replace the -shareoffset=,-shared-file and -make-shared options, I'll also remove `__SHARED__` (and related -use-shared option) and `__SHAREDC__` in favour of the new mechanisms in this PR. The  `__SHARED__` functionality can be achieved by linking another library providing those stubs without the `_sl` suffix, so doesn't really provide anything.

suborb:
sdcc patch that implements __z88dk_shortcall() and __z88dk_params_offset()

[sdcc_far_decorators.diff.txt](https://github.com/z88dk/z88dk/files/1813409/sdcc_far_decorators.diff.txt)

suborb:
@aralbrec can you do a windows build of the zsdcc with this patch?

aralbrec:
I will give it a try tonight.

aralbrec:
An updated windows zsdcc binary can be found in:
https://github.com/z88dk/z88dk/blob/master/libsrc/_DEVELOPMENT/sdcc_z88dk_patch.zip

suborb:
Thanks for that, the nightly build is now updated so will pick them up.

I'll merge this, but I do want to do an example l_farcall example.

PR #648, opened 2018-03-17
Sp1 one

derekfountain:
My first attempt at article 01 of an SP1 getting started guide. This one will need a careful proofread, if you'd be so kind. :)

I've also added an index to the getting started series. There's enough here now to formalise a bit. I added my name and link to it, hope that's acceptable.


aralbrec:
I don't want this to languish hear until I get around to proof-reading but will do so soon.

PR #651, opened 2018-03-18
rc2014 - acia polishing

feilipu:
Fixing the rc2014 acia to match the version I'm working on currently for CP/M off IDE FATFS.
Saves a few cycles.


feilipu:
Junk - a moment's reflection tells me so.

PR #663, opened 2018-03-24
yaz180  rc2014 - set PUBLIC __Restart for ROM

feilipu:
Make the `__Restart` label `PUBLIC` to provide a specific location for restarting applications to call into where the subtype is ROM.

Also some assorted fixes that don't warrant a separate PR.


aralbrec:
The __Restart label is actually not a place you can jump to directly to restart the code.  It works in combination with the exit code where the command line might be zeroed and the interrupt mode may be set to the start state.  You also have to use a rom model compile to ensure the variables are properly initialized.

feilipu:
I looked at the options to do a "clean" z88dk style exit from where I was, and it wasn't pretty.

The scenario is that with the rc2014 CP/M-IDE CCP running it is possible (probable) that a CP/M application has trashed the heap and stack created by the shell application, since they exist in the same RAM address space, so unwinding the call sequence to return back to the  shell application won't work.

I was originally going to use the `__Start` label as the destination, but when looking through the crt0 listing it seemed `__Restart` was a more logical name and place to jump to restart without a command line.

Happy to hear an alternative option.

aralbrec:
Looking at the crt, on exit from main() via ret, the exit stack, if present, will be run:

```
   ; run exit stack

   IF __clib_exit_stack_size > 0
   
      EXTERN asm_exit
      jp asm_exit              ; exit function jumps to __Exit
   
   ENDIF
```

If you call the stdlib function exit() instead, the exit stack will be run and then it jumps to the `__Exit` label in the crt.  (There are similar rules for quickexit() which is a new-ish add-on to the c standard).

If you bypass that, you miss out on the exit stack (if needed).

The pragma options do one of a few [return scenarios](https://www.z88dk.org/wiki/doku.php?id=libnew:target_embedded#what_happens_at_program_exit).  If everything is trashed you can jump to a fixed exit point in the monitor rom to return (3. JUMP TO ABSOLUTE ADDRESS OR LABEL ) or you can jump to the caller instead of return there (5. JUMP TO CALLER).  With this latter one the crt pops the return address at the start and saves it to jump to on exit.  The restart option is also there but your ram variables will not be initialized unless the compile was using a rom model as that's the only compile model that stores a copy of the data section for initialization purposes.  ram model compiles can be told to initialized the bss section to zero.  In all cases, HL holds the return status on exit.

There's also a section "code_crt_exit" that is the counterpart to section "code_crt_init" that allows you to place code that runs on exit.  The [exit code](https://github.com/RC2014Z80/RC2014/blob/master/ROMs/CPM-IDE/cpm22.asm#L1232) you have maybe could be placed there.

aralbrec:
BTW I think you missed committing `cpm22.lst` to the repo :)

In the preamble code there is a `call _acia0_init` but the library should be placing that call itself in section `code_crt_init` since it's set up to do that if any of the acia functions are used.  There may be two of those in the output binary but it probably won't hurt anything.

feilipu:
Noted on both points. You are correct. Thanks. :-)

PR #665, opened 2018-03-25
z80asm: symbol declared public and not defined should raise an error

pauloscustodio:
@aralbrec please check the changes I had to do in the library to get a successful build after including the new error message, before the merge


aralbrec:
Amazing how many copy and paste errors were caught.  Should be good now if it passes.

pauloscustodio:
Thanks for the fixes in the library. I'm running a final check and will merge when it passes.

PR #671, opened 2018-03-29
Feature/issue 658 comparisons

suborb:
Redo comparisons to constant within sccz80.

Allow chars to pass down to the codegen so we can use short routines for char comparisons. 

Remove the old comparison to 0 handling - do it via xxx_const() now which means we can catch all clauses of an if statement.

Add range checking which will evaluate to true false and eliminate the dead code that will result (dead code elimination previously done)

This PR will sit here for a few days whilst I think about it.

Note to self: md5, -Cc--opt-code-speed -sccz80-inline-ints = 44672387 vs 45585709 in master


suborb:
Need to finish of constant handling in ? operator. At present in this branch the constant isn't output/we don't skip one of the branches.

PR #675, opened 2018-04-01
Create a Dockerfile for z88dk

garrafonsoft:
This file creates a container that includes a compiled z88dk git snapshot including a patched SDCC version to be used with Docker.
Note: in SDCC compilation the ds390 and ds400 ports had to be disabled to avoid compile errors. On the other hand I am not sure which other ports can be disabled to have a minimal SDCC compilation to be used in z88dk.

### DIFF
--- a/z88dk.Dockerfile
+++ b/z88dk.Dockerfile
@@ -1,0 +1,19 @@
+FROM ubuntu:latest
+
+LABEL Maintainer="Garrafon Software (@garrafonsoft)"
+LABEL Version="0.1"
+LABEL Date="2018-Apr-01"
+LABEL Description="A basic Docker container to compile and use z88dk from GIT"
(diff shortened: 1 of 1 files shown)


feilipu:
> I am not sure which other ports can be disabled to have a minimal SDCC compilation to be used in z88dk.

This works for me. It may not be "minimal", but it removes everything unnecessary.
The `gbz80-port` I wouldn't use. If you want it then you know what to do.

```bash
cd sdcc
./configure --disable-mcs51-port --disable-gbz80-port --disable-avr-port --disable-ds390-port --disable-ds400-port --disable-hc08-port --disable-pic-port --disable-pic14-port --disable-pic16-port --disable-stm8-port --disable-tlcs90-port --disable-s08-port --disable-ucsim --disable-device-lib --disable-packihx
make
```

PR #676, opened 2018-04-01
 Create a Dockerfile for z88dk

garrafonsoft:
This file creates a container that includes a compiled z88dk git snapshot including a patched SDCC version to be used with Docker.
Note: in SDCC compilation the ds390 and ds400 ports had to be disabled to avoid compile errors. On the other hand I am not sure which other ports can be disabled to have a minimal SDCC compilation to be used in z88dk.

### DIFF
--- a/z88dk.Dockerfile
+++ b/z88dk.Dockerfile
@@ -1,20 +1,20 @@
-FROM ubuntu:latest
+# To create the image:
+#  macOS / GNU/Linux:
+#   $ docker build --squash -t z88dk - < z88dk.Dockerfile
+#  Windows 10 PowerShell:
+#   PS> cmd /c 'docker build --squash -t z88dk - < z88dk.Dockerfile'
+# To run the container (works for macOS, GNU/Linux & Windows PowerShell):
+#  docker run -v ${PWD}:/src/ -it z88dk <command>
 
-LABEL Maintainer="Garrafon Software (@garrafonsoft)"
-LABEL Version="0.4"
-LABEL Date="2018-Apr-03"
-LABEL Description="A basic Docker container to compile and use z88dk from GIT"
-LABEL Creation_Command="docker build -t z88dk - < z88dk.Dockerfile"
-LABEL Execution_Command="docker run -v ${PWD}:/src/ -it z88dk <command>"
+FROM alpine:latest
 
-RUN apt-get update && apt-get install -y build-essential git m4 libxml2-dev subversion bison flex libboost-dev texinfo && apt-get clean && rm -rf /var/lib/apt/lists/*
+LABEL Version="0.5" Date="2018-Apr-05" Maintainer="Garrafon Software (@garrafonsoft)" Description="A basic Docker container to compile and use z88dk from GIT"
 
-RUN cd /opt && git clone --recursive https://github.com/z88dk/z88dk.git && cd z88dk && chmod 777 build.sh && ./build.sh
+RUN apk add --no-cache build-base m4 bison flex libxml2 boost && apk add --no-cache -t .build_deps libxml2-dev git subversion boost-dev texinfo && mkdir /opt
 
-RUN cd /opt && svn checkout -r 9958 svn://svn.code.sf.net/p/sdcc/code/trunk/sdcc && cd sdcc && patch -p0 < /opt/z88dk/src/zsdcc/sdcc-z88dk.patch && ./configure --disable-mcs51-port --disable-gbz80-port --disable-avr-port --disable-ds390-port --disable-ds400-port --disable-hc08-port --disable-pic-port --disable-pic14-port --disable-pic16-port --disable-stm8-port --disable-tlcs90-port --disable-s08-port --disable-ucsim --disable-device-lib --disable-packihx && make && cd bin && mv sdcc /opt/z88dk/bin/zsdcc && cp sdcpp /opt/z88dk/bin/zsdcpp
+RUN cd /opt && git clone --depth 1 --recursive https://github.com/z88dk/z88dk.git && cd z88dk && chmod 777 build.sh && ./build.sh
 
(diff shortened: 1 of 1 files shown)


garrafonsoft:
Needed to update the previous PR :-( Seemed that single quotes are not needed for Execution_Commmand.

aralbrec:
From philip's last input:

```
./configure --disable-mcs51-port --disable-gbz80-port --disable-avr-port --disable-ds390-port --disable-ds400-port --disable-hc08-port --disable-pic-port --disable-pic14-port --disable-pic16-port --disable-stm8-port --disable-tlcs90-port --disable-s08-port --disable-ucsim --disable-device-lib --disable-packihx
```

You can disable all ports except the z80-related ones (z80, z180, rabbit).  The libraries do not support gbz80.

aralbrec:
Is it possible to supply an install directory to override the default /bin/z88dk ?

garrafonsoft:
Regarding the z88dk install path being hardcoded, I think it doesn't matter where it installed because when the Docker container is use is transparent for the user. But maybe is better to locate both source code trees under /opt rather than in the root dir.

suborb:
Could you provide some (markdown) instructions on usage please?

- How to build
- How to launch
- How to run commands

Then it can go on the wiki as a canonical source of information

suborb:
This is working nicely on a MacOS, rather scarily, compiling the examples is quicker using docker than natively!

Instructions are fantastic too, thanks a lot @garrafonsoft 

If someone can confirm that this works on windows, this can be merged.

aralbrec:
I had a quick go last night but docker was unable to run (no fault of this dockerfile).  Unfortunately I only have a win 8.1 machine at home and docker has moved on to win 10.  I used their previous offering docker tools and that fails to run because bash is missing.  I assumed docker provided the environment but maybe I am wrong about that.  I will see if I can get bash on my msys install and try it that way.

garrafonsoft:
Please don't merge for now.
I have recently get access to a Windows 10 installation and I am starting to test this Dockerfile. Until now I could only test on macOS and GNU/Linux.
As there are 3 possible shells (PowerShell, cmd.exe and WSL) in Windows 10, seems that Windows PowerShell is the most suitable for Docker.
I will update the instructions comment above and the Dockerfile with my findings. And tell you when Dockerfile is in a more "mature" state. :-)

garrafonsoft:
Right now I am quite happy with the current Dockerfile status, but the main issue is that it currently provides a vanilla compilation of the z88dk GIT repository.

I am aware that some files can be customized to get a more suitable z88dk compilation. I.e.:
```bash
https://github.com/z88dk/z88dk/blob/master/libsrc/_DEVELOPMENT/target/zx/config/config_sp1.m4
```
I am thinking of files that a videogame developer would need to customize for its needs. Any more examples of different customizable files in the z88dk source tree?

I want to try to provide a simple mechanism to create a Docker image with a customized z88dk compilation. Let's see where I accomplish! :-)

aralbrec:
Library customization at build time is mainly a newlib thing.  All newlib targets have a directory in [z88dk/libsrc/_DEVELOPMENT/target](https://github.com/z88dk/z88dk/tree/master/libsrc/_DEVELOPMENT/target) and they each hold their own config files in a config subdirectory, eg for zx [z88dk/libsrc/_DEVELOPMENT/target/zx/config](https://github.com/z88dk/z88dk/tree/master/libsrc/_DEVELOPMENT/target/zx/config).  It's these config files that can be customized.   Most of them are straight defines so that they wouldn't normally be changed but several are intended to be modifiable.  For the zx target, the bifrost / nirvana / sp1 / target / clib config files are intended to be modifiable.  All the newlib targets will have target / clib config files.  Target may control things like setting a specific model (on the zx setting pentagon means nirvana games can be compiled for the pentagon) and clib controls things like whether you want the fast integer library, faster versions of string functions, etc.

Once changes are made there, the user is expected to rebuild the library from [z88dk/libsrc/_DEVELOPMENT](https://github.com/z88dk/z88dk/tree/master/libsrc/_DEVELOPMENT) by either running `Winmake target` or `make TARGET=target` (or build them all with `Winmake all` or `make TARGET=all`.  This will place new library files in [z88dk/libsrc/_DEVELOPMENT/lib](https://github.com/z88dk/z88dk/tree/master/libsrc/_DEVELOPMENT/lib) where compiles will pick them up.

I'm not sure how this would fit in with docker or with a system install for linux (not docker).  For the latter, ideas for the future include having a user-defined portion of the install in the user's home directory that is searched for important files before the system portion is searched.

There is also a 3rd party library install mechanism.  Third parties can distribute headers and libraries as user files that can be made part of a compile easily but this other mechanism puts it inside the z88dk include/library locations so that it is always available for any compile without having to have the 3rd party files always present among the project's files.  There is also a bit of complication having to do with using the right library for each compiler and compile type which is what this library install mechanism solves.  The installer is z88dk-lib and it puts stuff into z88dk/lib, z88dk/include, z88dk/libsrc/_DEVELOPMENT/lib, z88dk/include/_DEVELOPMENT and may run z88dk/include/_DEVELOPMENT/Winmake or make to build newlib headers.  But this is a step further.

garrafonsoft:
Ok, I think I am done with this PR... the mechanism to provide a customized z88dk compilation is not very straightforward but I couldn't come up with a better solution without getting into complicated Docker pipelines or completely modifying the Dockerfile.

aralbrec:
Let's give it a try.  For lack of a better location, I've put the documentation above into z88dk/doc.

garrafonsoft:
It's Ok for me. :-) Thanks all for the support with this PR!

suborb:
I've copied the documentation onto the GitHub version of the wiki: https://github.com/z88dk/z88dk/wiki/Docker-Usage

PR #677, opened 2018-04-02
RC2014 - SIO driver

feilipu:
This PR is for the RC2014 SIO.
It can be used with any SIO device provided that the control and data registers are mapped appropriately.

There are two SIO boards available for the RC2014, that of Dr Baker, and the official one by Spencer.
Dr Baker followed the data sheet when defining channel / registers. Spencer did something original.
Hence there are two different channel / register locations.

The SIO would be an alternative for the ACIA, but isn't going to be the basic option.
The ACIA will remain the cheapest option for RC2014 serial interfacing.

Both SIO and ACIA are located at `0x80` by default.
The ACIA can be moved to `0x40`, but this is not the normal configuration.

All of the relevant defines are included in the PR currently, but...
- [x] A new startup option will need to be defined, to allow the SIO to replace the ACIA.
- [x] Interrupt code in `IM 2`, and `putc`.
- [x] Implement `RTS` flow control.
- [x] The drivers have been tested on a SIO board.

Key features.
 - Uses [vectored interrupts](https://github.com/z88dk/z88dk/pull/677/files#diff-496f07ba4e9da2fcd6c01a625d3c8f91), based on `IM2` and the SIO capability to provide a separate `IM2` address for each interrupt. Vector priority is internally daisy-chained within the SIO.
 - Transmit is buffered with cut-through if the buffer is empty.
 - Receive is buffered to 255 bytes, to prevent overflow.
 - `RTS` code is present, but is commented out based on most USB interfaces not supporting `RTS`.

 @aralbrec @suborb. This can be merged at any time. I'll continue to polish, and may have one or two updates over the next weeks (after integration into CP/M), but I think it is done, done.


feilipu:
I'm having an issue implementing the `IM 2` interrupts for this SIO driver. My issue is that the [`SECTION interrupt_vectors`](https://github.com/z88dk/z88dk/pull/677/files#diff-d8a4837052b63b279603ee6babeaf5b9R2) code is not being picked up, as I believe it should.

Also, the source labels in [`sio_interrupt.asm`](https://github.com/z88dk/z88dk/pull/677/files#diff-b976a9a7aa6b59670a183d6f3669fbfd) are not being generated.
They're just stubs now, but they should exist. Right?

I'm testing with a simple C program, to ensure all the things are being placed correctly.
```C
#include <stdio.h>
#include <string.h>
#include <arch.h>
#include <arch/rc2014.h>

// zcc +rc2014 -subtype=sio -SO3 -clib=sdcc_iy -v -m --list --c-code-in-asm --max-allocs-per-node20000 test.c -o test -create-app

void main(void)
{
    fprintf(stdout,"Hello World\r\n");
    fprintf(ttyout,"Hello World\r\n");
}
```
I note that the code in `SECTION interrupt_vectors` is not being collected because the resulting map file shows this.
```
__code_crt_return_head          = $00C7 ; const, public, def, , ,
__code_crt_return_tail          = $00CB ; const, public, def, , ,
__code_crt_return_size          = $0004 ; const, public, def, , ,
__code_crt_common_head          = $00CB ; const, public, def, , ,
__code_crt_common_tail          = $00CB ; const, public, def, , ,
__code_crt_common_size          = $0000 ; const, public, def, , ,
__interrupt_vectors_head        = $00CB ; const, public, def, , ,
__interrupt_vectors_tail        = $00CB ; const, public, def, , ,
__interrupt_vectors_size        = $0000 ; const, public, def, , ,
__code_driver_head              = $00CB ; const, public, def, , ,
__code_driver_tail              = $00E2 ; const, public, def, , ,
__code_driver_size              = $0017 ; const, public, def, , ,
```
Any thoughts as to where to look to fix this concern, please?

aralbrec:
> My issue is that the SECTION interrupt_vectors code is not being picked up, as I believe it should.

There has to be a reference in the compile to public symbol in [SECTION interrupt_vectors](https://github.com/z88dk/z88dk/pull/677/files#diff-d8a4837052b63b279603ee6babeaf5b9R2) otherwise the linker won't bring it into the binary.  There are no public symbols here so this won't make it into the binary.  At the bottom there is the NEED trick that makes reference to a file with the NEED symbol defined so that that file is brought into the compile.

Interrupt vectors are something that change according to the programmer and having their inclusion dependent on the linker means you can't control what order they will appear in the table (and on what vector they will lie on) when multiple im2 devices are present.

Instead of placing the vectors into the library or an object file, I think the best course is to make at an asm include file.

For the zxn target there is an [include directory](https://github.com/z88dk/z88dk/tree/master/libsrc/_DEVELOPMENT/target/zxn) and you can just add one for the rc2014 and put the vectors in there.  During compile, `target/rc2014` will be part of the asm search path so asm files can contain something like `INCLUDE "include/im2v_sio.asm"` (I'm not sure what a good naming scheme would be) to pull in vectors.  Then the idea would be the programmer has responsibility and control over building the im2 vector table in a project file like "interrupt.asm"

```
section interrupt_vectors

INCLUDE "include/im2v_sio.asm"
INCLUDE "include/im2v_pio.asm"
...
```

And `interrupt.asm` would be added to the compile line.

You would probably need to add a public label to the include file so that the library can know which vectors the sio is placed on.  You can just take the address of the label & 0xff as the vector id for any initialization code put in `code_crt_init`.

>Also, the source labels in sio_interrupt.asm are not being generated.
>They're just stubs now, but they should exist. Right?

Yes they should exist.  It may be another case of code not making reference to anything in that file so it's not in the output binary.  I will take a look later when I get a chance.

aralbrec:
There may be a way to weak-define the interrupt table in case the programmer doesn't supply one and maybe that's what you were attempting.  Will have to look later.

feilipu:
> Instead of placing the vectors into the library or an object file, I think the best course is to make at an asm include file.

Thanks. Yes. Sometimes a nudge in the right direction is all that is needed.

I moved the interrupt table to an [include file](https://github.com/z88dk/z88dk/pull/677/commits/0a1404393f2d0d0a88af70a36534ce7d599e69ba#diff-8fe5cd364b824bbf5d41069211d7f809R217), that is [only included when the SIO startup](https://github.com/z88dk/z88dk/pull/677/commits/0a1404393f2d0d0a88af70a36534ce7d599e69ba#diff-496f07ba4e9da2fcd6c01a625d3c8f91R6) is being used.
Works, I think (noted that I don't have any way of testing this yet).

aralbrec:
This is ready to go?

feilipu:
Yes. Fully cooked.

Thanks.

aralbrec:
Nice addition Philip.

Later, I'd like to revisit how the vector table is done so that the user has control over (optionally) constructing his own.

The vector base address for the sio is hard coded in `__sio_init_async_rodata` so that the sio vectors must appear first in the table (see `defb    __crt_org_vector_table&$F0`).  A label in the include file `crt_interrupt_vectors_sio.inc` can get around that.

I'd also like to make it easier for the user to create his own im2 table so that he just includes the sio into that and assembles his as part of the compile.  crt4 includes the sio vectors directly.  This will put the sio vectors into a user-defined table (without the user including them) but it will always appear first in the table because it is seen first (being in the crt).  Maybe that is the plan, in combination with the above.  This system should work with custom crts, crt256 and custom driver instantiations too.

rc2014 and yaz180 are also defining their io ports in lower case (`__sfr __banked __at __IO_PIO_PORT_A  io_pio_port_a;`) which is different from other ports where they are defined in upper case.  One or the other should be chosen.

```
io_pio_port_a = 0xff;
IO_PIO_PORT_A = 0xff;
```

feilipu:
I have a few other clean ups and back ports to do later on the ACIA and YAZ180, that didn’t fit in here. 

Perhaps we can put it all in that PR?

And of course there’s demand for simultaneous ACIA and SIO support. Although the default base Port is `0x80` for both. TBD.

aralbrec:
There looks to be a bug with creating the im2 table.

The crt supports four scenarios:

1. `__crt_org_vector_table = -0x80`  The im2 vector table is half size starting with vector 0x80 and is located at address 0x80 inside the library's page zero code.
2. `__crt_org_vector_table = -0x100`  The im2 vector table is full size and is located at address 0x100 inside the library's page zero code.
3. `__crt_org_vector_table = 0`  The im2 table is generated as a separate output binary that the user is responsible for placing in memory somehow.
4. `__crt_org_vector_table = anything else`.  The im2 table is generated and held in the output binary.  But it is not copied to its destination address in ram - that is the user's responsibility.

crt4 is set up for no.4 above by default `defc TAR__crt_org_vector_table      = 0xFFF0`.  But I don't see anyplace that the im2 table is copied into ram.  There should be some code inserted into `code_crt_init` that copies the table to the destination address, something like:

```
ld hl,__interrupt_vectors_head
ld de,__crt_org_vector_table
ld bc,__interrupt_vectors_size
ldir
```

Maybe this is something all crts in z88dk should be doing by default when case 4 is chosen.

feilipu:
I found the [code that did the copying for me](https://github.com/z88dk/z88dk/blob/667cc8b24dbf52f07563f940b04b3198e7c3b3c9/libsrc/_DEVELOPMENT/target/crt_set_interrupt_mode.inc#L1), and the code works.

BTW. I'm not 100% certain what is the problem with `defb __crt_org_vector_table&$F0`?
Is it that `__crt_org_vector_table` is fixed during the library compile, and not the end user compile?

For reference, the SIO demands a whole nibble for its vector table when it is set so that [interrupt affects the vector](https://github.com/z88dk/z88dk/blob/667cc8b24dbf52f07563f940b04b3198e7c3b3c9/libsrc/_DEVELOPMENT/target/rc2014/device/sio/__sio_init_async_rodata.asm#L25), so the low nibble of the low byte of the IM2 vector has to be zero. Hence the value that is written to Channel B Register 2 has to be `$F0`.

aralbrec:
> Nope. Creating the IM2 table is already there.

It's in `crt_set_interrupt_mode.inc`:

```
   IF (__crt_org_vector_table > 0)

      ; copy interrupt vector table to final destination

      EXTERN __interrupt_vectors_head
      EXTERN __interrupt_vectors_size

      ld hl,__interrupt_vectors_head
      ld de,__crt_org_vector_table
      ld bc,__interrupt_vectors_size

      EXTERN asm_memcpy
      call   asm_memcpy
   
   ENDIF
```

I must be going senile.

> I'm not 100% certain what is the problem with defb __crt_org_vector_table&$F0?
> Is it that __crt_org_vector_table is fixed during the library compile, and so end user compile can't change it?

The sio vectors must be first in the table.  That is guaranteed by crt4 since it includes the sio vectors in its source and will be the first thing the linker sees.

It would be better if the user could arrange the sio vectors where he wants them if he makes his own vector table.

aralbrec:
> And of course there’s demand for simultaneous ACIA and SIO support. Although the default base Port is 0x80 for both. TBD.

Well you can instantiate both drivers now and have them coexist in software easily.  A bigger problem is the ACIA is an im1 device while the SIO is implemented as im2.  You can create a fake vector for im1 devices by putting pullups on the data bus next to the z80 (vector id = 255) but otherwise the SIO drivers may have to have an im1 variant.

feilipu:
> The sio vectors must be first in the table...

IMHO, not necessarily. The only requirement is that the Channel B Register 2 contain the starting address for the SIO vectors. This doesn't need to be the starting address for the whole table.

> and the requirement is the table org must be aligned to 16.

Yes, this is true for these drivers, as I've configured the drivers to use 8 separate interrupt vectors.

Other solutions may configure a single interrupt vector and then read Register 2 within the interrupt to determine which routine is needed. But we aren't talking about other solutions. 😄

aralbrec:
>>The sio vectors must be first in the table...
>IMHO, not necessarily. The only requirement is that the Channel B Register 2 contain the starting
>address for the SIO vectors. This doesn't need to be the starting address for the whole table.

The requirement is not coming from the implementation, it's coming from this:

`defb __crt_org_vector_table&$F0`

The default vector table address is set by `__crt_org_vector_table = 0xFFF0` so the ISR table will be located at 0xFF00 (with `I` = 0xFF) with the SIO vectors at the fixed location 0xFFF0.

If someone wants the ISR table someplace else maybe with `__crt_org_vector_table = 0xFE00` or inside page 0 with `__crt_org_vector_table = -0x80` then the SIO vectors will be located at fixed locations 0xFE00 and 0x0080 respectively - ie the vectors must always be at the beginning of the vector table.

This is easy enough to remedy by adding a label to the start of the SIO vectors and using that in the defb instead.

>Other solutions may configure a single interrupt vector and then read Register 2 within the interrupt to determine which routine is needed. But we aren't talking about other solutions. smile

I don't argue with that - it should be done the way you did it.  I'm only thinking of how the user can define his own vector table without there being constraints on where the SIO vectors go.  There can only be one device at the beginning of a vector table so the method used can only be done once.

The idea for a user-supplied vector table would be a file looking something like this (and this is the first time we've had to worry about this so it's breaking new ground too):

```
section interrupt_vectors

vector_dma:
include "crt_interrupt_vectors_dma.inc"

vector_sio:  ; probably inside the include
include "crt_interrupt_vectors_sio.inc"  ; with an "align 16" at the front maybe

etc.
```

and then the defb becomes `defb vector_sio & 0xff`.

Now the sio vectors can be anywhere but you'd still want to weak default to something if the user doesn't do this.

PR #690, opened 2018-04-13
(classic) Generic console driver

suborb:
Based on the ZX control codes, provides a consistent behaviour and operation across multiple ports.

Still to do:

* [ ] More ports
* [x] Enabling this by default for some targets (pps, nascom, z1013, trs80)
* [x] Implementing the ioctl hook to allow screen mode changes.
* [x] Allow ioctl to provide size of screen
* [x] Docs on the wiki (First version https://github.com/z88dk/z88dk/wiki/Classic-GenericConsole)


suborb:
@zx70 do you think any more ports are worthwhile? / Any other targets where it should be enabled by default.

zx70:
need time to look at your work,   sorry I can't answer immediately

suborb:
I'm going to take this off PR for a while.

I'm going to attempt to share conio between the generic driver and ANSI - so gotoxy() etc all work the same.

So a few things will be changing and no point spamming whilst this is progress.

PR #691, opened 2018-04-13
Fixed Visual Studio x64 Build

mich181189:
The vcxproj files were missing some definitions and included on
the x64 targets. This meant they didn't build.


aralbrec:
As advertised, thanks.

PR #705, opened 2018-04-23
RC2014 - ACIA repository harmonisation

feilipu:
I've RC2014 ACIA code across 3 separate repositories, and it was beginning to drift apart.
Doing the SIO code also exposed a few weaknesses of the original ACIA code.

This harmonises the ACIA code across these three repos for RC2014.
- here
- [RC2014 CP/M-IDE](https://github.com/RC2014Z80/RC2014/tree/master/ROMs/CPM-IDE)
- [NASCOM Basic](https://github.com/feilipu/NASCOM_BASIC_4.7/tree/master/rc2014_NascomBasic32k_HexLoadr) with HexLoadr


aralbrec:
There are some direct uses of `di` and `ei` but since those routines require interrupts to be enabled for proper operation, it's ok to do that.  Otherwise the library routines `asm_z80_push_di` and `asm_z80_pop_di` would be preferred to preserve the di/ei state of the caller.

The declaration of aligned data is more problematic but I will probably revisit it all for all systems in the coming months anyway.  Sections like `bss_align_256`, etc, require data stored in them to fill up their entire 256 bytes (in this case) so that all objects added to the section remain aligned when the linker puts things together.  That's why there are other smaller sections `bss_align_128`, `bss_align_64` etc so that the data can be placed in the closest fit.

z80asm's `align` directive only applies locally within the file because it's applied to an address that has not been adjusted to its final location in memory by the linker.

So:

```
SECTION bss_align_256

ALIGN __IO_ACIA_RX_SIZE

PUBLIC aciaRxBuffer

aciaRxBuffer:   defs __IO_ACIA_RX_SIZE  ; Space for the Rx Buffer
```

The first `ALIGN` after the section probably does nothing because the file's address pointer for that section will start at zero and this is the first thing added.

Then the addition of the `defs` will mean stuff added to section `bss_align_256` elsewhere (and following sections `bss_align_128`, ...) will no longer be aligned.

To use these aligned sections, it's assumed the section is aligned already when you add to it.  Then you have to pad to make sure the alignment is preserved for following items.

So this is more correct:

```
SECTION bss_align_256

PUBLIC aciaRxBuffer

; already aligned to 256 here

aciaRxBuffer:   defs __IO_ACIA_RX_SIZE  ; Space for the Rx Buffer

; restore 256 align

ALIGN 256
```

This can potentially waste space so choosing the smallest aligned section is best.

Space maybe can be saved if a lot of data that would be needed together were brought into the same file so that they could be distributed to fill in empty space.

aralbrec:
There may be some changes to `ALIGN` as it may be a section attribute now:
https://github.com/z88dk/z88dk/wiki/Tool-zobjcopy

Anyway I will merge and then open this and the im2 thing as issues.

aralbrec:
> Ah. Didn't know that was a reliable assumption, and hence the ALIGN comes after. This should be fixed.

It's only reliable and workable if the DATA org (for data_align_*) and BSS org (for bss_align_*) are aligned to 256 because the data_align_* and bss_align_* sections appear first in the memory map for the DATA and BSS containers.

If the DATA and BSS sections are stuck together, as is typical with DATA org being set to the first ram address and BSS being appended to it, then this may not be the case for the bss_align_* sections.  If BSS is given its own org separate from DATA then it can be guaranteed that BSS is aligned by manually setting the ORG to an address that is aligned to 256.

For now, for the rc2014 target, it may be better to put these things into data_align_* sections instead and use the compressed rom model compile because the DATA org is aligned to 256 in rom compiles (0x8000).  These will be large 256-byte sections of zeroes added to DATA, which must be stored in ROM, but the compression (lz77) should make it small.

For the basic compile where code,data,bss are stuck together in memory, alignment can't be guaranteed but here it doesn't matter because the basic rom routines are used.

This is not really solved in a satisfactory manner but as it looks like sections may be getting an alignment attribute in z80asm, ( @pauloscustodio ) maybe this can be done properly soon.

Sections with names containing "_align_n", including user-defined ones, are special in appmake.  For +glue and other targets that use the banking model in appmake, the org of these sections are checked to make sure they are aligned and will emit warnings if they are not.  But the +rom target is not using the internal banking model now so there will be no warnings.

>At the moment the Rx buffer is assumed to be 0x100 bytes, even though there is a define 
>__IO_ACIA_RX_SIZE. i.e. if __IO_ACIA_RX_SIZE is not 0x0100 then the code will break. This is brittle, and I 
>need (at some stage) to add define code for the case where __IO_ACIA_RX_SIZE takes a different value.

Maybe this can be solved by using IFs:

```
IF __IO_ACIA_RX_SIZE >= 256
   SECTION data_align_256
ELSE IF...
```

> Should I do a PR for this small additional code?

Yes but we'll see if something can happen in z80asm to solve it satisfactorily.

pauloscustodio:
z80asm ALIGN works as follows:  
- If the ALIGN statement appears before any code in the section, it sets the section alignment attribute, which is honoured in the linker (and also in zobjcopy).  
- If the ALIGN statement appears in the middle of some code, then it aligs data assuming the block of code started at $0000 (not very useful).

pauloscustodio:
> This applies to the first time the linker sees the section? 
Yes. So if the section alignment is defined in the crt, all should work as expected. Please let me know if something is not right.

aralbrec:
One other question paulo:  if the aligned section is empty, it is removed entirely?

So with this memory map:

```
section stuff
section aligned_256
align 256
section more_stuff
```

If section `aligned_256` is completely empty then the linker will not add padding after section `stuff` for the alignment?

pauloscustodio:
It works as expected. Although section aligned_256 is empty, it is still considered during linking and the alignment is respected.

aralbrec:
Is it possible to have the section effectively removed if it is empty so that there is no padding added to the binary?

Right now the newlib defines aligned sections in its memory map but if the linker is going to pad for empty sections, then it's going to be increasing the program size and we wouldn't want that in there then.

The other option is to have the user supply sections that are included into the memory map so that the user is responsible for defining aligned sections.  This mechanism is there too so it could be used to avoid the unnecessary padding problem.

PR #708, opened 2018-04-23
Feature/zobjcopy

pauloscustodio:
Factor code of z80nm and zobjcopy into common module


suborb:
I don’t know how feasible it is but an option to disassemble the code sections would be useful - even without the code relocations.

suborb:
Yes, the disassembler does an instruction at a time and returns the length consumed.

PR #721, opened 2018-04-29
SP1 guide, pt 2, fix interrupts explanation

derekfountain:
Bit of rearrangement to get the custom ISR in earlier in order to bypass the issue with the Spectrum ROM  ISR. I've lost the simplicity of the first example, so I've added a bit more careful explanation of the use of the ISR.


derekfountain:
OK, I see you've merged this one.

I take your point about constant values being compiled in, as opposed to initialised by code. I'll change that in future examples. Also the *LOAD2* draw functions, which I'd forgotten about.

So much to learn... :)

aralbrec:
I hope you don't mind that I've moved the rectangle outside of main to the global vars... people will be looking at your code for best practice so I thought it was a good idea to change that.

I added some info about LOAD2LB and LOAD2RB to the exercise and then one more challenge to mix sprite types on screen.

derekfountain:
No, I don't mind at all. I should try that mixing up of sprite types myself since I haven't done that. :)

PR #723, opened 2018-04-30
YAZ180 - revert to ROM model

feilipu:
This PR reverts the YAZ180 to a flat 64kB memory ROM compilation model, which is suited for embedded use.

It is better for the standard YAZ180 Z88DK ROM model to be built this way, for an eventual z88dk v2.0.0 release, as it is more general than the specifics associated with the YABIOS drivers.

The YABIOS driver model had hacked the driver files to be located in a single file which was relocated on initialisation. The special YABIOS functions / files will be maintained separately, and will overwrite their defaults in the ROM model libraries as needed.

The APP compilation model (based on the YABIOS drivers) remains as is, with the origin at `0x0100`, and the provision of the YABIOS calling points as per normal.


feilipu:
@aralbrec it's done.

Issue #747, opened 2018-05-12
psg lib enhancement for sn76489


zx70:
So we now have a number of targets potentially able to play sound  via the sn76489 chip (sc3000, m5, mz800...).
That chip should also be able to generate simple crappy analog output in 4 bit as the SID does, and should be able in a similar way to do bit banging for our simplest way to support sound. 
The psg library is obviously the first one we want to have though, and the snakes.c example is simple and portable enough to test and compare the results.

PR #749, opened 2018-05-14
#747

zx70:


zx70:
@suborb ,   I'm still filling-in the proper expressions in psg.h, but during the tests I already noticed that "make -C psg deps" is not working as I expected.   ..hints?

PR #765, opened 2018-05-24
Sp1 animatedsprite doc

derekfountain:
Here's the next article in my SP1 series. Your usual proofread and comments/improvements are welcome. :)


aralbrec:
Just FYI, SevenuP can create sp1 sprites and it has tools that make it easier to edit masks.  I don't recall if it imports from animated gif let alone gif but probably not.

derekfountain:
Yes, I tried SevenuP but the most recent version is 6 years old and I couldn't get it to work with Linux, either from the supplied binaries or compiling from source. The version of WxWidgets it requires is so old it's hard to get working on a modern box.

I tried the Windows binary on a Win7 VM, but it just produced an empty grey window. Graphics driver issue maybe? I don't know much about Windows. I did get it working on an ancient XP VM I dug out, but by then I'd rather given up on it.

The next article will move on to bigger sprites so I'll probably need to solve the issue of a compatible graphics editor at some level.

suborb:
@derekfountain I'm finding that I have quite a bit of success using wine to run windows emulators and the like. I just tried SevenUp and it looked like it worked.

It's not ideal, but it's always a good backup option.

PR #768, opened 2018-05-24
Update ROM target for SC-3000 + SG-1000

suborb:
Remove code that calls the basic ROM - manipulate the VDP directly.

Scan keyboard directly - provide inkey() and friends.

Add the missing joystick function


zx70:
so fast  !?!!   impressive !

PR #799, opened 2018-06-05
YAZ180 - Am9511A APU 64kB banked device driver

feilipu:
Am9511A device driver revised to include the absolute bank (BBR) in the operand pointer buffer. Effectively becoming a far pointer to 64kB banks.

However to save time in the interrupt the operand pointer is stored by the operand load command as a ready-to-use BBR plus address, rather than storing a 20 bit linear address pointer.

This enables integrated load / unload of operands even in the banked yabios environment.


feilipu:
Fixed the C linkages for the Am9511A driver

aralbrec:
I guess this is ready to merge?

The c interface can have a minor tweak in one of the callee functions.  If you have this:

```
pop af   ;; return address
pop bc
pop hl
push af
```

where HL is the last param popped, you can do this:

```
pop hl   ;; return address
pop bc
ex (sp),hl
```

It saves two cycles and one byte (but maybe different on the z180) so not really that much except you get the pleasure of using `ex (sp),hl` :) 

The interface presents the possibility of the math library operating in synchronous or asynchronous mode.  Ie the c compiler could either wait for a result or the c program could indicate a block of code queues math instructions and then at the end of the block it waits for results or tests if results are ready.

feilipu:
Another thing needed to do is to build a floating point conversion library, that can convert to the APU format (which is different to the IEEE758 format), to IEEE758, and to other floating point formats.

There is an implementation of this in the LLL Floating Point library, as it used the same floating point representation. I’ll have to extract the conversion, and then add it to the z88dk libraries.

aralbrec:
> Just also confirming, that szcc80 is still left to right parameter pushing and minimum 16 bit parameters. I can’t believe that it would have changed, but there was a lot of work done recently on the library interface. So, just confirming.

Yes that's still the case.  sccz80 can actually change that to right-to-left and bytes for chars via command line option now and that's what will happen down the line, in the newlib at least.  Then there will be a large exercise in changing the entire sccz80 interface in the library which will be a lot of work but will make it more possible to mix sccz80 and zsdcc object code in a single compile.

> It could be interesting to allow the C compiler to generate APU intrinsic math functions. But, given the APU is obsolete and the YAZ180 is the only platform that supports it, I don’t see much return on that investment of energy.

Sometimes it's worth doing just because it's an interesting idea.

feilipu:
I have a few relevant Am9511A PDF documents, explaining the algorithms, data formats, and application to atomic energy research at Harwell, that I've searched out and would like to put in the techdocs archive. Separate PR for that.

feilipu:
@aralbrec think it’s done, now.
Any further mods will flow into z180 math connect PR.

PR #865, opened 2018-08-04
Filled in missing definitions in stdbool.h

michaelkamprath:
`stdbool.h` was missing definitions for `true` and `false`. This change brings `stdbool.h` inline with the C99 standard that `true` and `false` be defined. However, since z88dk doesn't have a `_Bool` type, I left the bool typedef pointing at an `unsigned int`. `bool` probably should be typedef'ed to an `unsigned char` for compactness, but changing that now might break existing code written to expect that a `bool` is two bytes rather than one.

### DIFF
--- a/include/stdbool.h
+++ b/include/stdbool.h
@@ -9,7 +9,15 @@
 
 #include <sys/compiler.h>
 
+// bool should really be an unsigned char, but because it was originally 
+// typedef'ed to an unsigned int in z88dk, leaving it as such as to maintain 
+// backwards compatibility.
 typedef unsigned int bool;
+
+#define true 1
+#define false 0
+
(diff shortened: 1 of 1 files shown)


aralbrec:
The build passes - it's just an initialization error on one of the compiles that fails.

The newlib has a version compatible with zsdcc:
https://github.com/z88dk/z88dk/blob/master/include/_DEVELOPMENT/proto/stdbool.h

michaelkamprath:
Do you want me to update my pull request to make `bool` an `unsigned char`?

suborb:
I’m away at the moment otherwise I’d do it. 

If you could that would be grand - I don’t think anyone is actually using it in any case. 

Im not sure how sdcc interop will work but I can look at that when I’m back.

PR #837, opened 2018-07-04
z80-zxn - integer math routines

feilipu:
Integer math routines using the z80-zxn `mul de` instruction, from #481, #312, and #820.

Used the z180 `mlt de` hardware for testing purposes.

- [x] `l_mulu_16_8x8`
- [x] `l_mulu_16_16x8`
- [x] `l_mulu_16_16x16`
- [x] `l_mulu_24_16x8`
- [x] `l_mulu_32_16x16`
- [x] `l_mulu_32_32x32` (using `l0_mulu_32_32x32`)
- [x] `l_mulu_40_32x8`
- [x] `l_mulu_64_32x32` (using `l0_mulu_64_32x32`)
- [x] `l_mulu_64_64x64` (using `l0_mulu_64_32x32`and `l0_mulu_32_32x32`)
- [x] `l_mulu_72_64x8`

I dislike the z80-zxn `mul d,e` mnemonic. The incorrect implication is that `mul` is done on `e` and is stored in `d`. Whereas `mul de`, or the z180 mnemonic `mlt de`, more obviously and correctly refers to **both** the **d** and **e** registers being modified by `mul`

PR #878, opened 2018-08-18
z80 - small 32_16x16 multiply

feilipu:
As noted in PR #837, the small 32_16x16 was implemented by setting the high words to zero, and using the small 32_32x32. This results in an inefficient outcome.

This PR implements a straightforward `l_small_mul_32_16x16` that provides a 3x improvement in throughput, but remains small in comparison to the previous impementation (not least because the zero setting code falls away).

The original code took up to 3,631 ticks.
This PR requires 1217 ticks.


aralbrec:
Good add :+1:

PR #920, opened 2018-08-29
Integrate z80asm regression tests in global make test

pauloscustodio:
Add a cpm submodule to be able to run enigma.c and verify the expected result automatically. This is called from z80asm regression tests. 

Had to fork cpm to modify it to run in cooked mode and in Win32 if a raw terminal is not available.

Add the build of zsdcc and zsdcpp in the main Makefile, unless the tools already exist in bin. This allows Windows users to copy the binaries to the bin directory and the build is bypassed. 

Remove "rm -rf bin" from build.sh on the clean option - it deleted the zsdcc and zsdcpp.


suborb:
I'll need to change the nightly build to copy the binaries in before the build starts - so the nightly is broken at the moment.

suborb:
Okay, should all be sorted now.

@pauloscustodio just out of interest why did use CP/M rather than the ticks/+test target?

pauloscustodio:
I did not know I could use ticks to read standard input and write standard
output, to be able to run an automated test.
How does it work with ticks?

suborb:
Fair enough!

Should be the same as the CP/M emulator - ticks just reads/writes from stdin/stdout. Command line arguments are also supported.

pauloscustodio:
Thank! Good to know!

pauloscustodio:
Added a test with zcc +test and ticks.

PR #936, opened 2018-09-02
correct header2 in vz700 tape format

nippur72:


suborb:
Ooops, quite a misread by me there...

PR #941, opened 2018-09-03
+vz700 fix loading of tape files

nippur72:
switched from using "T" type files (Basic programs) to "B" type (binaries). This because the vz700, after loading a "T" program, rebuilds the Basic pointers thus corrupting the z88dk code. 

With "B" there are no problems, files can be `CLOAD`ed and `RUN`.


nippur72:
sorry wrong branch

PR #987, opened 2018-09-24
+G800 monographics

ashitani:
Added monographics support to g800 (Sharp PC-G8xx, PC-E2xx)
These functions are checked (only checked on the g800 emulator).

- plot() / unplot() / xorplot()
- point() / multipoint()
- draw() / undraw() / xordraw()
- drawto() / undrawto() / xordrawto()
- drawr() / xordrawr() / undrawr()
- clg() / clga() / fill()
- circle() /  uncircle()
- drawb() /  undrawb() / xorborder()
- getmaxx() / getmaxy()


suborb:
Thanks for this, could I ask for a couple of changes:

- Can you split plot.asm into separate files for plot/res/xor?
- Likewise with draw.asm - can you make each entrypoint a separate file?

This way only those routines that are needed will be linked in which reduces the binary size.

There's a few variables about the place - could you place them into the bss_clib section - that way they get reset every time the program is run.

aralbrec:
The linker works on a per file basis.

In `draw_main.asm` several symbols are exported, in particular `call_plotpixel`, `call_unplotpixel` and `call_xorplotpixel`.  These seem to be independent of `draw_main` but if anyone calls them, they will get all the draw code added to their binary even if they are not using draw.

There is a similar situation in `pointxy.asm`.  `point` and `pointxy` seem to be independent yet the output binary will contain both if either one is called.  There are several other things in this file that may also be independent; other functions in other files may be calling `getpat`, eg, so the binary will contain `point` and `pointxy` even if the program does not call those functions.

It does add up on a library-wide basis so we try to whittle things down to the minimum to keep the binary minimum size.

ashitani:
Thanks to your kind instruction. I'll try.

ashitani:
I have removed `call_plotpixel` and so on, and separated all independent entry points into files.

aralbrec:
Looks good.. maybe @suborb or @zx70 can try it out.

PR #988, opened 2018-09-25
Changed character input/output to handle DELete

UncleBod:
Changed standard origin to 1200H
Changed so DElete key on MZ-700 is handled correctly.

Tested with https://sourceforge.net/projects/mz800emu/ and works there.
(I don't have any physical MZ-700 any more.)


UncleBod:
Updated.
Had an '-' in the beginning of mz_crt0.asm

PR #993, opened 2018-10-01
First step for MZ80B

UncleBod:
Machine mz80b added
For now most files are identical to machine mz
For the moment it only works with basic SB5510 loaded.
I am working on a stand alone version


suborb:
It looks like this change doesn't contribute to a library and is just a straight copy of the code? 

If that's the case I'll hold of merging until there's a functionality change.

zx70:
I really don't want to loose your contributions, I propose to change the approach on the MZ80B target.   Let's get back to the forum for a discussion on how to feed programs to this specific target, we may need to start from appmake and think at the compiler and libraries skeleton in a second time.

PR #997, opened 2018-10-07
remove unnecessary signature

nippur72:
I don't know if there is any purpose to the signature `Small C+ VZ`, if it's safe let's remove it saving the bytes.


suborb:
It doesn't do much - when I had a much slower machine I used to figure out what target I'd compiled a binary for.

PR #998, opened 2018-10-08
[g800] modified monographics addressing and add examples

ashitani:
Modifed g800 monographics addressing bugs discussed [here](https://github.com/z88dk/z88dk/commit/d3bb5f0d7d88b622c6b3ae0a814c6939a0e900b9#commitcomment-30729937). Checked on [g800](http://ver0.sakura.ne.jp/pc/) emulator and a real PC-G850V machine.

And added some examples for PC-G850.
physics_demo


zx70:
thankyou!
there are few details I'd like to have different but overall it is definitely interesting.  Don't be shy and leave your name in the pieces of code you mostly are proud of and please go on hacking   :)

PR #1000, opened 2018-10-08
remove unnecessary signature

nippur72:
As for `+laser500`, this removes the "Small C+ VZ" signature string for `+vz` targets

### DIFF
--- a/lib/target/vz/classic/vz_crt0.asm
+++ b/lib/target/vz/classic/vz_crt0.asm
@@ -132,10 +132,6 @@ l_dcal:
         jp      (hl)
 
 
-
-	defm  "Small C+ VZ"
-	defb   0
-
         INCLUDE "crt/classic/crt_runtime_selection.asm"
 
 	INCLUDE "crt/classic/crt_section.asm"


zx70:
I thing that choosing whether to keep those signatures (present on almost all the targets) should be a general decision rather than a one-by-one hunt.
Those few extra bytes (12, in this case) are harmless, in my opinion.. the main reason for having them in is, in my knowledge just to recognize the z88dk compiled code.
This is quite frequent in the compilers, it could be of some interest for future external tools, and yes.. probably it is useless.
In some case I think I used that space somehow,  IIRC there's a trick on the ZX81 target.
Opinions ?

zx70:
@aralbrec, @suborb ?

nippur72:
if you want to keep signatures, at least let's put them under an IF.

suborb:
I'm happy to remove them all.

suborb:
Closing to make way for #1002

aralbrec:
Yeah I'd say remove them too.  I decided not to use them in the newlib but did retain a z88dk version number in page zero code where nothing else would fill the gaps between restarts.

zx70:
ok

Fabrizio-Caruso:
Thanks! 12 bytes saved can make a difference in some very memory-tight situations. Sometimes I do indeed struggle for an hour to save just a few bytes.

PR #1010, opened 2018-10-20
Add replacement rs232_get and rs232_put routines which pull the IF1 R…

tschak909:
…OM routine, and patch it to not react to BREAK signal. Thanks to Owen Reynolds.

This makes the IF1 RS232 routines usable for terminal program type applications.

As Owen doesn't have much experience with git, I've decided to fork and push this change, but all credit should go to him.

-Thom


suborb:
Cool thanks for doing this, removing the break check and the border colour changes is great.

Issue #971, opened 2018-09-14
Upcoming SDCC 3.8.0 release


spth:
SDCC developers are working on a SDCC 3.8.0 release. There will probably be a RC1 next week followed by a release the week after. If any serious issues are found in SDCC, the release might be delayed.

I guess now and next week would be a good time for z88dk developers to test current SDCC trunk to see if there are any serious regressions or other bugs that affect z88dk.

Philipp

spth:
There is now a release candidate 1; unless issues that important enough to delay the release are found, the final 3.8.0 release will be in 10 days.

Philipp

aralbrec:
I still had subtle bugs in the last update I tried but I think they may be in our peephole rules.  I think the way to move forward is to get sdcc's regression tests running with our version zsdcc to help locate where these problems may be.

Fabrizio-Caruso:
@aralbrec 
Issue #817 is not relate to peephole rules: it is fixed by --max-alloc but it is present with no optimization. It may be related to some ZX81 routines, though.
Issue #819 is only occuring if sdcc is used with --opt-code-size. Only some targets seem impacted though (vz200, svi and maybe others).

aralbrec:
I found a bug in sdcc z80 peephole rule 10:

```
replace restart {
	ld	%1, %2
	ld	%3, %4
	ld	%5, %1
} by {
	ld	%5, %2
	; peephole 10 loaded %5 from %2 directly instead of going through %1.
	ld	%3, %4
} if canAssign(%5 %2), notVolatile(%1), operandsNotRelated(%1 %4), operandsNotRelated(%1 %3), operandsNotRelated(%4 %5), notUsed(%1), notSame(%3 %4 '(hl)' '(de)' '(bc)'), notVolatile(%5)
```

This is being erroneously applied to this code fragment:

```
ld a,(hl)
ld hl,(#_foo)
ld h,a
```

with:

```
%1 = a
%2 = (hl)
%3 = hl
%4 = (#_foo)
%5 = h
```

I'm testing a replacement with different condition:

```
if notUsed(%1), canAssign(%5 %2), notVolatile(%1), notVolatile(%5), operandsNotRelated(%1 %3), operandsNotRelated(%1 %4), operandsNotRelated(%2 %3), operandsNotRelated(%3 %5), operandsNotRelated(%4 %5), notSame(%3 %4 '(hl)' '(de)' '(bc)')
```

spth:
[...]

PR #1014, opened 2018-11-11
ZSDCC - patch for z80 in + z180 out0

feilipu:
I think a miss-count of characters for the patch of the sdcc for z80 `in` instruction and z180 `out0` instructions. Should be looking at 3rd and 5th characters respectively.

The `in` instruction is not followed by a `\t` character in the current SDCC revision 10649, hence I believe that looking for the third character is correct.


aralbrec:
`\t` must follow `in` in peep.c otherwise the function cannot tell the difference between `in` and other instructions with `in` as prefix as in `ini`, `inir`, `in0`, etc, unless a strict order of testing is made.  `inir` must be tested before `ini` which must be tested before `in` to avoid misclassifying an `ini`, eg, as an `in` instruction.

I don't know if current sdcc is bugged in that regard but I'd rather not depend on the order of testing to correctly identify which instruction is parsed.  This function in particular is bugged in sdcc in other ways and has been for a long time - it causes the sdcc peepholer to fail to apply all peephole rules that it can; this is hardly noticeable in sdcc itself as it has a small set of peephole rules but it is very clear under z88dk which has a much larger set with many interactions between rules.

The changes in index values don't affect the correctness of what's there now.  Looking for a comma from index 4 in an `in` instruction is fine because all `in` instructions are of form `in r,...`.   For `out0` there is no other instruction with `out0` as prefix so it should be ok to search for comma after the `out0` prefix.  But maybe this can be changed in an update to current sdcc as mentioned in another issue.

feilipu:
Ok. I've added the whole patch to #971.
Since the sdcc peephole is not used in z88dk, this issue is not important.

But perhaps the rest of the patch might help to move onto 3.8.2 as a new base.

PR #1024, opened 2018-12-07
ZSDCC - updating patch for SDCC 3.8 - do not merge

feilipu:
As @spth is adding the zsdcc modifications into the trunk of sdcc, this PR reflects the deletions from the patch file in zsdcc.

- [r10712](https://sourceforge.net/p/sdcc/code/10712/)
- [r10711](https://sourceforge.net/p/sdcc/code/10711/)

Current status is [version 10748](https://sourceforge.net/p/sdcc/code/10748/log/).
Updates until cleared from issue #971.


feilipu:
Travis failed because the `sdcc-z88dk.patch` is being applied to r9958, rather than to the most recent version. I had hoped to maintain this as a delta, but probably better to back out this change, so that Travis can work.

feilipu:
I have split the patch into two pieces.

 1. [z88dk specifics](https://github.com/z88dk/z88dk/pull/1024/files#diff-455538ec8a04e1da86990e149262a59c) - the minimum required to compile within z88dk with typical `zcc` command lines.
 2. [Improvements to sdcc `peep.c`](https://github.com/z88dk/z88dk/pull/1024/files#diff-b2a2cf83665c14710a113de7adc5fa75) - will be submitted to the sdcc repository separately, for @spth to handle. Code generated by `zsdcc is not affected by this patch, because the sdcc peephole optimiser is not used.

Please review to see if we can move to sdcc 3.8.x. from r10761.

suborb:
@feilipu I'm not sure what's happened here, I merged the CPM one and this one has gone in as well!

Ah well, probably need to generate new builds for osx and win32. I'll do osx later and then hopefully I'll be able to do win32 as well

feilipu:
@suborb I messed up with git and the needed files were deleted by #1066.

But really to move to 3.8.x or 3.9.0 the ZSDCC specifics patch needs to become the default patch.

And the SDCC peephole patch is just remaining for reference, until integrated SDCC itself.

feilipu:
I can fix this in a few days. AFK currently. 

But my fix will always fail Travis because it will be building against r9588 and the patch application will fail. 

The ZSDCC specifics patch is for r10788 or later. So you’ll need to flip both at one Commit.

suborb:
Looking forward to it again. You should be able to modify the top level makefile to update the SDCC version which should fix the travis build.

PR #1033, opened 2018-12-19
Galaksija hires

suborb:
Support the hires graphics for the Galaksija Plus model.

Wiki needs updating before merge.


zx70:
Dom, I think the "gfxbase" default inclusions should stay on top.  The subsequent entries will override it.

suborb:
Thats what I thought originally. 

However it’s first match - hence why user libraries will override system libraries.

zx70:
ugh... I probably need to repeat my experiments, that's how I finally determined the priority.   Nevertheless I'll probably stay quiet I  the next days for obvious reasons  :)   🎄

suborb:
It's easy to spot with getmaxx() - for those platforms with different screenmodes it needs to go to the platform specific version rather than the general purpose one - you'll end up not being able draw a pixel border if it's in the wrong order!

PR #1066, opened 2019-01-16
[RC2014] - add CP/M subtype

feilipu:
Add a `-subtype=cpm` option, leaning on the code in the new library CP/M target.

Fix [an inconsistency with he font path](https://github.com/z88dk/z88dk/compare/master...feilipu:rc2014?expand=1#diff-0e3800521f04e92732d35d1e88b7388cL11) in CP/M target. This one is not like the others. Or, are the other two wrong?

Remove some whitespace in passing.


feilipu:
Except for the inconsistent font path in the CP/M target that @suborb could comment on, this PR is all tested working with RC2014 hardware. So AFAIK, everything is OK.

I'd like to get some other RC2014 user input, which which will be easier if the PR is merged.
Thoughts?

libsrc/_DEVELOPMENT/target/cpm/library/cpm_sdcc_iy.lst
```diff
@@ -8,7 +8,7 @@
 @fcntl/fcntl_sdcc_iy.lst
 @font/font_4x8/fonts.lst
 @font/font_8x8/fonts.lst
-@font/fzx/fzx_sdcc_iy.lst
+@font/fzx/zx/zx_sdcc_iy.lst
```
suborb: FZX is a ZX only library so I don't think these files are actually needed.

suborb:
@feilipu I'm going to promote a nightly to release this weekend (https://www.z88dk.org/forum/viewtopic.php?id=10704) so I'm going to hold off on this merge until that's done.

Once that's done I'll click merge on this.

I'd also like to get the newer SDCC in after that (with a proper release we can afford to have potentially defective nightlies for a while)

libsrc/_DEVELOPMENT/target/cpm/library/cpm_sdcc_iy.lst
suborb: Actually, given that it's CP/M I'm not sure there's much point in the other fonts either since there's no graphics functionality in newly

feilipu: I don't know, but there is inconsistency between the scz80 target and zsdcc_ix target paths for this library, and the zsdcc_iy target path. One of them must be right, and the other is wrong.

feilipu: Then, remove them all for all the cp/m targets seems like the relevant solution?
For this PR, that won't any effect, as only the referring `.lst` file is included in the core of the PR.

feilipu:
@suborb unfortunately I messed up with git and tried to delete the effects of #1024 on this patch to keep things clean. 

Net effect is applying this PR deleted the relevant files needed for #1024. Being the patch for SDCC, and the remaining peephole patches that will be (perhaps) applies by @spth in a separately submitted SDCC patch.

The best result is for the patch file for ZSDCC to be recovered and used instead of the current r9588 one. 

Sorry for being bad at git.

PR #1074, opened 2019-01-24
zsdcc - update & improve readme

feilipu:
A numerical typo crept into the zsdcc `readme.md` file, so this is a correction and slight expansion to aid understanding.


suborb:
Thanks! Fat finger syndrome from me there.

PR #1109, opened 2019-02-17
Fixes to Makefiles for MinGW

pauloscustodio:
Fixes to pass both msys-1.0 and msys2 versions of MinGW on Windows.

The way to detect how a path is translated from Unix-like to Windows-like in lstmanip.c is brittle, but should work in a standard installation of either the old MinGW (http://www.mingw.org/) or the newer msys2 (https://www.msys2.org/). The problem is how each system translates the drive letter to upper or lower case, and how the Makefile needs an exact-case to replace the source path to the object path.


suborb:
Thanks Paulo. Lstmanip is only used for fast math now so I should be able to get rid of it pretty soon.

pauloscustodio:
There seams to be a problem with the Travis build in z80asm. I'll check it before merge.

pauloscustodio:
@suborb, I've merged your remove_lstmanip with my Makefile updates. Waiting for Travis...

Makefile
```diff
@@ -77,56 +79,52 @@ bin/zsdcc$(EXESUFFIX):
 	cd $(SDCC_PATH) && mv ./bin/sdcpp $(Z88DK_PATH)/bin/zsdcpp
 	$(RM) -fR $(SDCC_PATH)
 
-bin/appmake$(EXESUFFIX):
+bin/appmake$(EXESUFFIX): $(wildcard src/appmake/*.c src/appmake/*.h)
```
suborb: Maybe we should go back to .PHONY targets here? I'm not too keen on this top-level makefile having incomplete dependencies for the binaries.

By incomplete I mean they don't reference the modules that are used.

pauloscustodio: Agree with .PHONY. I'll revise and commit.

suborb: Thanks Paulo, sorry I was at fault changing it initially

Issue #980, opened 2018-09-17
Porting Hi-Tech C 32 bit Float library to zxn & z180 - question


feilipu:
Reading the code discussed in #972, it looks quite simple (but still a lot of work), to convert the Hi-Tech C 32bit floating point libraries to the format required by SDCC, and use them within z88dk.

And, also it should be possible utilise the multiply hardware on the zxn and z180 to speed up [`eval_poly`](https://github.com/feilipu/yaz180/blob/master/docs/Hi-Tech_C_Libraries/FLOAT/evalpoly.c) and other routines, where the hardware is available.

It is an interesting task, since the Hi-Tech C float format also looks like that of the Am9511A-1 APU, except that the Hi-Tech C exponent is biased by 64. So, again it would be quite simple to patch in the APU, where it was available.

This looks like an interesting diversion for a few weeks. But first a question...

Where is the z88dk patching code to match the Gen48 floating point library into the required SDCC 32bit (nearly) IEEE-754 format? I've poked around, but I can't find it.

aralbrec:
The root of math48 is here:
https://github.com/z88dk/z88dk/tree/master/libsrc/_DEVELOPMENT/math/float/math48

`z80/core` contains the original math48 implementation with some bugfixes.

`z80` contains additional code for meeting what's expected in math.h.  These functions define the asm interface to math48 and still operate on floats in the format math48 defines.  They call into functions in `z80/core`.   Functions are prefaced with "am48_" for "asm math48".

`c` contains the interface expected by math.h and the c compilers.  These functions do the conversion to/from the float format expected by the compiers and call into the "am48_" code to carry out the operations.  All functions are namespaced with "cm48_".

As an example, [cm48_sdcciy_asin_fastcall.asm](https://github.com/z88dk/z88dk/blob/master/libsrc/_DEVELOPMENT/math/float/math48/c/sdcc_iy/cm48_sdcciy_asin_fastcall.asm) executes "call am48_asin" to carry out the actual operation but this call is bracketed by functions that do the floating point conversion "cm48_sdcciyp_dx2m48" and "cm48_sdcciyp_m482d".  They contain "sdcc_iyp" in their names means sdcc_iy library primitive.  Some primitive functions in this directory also implement compiler float primitives expected by the compiler.

At this point math48 is namespaced so that it can exist in all target libraries without interfering with any other float implementations also in the target libraries.  It's possible to call several different float libraries in the same program.

[...]

Issue #972, opened 2018-09-15
Hi-Tech C Libraries Source Code - Informational


feilipu:
It may not be news for anyone but me, but I've put the Hi-Tech C v3.09 Libraries Source Code in my YAZ180 repository.

It was quite a bit of work to get them into the modern world, via both a DOSBOX and CP/M DEHUFF tools, and I haven't seen these files elsewhere, so I thought it worthwhile to post them.

Why are they interesting?

There is a[ CP/M](https://github.com/feilipu/yaz180/tree/master/docs/Hi-Tech_C_Libraries/CPM) implementation, with [stdio file access](https://github.com/feilipu/yaz180/tree/master/docs/Hi-Tech_C_Libraries/STDIO) complete.
The fast (but inaccurate) [32 bit float](https://github.com/feilipu/yaz180/tree/master/docs/Hi-Tech_C_Libraries/FLOAT) implementation.
And, [generally understanding the Hi-Tech C code generation](https://github.com/feilipu/yaz180/tree/master/docs/Hi-Tech_C_Libraries/GEN), and what the C compiler expects.

I've also [attached a ZIP of the files to this issue](https://github.com/z88dk/z88dk/files/2385343/Hi-Tech_C_Libraries.zip), in case there is interest to store them here in the docs section too.

aralbrec:
Where did you find the source code?

feilipu:
I understand that the Library source, and the source for the `c.com` tool, was distributed when the Hi-Tech C compiler was put into the public domain.

```
The HI-TECH Z80 CP/M C compiler V3.09 is provided free of charge for any
use, private or commercial, strictly as-is. No warranty or product
support is offered or implied.

You may use this software for whatever you like, providing you acknowledge
that the copyright to this software remains with HI-TECH Software.
```
The file I started with was [`LIBSRC.EXE`](http://www.retroarchive.org/cpm/cdrom/SIMTEL/HITECH-C/). The annoying thing about the file is that it is a DOS self expanding archive. The expanded archive then needs to be `DEHUFF`ed using the tool in the `Z80V309.EXE` file, on a CP/M machine. So there's a bit of shuffling back and forward. Also of note I changed the file extension of the asm files from `.as` to `.asm`, and copied the C header files into the directories from the compiler executable directory for easy reference.

[...]

Issue #305, opened 2017-08-08
z80asm: add directive to create floating point values


aralbrec:
In addition to defb for bytes, defw for ints, etc it would be useful to be able to declare floating point values.

There can be many different floating point formats.  The two main ones are genmath and math48 formats which are 48-bit floats as these are z88dk's floating point libraries.  However some targets have their own native float libraries with different formats, eg the zx spectrum has a 40-bit fp format.

I'm not sure what the best way is to deal with this but I do know that it's useful to be able to do something like this so I'll leave this feature request here for contemplation.

pauloscustodio:
I think this should be implemented with two commands:
1. One to define the float library to be used in the code as a command line option
2. One to declare a list of floats in the current format

An assembler macro should indicate the current format in use, i.e. __FLOAT_MATH48__, ...

This way, the same source code can be used for different float types by changing the command line options.

Suggestion:
```--float=genmath|math48|zx|zx81 ...``` and maybe others to define the float type for the source
```float expr, expr, ...``` to compute the compile-time expressions and store the float value; the expressions are calculated in the assembler as ```double``` entities and the result converted to the target float type

Any comments? What should be the default float type?

Can someone supply pointers to documentation describing the different float formats?

feilipu:
Just for reference, there is a good [summary of floating point formats](http://www.andreadrian.de/oldcpu/Z80_number_cruncher.html#mozTocId774334) by Andre Adrian, with some [references](http://www.andreadrian.de/oldcpu/Z80_number_cruncher.html#mozTocId485748).

I'm personally interested in the [Lawrence Livermore Laboratory floating point package](https://github.com/feilipu/LLL-Floating-Point) from 1975, [UCRL-51940, Floating-Point Package for Intel 8008 and 8080 Microprocessors](https://github.com/feilipu/LLL-Floating-Point/blob/master/LLL%20Floating%20Point%20Basic.pdf), because this was the same format used by the Am9511A Arithmetic Processing Unit. It predates IEEE754, by a year or so.

It is on my list to integrate a [driver for the Am9511](https://github.com/feilipu/LLL-Floating-Point/blob/master/am9511a.asm) into z88dk. It is done, but I'm working on some other things currently, so #64 is not a high priority.

So I'd like to see at least the following options
- Math48 - 48 Bits (Z88DK)
- GenMath - 48 Bits (Z88DK)
- Z80 - 40 Bits
- IEEE754 - 32 Bits (SDCC)
[...]

Issue #120, opened 2017-03-20
z88dk: implement a fast ieee 32-bit float type


aralbrec:
I may do this at some point but this could be a project taken on by anyone interested.
There were some surprising results in the benchmarks related to this.

Hitech C v309 for CP/M (a native z80 compiler) has by far the fastest 32-bit float implementation.  Its performance best manifested in the [mandelbrot benchmark](https://github.com/z88dk/z88dk/tree/master/libsrc/_DEVELOPMENT/EXAMPLES/benchmarks/mandelbrot) with an execution time almost half that of the next 32-bit implementation.  z88dk's float library is 48-bit so you can't quite compare apples to apples but even taking into consideration the 24-bit mantissa vs 40-bit, the hitech implementation is still likely a faster implementation.  In any case, it shows that a 32-bit library can be significantly faster.

The problem with the Hitech implementation is that it is not usable for anything except add, subtract, multiply and divide.  The transcendental functions have large errors associated with them.

How is it so fast?  I think the multiply and divide operations are being done with integer math.  Both of z88dk's 48-bit float implementations are preferring to keep the mantissa in a sequence of registers for which multiply and divide are done with sub-optimal shift & add/sub.  It's sub-optimal because holding values in certain registers allows integer math to be much faster for the same operations.  The other compilers' 32-bit math implementations barely outperform z88dk's 48-bit which tells me they are doing something similar.

z88dk has some very fast integer math, especially in the new c library which has a fast math option.  This integer math can be more than 3x faster than Hitech's (see the [pi benchmark](https://github.com/z88dk/z88dk/tree/master/libsrc/_DEVELOPMENT/EXAMPLES/benchmarks/pi)) if you're willing to pay the increased code size.  I think making use of this integer math could lead to a very fast float package.

Everyone wants IEEE so doing that at the same time may be the way to go.

Anyway this is a reasonably big project so it will likely fester here for a while until / unless someone takes an interest in the challenge.

feilipu:
It is certainly an interesting and worthwhile challenge.

I'm quite interested to help with this, but not sure that I have enough experience to be able to lead it.

Simply researching the different 32bit formats, and how to convert between them effectively is sufficient for me to chew currently.

suborb:
Here's the Rabbit implementation (mostly IEEE compliant):

[...]

PR #1121, opened 2019-03-15
m32 - digi international & Hi-Tech float package

feilipu:
Building a 32 bit IEEE (mostly) standard floating point math package, to work with SDCC 32 bit (mostly) IEEE standard interfaces.

Where not written by me, sourced from:
 - the [Digi International Rabbit IEEE 754 32 bit library](https://github.com/digidotcom/DCRabbit_9), copyright (c) 2015 Digi International Inc.
 - the [Hi-Tech C 32 bit floating point library](https://github.com/feilipu/yaz180/tree/master/docs/Hi-Tech_C_Libraries/FLOAT), copyright (C) 1984-1987 HI-TECH SOFTWARE.
 - the [SDCC 32 bit floating point library](http://svn.code.sf.net/p/sdcc/code/trunk/sdcc/device/lib/), copyright (C) 1991 by Pipeline Associates, Inc, and others.
 - various Wikipedia references.

Related to #120, #64, and peripherally #305, #980 and #972.

To do:
- [x] - write z80, z180 and z80-zxn function for multiply
- [x] - write z80, z180 and z80-zxn function for invert, and divide
- [x] - write z80, z180 and z80-zxn function for add and subtract
- [x] - write z80, z180 and z80-zxn function for inverse square root, square root, and square
- [x] - write z80, z180 and z80-zxn function for polynomial evaluation
- [x] - write z80, z180 and z80-zxn function for hypotenuse
- [x] - convert to sccz80 calling convention (stack left, DEHL right)
- [x] - implement fastcall and callee calling convention alternate entries
- [x] - build connectors to asm and c libraries
- [x] - integrate Hi-Tech C source files for trigonometric and logarithmic functions
- [x] - integrate SDCC & GCC C source files as needed
- [ ] - build linkages to create library
- [ ] - build missing functions where needed
- [ ] - testing, and bench-marking

To be decided:
- [x] - how to connect the library, suggestion `-lm32`


feilipu:
Looking at the rabbit instructions, for z80 equivalents.

`ld hl,(sp+n)`

`bool hl`

`or hl,de`

`rr rp`
```asm
rrc rpl 
rl rpl
rr rph
rr rpl
```
`rl nn`
```asm
rlc rph
rr rph
rl rpl
rl rph
```
but the code comments always talk about "shifting" the register pair, so perhaps given we don't have an intrinsic shift instruction, it is sufficient (and better) to do this...
`srl rp`
```asm
srl rph
rr rpl
```

`sll rp`
```asm
sla rpl
rl rph
```

`ex de,hl'`
```asm
push de
exx
ex (sp),hl
exx
pop de
```

`ex de',hl`
```asm
push hl
exx
ex de,hl
ex (sp),hl
ex de,hl
exx
pop hl
```

`mul` is a signed 16 bit multiply HL:BC = BC • DE, but the code comments talk about working unsigned, and removing the signed components. it will probably be better to understand what it is doing, and replace it by the z88dk library 32 bit multiply, and shift. Or, just write our own 24 bit version.

suborb:
This seems like a good motivator for me to add 4 byte float support to sccz80 and fix probably the last sdcc/sccz80 interop issue.

It makes sense to re-use the long style interface - dehl for holding the value for functions and the other value on the stack for the internal functions.

feilipu:
Just noting as a reminder the @aralbrec comment on #64 that the SDCC exponent bias is different to the [IEEE bias](https://en.wikipedia.org/wiki/Exponent_bias) here.
SDCC = 126
IEEE = 127

I don't know why SDCC didn't follow standards...?

From `float.h`
```c
/* the following deal with IEEE single-precision numbers */
#if defined(__SDCC_FLOAT_LIB)
#define EXCESS		126
#define SIGNBIT		((unsigned long)0x80000000)
#define __INFINITY	((unsigned long)0x7F800000)
#define HIDDEN		(unsigned long)(1ul << 23)
#define SIGN(fp)	(((unsigned long)(fp) >> (8*sizeof(fp)-1)) & 1)
#define EXP(fp)		(((unsigned long)(fp) >> 23) & (unsigned int) 0x00FF)
#define MANT(fp)	(((fp) & (unsigned long)0x007FFFFF) | HIDDEN)
#define NORM            0xff000000
#define PACK(s,e,m)	((s) | ((unsigned long)(e) << 23) | (m))
#endif
```

feilipu:
Too early to celebrate, but 1/3 of the basic functions are working. Multiply tests out!

It was easier in the end to rewrite the Digi multiply using z80 specific op codes, but following mostly the same algorithm. The multiply uses `mulu_32_16x16` functions, implemented for z80 using `l_mulu_32_16x16`, z180 using `mlt nn`, and z80-zxn using `mul de`.

This multiply function is also needed for the divide routine, so it will be reused there too.

suborb:
I've pushed a branch that switches sccz80 to 4 byte IEEE floats. This is enabled via the -fp-ieee option.

There's a bit more work to do, but the basics should be there.

The interface is as described above dehl for the long with the alternate value on the stack. The required functions are as per the other float libraries for sccz80 (dadd, dge, minusfa etc).

feilipu:
@suborb I think the sccz80 binding won’t be significantly different than the this [sdcc](libsrc/_DEVELOPMENT/math/float/mathd32/c/sdcc/cd32_sdcc_dread.asm) binding, provided the right to left calling convention is used. Both would push 4 bytes little ended onto the stack, right?

When I get all of the basic functions completed and tested, it would be good to have some help to connect the library in a flexible way. I was thinking to use `-lmd32` for this Digi International library. Mainly because I am also planning to add the Hitech-C 32 bit library at some stage, and that would be `lmh32`. The original sdcc math library could then be `-lms32`.

suborb:
That's where I ended up in a tangle. I forgot that the sdcc float interface is basically a C interface with right and left values pushed on to the stack from right to left.

The sccz80 float interface is left value on stack, right value in registers which is a bit more efficient. Once you've done the sdcc I can easily build the bridge to sccz80.

feilipu:
The Digi interface is more like the sccz80 interface with the left value on stack and the right value in `bcde`. So the sccz80 interface should be really efficient, possibly only needing to rearrange the right register value.

I've learned this is is very useful for `_fsdiv`, where the right value is simply inverted in registers and then sent to `_fsmul` to calculate. `z = x/y = x * 1/y` where x is on stack and y in `bcde`.

Just untangling the `_fsdiv` now. Hopefully within a few days, and best case even tonight.

feilipu:
Found that `_fsdiv` uses the alternate register set extensively, and that is just not efficient on the Z80 and Z180. Also in the early calculations for w[1] `16_8x8` multiplications are sufficient, and these are native for Z180 and Z80-ZXN.

Removed preservation of `de` from `mulu_32_16x16` as I don't assume it is provided.

And, rewrote the `_fsdiv` to use the stack (not alternate registers), and native sized `mlt` routines.
Still not accurate. Must be some math error somewhere.

feilipu:
I'm not seeing it.

Sign and exponent calculations are working well. But the mantissa Newton-Raphson doesn't converge.

I'm pretty sure I'm making a simple mistake with the decimal point in the binary math.
`w[n+1] = w[n] + w[n] - (w[n] * w[n] * y)` where `w[0] = 1/y` estimate found in lookup table.

Testing with the w[0] seed table and two iterations produces mantissa values all over the shop.
I need to get this right, before I can perfect the w[3] calculation.

feilipu:
This is harder than I thought it would be. And / Or my lack of understanding runs deep here.
I've rewritten with `l_mulu_64_32x32` taking just the most significant 32 bits. The `l0_mulu_32_32x32` doesn't keep the exact part of the multiplication that is needed. I'll optimise a high bytes version, call it `l_muluh_32_32x32`, which will be more efficient. But first I need to get the division working.
I'm sure I've got the algorithm right now, and the data is appearing in the right places. But, there is no convergence. I guess I'm not handling the subtraction process correctly, by having it misaligned.
Always, a learning experience.

feilipu:
Yes. Finally aligned the stars (and more importantly the binary points) and `_fsinv` is working correctly.
Now to go back and write an unsigned multiply that only does the high half bytes.

feilipu:
Just committed some clean ups and preparation to write the `l_mulu_32h_32x32` routines. But they've not appeared in this PR. I think something is broken at Github.

Confirmed with tests that although, in the `mulu_64_32x32` function the carry across from the 32 low bits to 32 high bits does change the final calculated mantissa, it only affects the lowest byte which is discarded. Therefore, it should be possible to calculate the high byte multiply entirely disregarding the lower terms. Of course this is then a broken multiply function, and can't be used generally. It will be more efficient.

suborb:
> Confirmed with tests that although, in the `mulu_64_32x32` function the carry across from the 32 low bits to 32 high bits does change the final calculated mantissa, it only affects the lowest byte which is discarded. Therefore, it should be possible to calculate the high byte multiply entirely disregarding the lower terms. Of course this is then a broken multiply function, and can't be used generally. But, it will be more efficient.

Doesn't that end up rippling due to rounding?

feilipu:
> Doesn't that end up rippling due to rounding?

Yes. You're right. I just did the implementation of the `mulu_32h_32x32` for z180 and z80-zxn and I DID need to [calculate the 3rd byte](https://github.com/z88dk/z88dk/blob/5bb2666ace2ec03f00125c34410768e0ee509cab/libsrc/_DEVELOPMENT/math/float/mathd32/z80/core/m32_z180_mulu_32h_32x32.asm#L83), and then discard all but the carry byte into the 4th byte. Actually, the effect was very small, but sufficient to cause some mantissas to be rounded incorrectly.

As it is now, by calculating 3rd through to 7th bytes, there is only maximally a small error in the final byte of the mantissa, which is discarded anyway.

Also checking the highest bit of the 3rd byte when shifting to readjust the binary point is not required.

I've not revised the z80 `l_mulu_64_32x32', because I don't really understand the shift and add multiply yet. Perhaps it can stay on the "to do" list? Or, perhaps it is good enough?

feilipu:
Starting to connect some stuff. This will probably break Travis.

@suborb would be interested to hear if there is a better way to do the sccz80 connection, than `pop` `push` stack for LHS value to preserve the final return address.

suborb:
@feilipu - take a look here: https://github.com/z88dk/z88dk/tree/master/libsrc/math/mbf32/c/sccz80 

That's the 32 bit Microsoft single precision maths interface. It's slightly different from the normal double interface since we can pass by register.

feilipu:
@suborb from your code, it seems there's no way around the `pop` `push` process, if we have to convert from `bcde` return to `dehl` returns on exit from the library.

``` asm
    pop hl		;Return address
    pop af		;Caller return address
    pop de		;Left LSW
    pop bc		;Left MSW
    push bc	;Left MSW for clean up
    push de	;Left LSW for clean up
    push af
    push bc	;Left MSW for d32
    push de	;Left LSW for d32
    jp (hl)		;Our return
```
I guess another alternative would be to put the caller return address above the left value for the library call, and then set up the return by replacing the caller return address with the right number of bytes to discard (or make the function `__callee`).

feilipu:
And so, it is done. :weary:
The four basic functions are done and imho working properly. It would be good to get some wider testing done.

Linkage to both C compilers is available. But, the functions in the library are not connected into the compilers with the connection defines in the `~/math/float/d32/ldm` directories, as this depends on how the linker will call the library. To be decided but my suggestion is `-lmd32`.

Looking at the further functions in the digi library [math.c](https://github.com/z88dk/z88dk/blob/d99c77704a259959d5e0e3ab28dd2fd0ac4548af/libsrc/_DEVELOPMENT/math/float/mathd32/c/math.c) and [digifloatmath.c](https://github.com/z88dk/z88dk/blob/d99c77704a259959d5e0e3ab28dd2fd0ac4548af/libsrc/_DEVELOPMENT/math/float/mathd32/c/digifloatmath.c) files, it is a mess of C and assembly. Will be painful to sort out, and may not be large gain over using the remainer of the SDCC C math library now that the intrinsic functions are in assembly. Not looking forward to that job.

feilipu:
@suborb I've taken the "DO NOT MERGE" flag off the title. This PR can be merged, and it won't affect the build. It will make testing across the various platforms easier. Seeking comment on merging.

libsrc/_DEVELOPMENT/math/float/mathd32/c/sccz80/cd32_sccz80_dload.asm
```diff
@@ -0,0 +1,31 @@
...
+    ; enter : DEHL = sccz80_float right
+    ;
+    ; enter : stack = sccz80_float left, ret, d32_float left
+    ; exit  : DEHL = sccz80_float right
+    ; 
+    ; uses  : af, bc, de, hl
+
+    pop hl                  ; ret from here
+    pop af                  ; ret
+    pop bc                  ; sccz80_float left
```
suborb: Aren't you trashing float right here?

libsrc/_DEVELOPMENT/math/float/mathd32/c/sccz80/cd32_sccz80_dsadd.asm
```diff
@@ -0,0 +1,35 @@
...
+PUBLIC cd32_sccz80_dsadd
+
+EXTERN cd32_sccz80_dread, cd32_sccz80_dload, md32_fsadd, cd32_sccz80_dret
+
+.cd32_sccz80_dsadd
+
+    ; add two sccz80 floats
+    ;
+    ; enter : stack = sccz80_float left, ret
+    ;          BCDE = sccz80_float right
```
suborb: Enters with dehl = float right?

libsrc/_DEVELOPMENT/math/float/mathd32/c/sccz80/cd32_sccz80_dsneg.asm
```diff
@@ -0,0 +1,29 @@
...
+; float __fsneg (float number)
+
+SECTION code_clib
+SECTION code_math
+
+PUBLIC cd32_sccz80_fsneg
+
+EXTERN cd32_sccz80_dread, md32_fsneg, cd32_sccz80_dret
+
+.cd32_sccz80_fsneg
```
suborb: I'd be very tempted to just xor the appropriate bit in the raw registers here. 

I've inlined this operation in sccz80 and I think sdcc does it too.

libsrc/_DEVELOPMENT/math/float/mathd32/c/sccz80/cd32_sccz80_dload.asm
feilipu: This is called with the alt registers `exx`, so there's nothing in registers of value.
I couldn't see a neat way to shuffle the left float and returns, without using the index registers, unless it was called with the alternate registers already switched in. This preserves the right float in the primary registers.

libsrc/_DEVELOPMENT/math/float/mathd32/c/sccz80/cd32_sccz80_dsadd.asm
feilipu: Yes. Will fix comment.

libsrc/_DEVELOPMENT/math/float/mathd32/c/sccz80/cd32_sccz80_dsneg.asm
feilipu: Yes. It would be faster and neater.

But, until I get everything working, I'd prefer to keep linkages and work separate. That way I only have to fix broken things once.

feilipu: Actually, just to comment further. It is not a stretch to simply change the entry and return values for the intrinsic functions to match `DEHL`. And, I think that is the right thing to do. It will save needing to shuffle the stack because there will be no need to "post process" the library results with its complicating return address. The linkage can be done with jumps rather than calls.

But, let me get the rest of the library working, and come back for this change on the todo list.

libsrc/_DEVELOPMENT/math/float/mathd32/c/sccz80/cd32_sccz80_dload.asm
suborb: oking at the further functions in the digi library [math.c](https://github.com/z88dk/z88dk/blob/d99c77704a259959d5e0e3ab28dd2fd0ac4548af/libsrc/_DEVELOPMENT/math/float/mathd32/c/math.c) and [digifloatmath.c](https://github.com/z88dk/z88dk/blob/d99c77704a259959d5e0e3ab28dd2fd0ac4548af/libsrc/_DEVELOPMENT/math/float/mathd32/c/digifloatmath.c) files, it is a mess of C and assembly. Will be painful to sort out, and may not be large gain over using the remainer of the SDCC C math library now that the intrinsic functions are in assembly. Not looking forward to that job.

As long as the low level routines (sin, log, exp, sqrt, pow, atn) are in assembler then the C wrappers for the rest (cos, tan, asin etc) end up pretty small and not too different from what you'd write manually - mucking around with M_PI and M_E

feilipu: OK. I'll do those too.
SQRT will be easy. Just a N-R searching for 1/sqrt(x). So it is a divide with a different seed and optimisation target.
The others I'll need to look into how they're done, and understand whether original thought is required.

suborb: Sorry my comment ended up in the wrong place - was meant to be top level.

Anyway, feeling fancy? https://en.wikipedia.org/wiki/Fast_inverse_square_root

feilipu:
> Sorry my comment ended up in the wrong place - was meant to be top level.
> Anyway, feeling fancy? https://en.wikipedia.org/wiki/Fast_inverse_square_root

Looks fancy. I'll try to get the `square root of doom` in the solution.

Standing back for a day, I'm now convinced that I should now rewrite my versions of the digi functions to use the sccz80 calling convention (lhs stack, rhs dehl). This will avoid a lot of stack bouncing overhead for both sccz80 and sdcc for everything in the future, and make the linkages (at least for sccz80) simple defines.

Then, do the other basic functions (`sin`, `log`, `exp`, `pow`, `atan`) using this more useful calling convention.

feilipu:
Chasing a problem with (at least) one of the `_fsadd` code paths, and I think the problem is related to an undocumented feature in the rabbit instruction set.

This code gives a hint, but there's no corroborating information that I can find, anywhere, supporting the supposed effect the comment refers to.
```asm
    rl a                    ; 38 different shift  rr a not rra, sets novf flag if upper 4 bits zero
    jp novf,.S8L4more       ; 45 if total is 7
```
which I've translated as...

```asm
    rl a                    ; different shift rr a not rra, sets novf flag if upper 4 bits zero
    jp PO,S8L4more          ; if total is 7
```
There are a few of these `jp novf,LABEL` instructions, and I think they may all be relying on the same undocumented effect.

feilipu:
It is hard to grind away at this `_fsinvsqrt` problem in asm, when a few lines of C work first time, and potentially call the same multiplies I've just written.
```C
    float InvSqrt(float x)
    {
        float xhalf;
        uint32_t i;

        xhalf = 0.5f*x;
        i = *(uint32_t*)&x;
        i = 0x5f375a86 - (i>>1);
        x = *(float*)&i;
        x = x*(1.5f-xhalf*x*x);
        x = x*(1.5f-xhalf*x*x);
        return x;
    }
```
But, I guess if it was easy, anyone could do it. :weary:

feilipu:
Yeesss. My precious...

libsrc/_DEVELOPMENT/math/float/mathd32/c/sccz80/cd32_sccz80_dload.asm
feilipu: Fancy fast inverse square root, done and tested.

feilipu:
@suborb @aralbrec to be honest, I'm about done with this library.

In my opinion, the interesting piece has been done, with optimised intrinsic functions customised for z80, z180 and z80-zxn.

Since the assembly functions in the PR are compatible with the SDCC 32 bit library by scraping the RHS from the stack into DEHL, and also is compatible with SCCZ80 in 32 bit float mode, I'm of a mind to just copy the SDCC floating point code in C into this PR, and link it all up.

I could use the Hi-Tech C library functions, but they're reportedly inaccurate, and I don't have the patience to write and debug trig and power functions in assembly for the next 6 months.

Thoughts, please?

suborb:
In general I don't have a problem with using C to generate the higher level functions.

However, beyond the intrinsics there's usually a few base functions that use a series generator to calculate, so if we can get log, pow, sin, atan in assembler (in that order) then most of the others are just dividing the result/adding a constant before calling.

We could of course, just get some C routines in there, and then migrate them later.

The gotcha is that newlib doesn't have any C in it at the moment, so the build/compile infrastructure isn't set up.

feilipu:
> However, beyond the intrinsics there's usually a few base functions that use a series generator to calculate.

In Hi-Tech C all of the base functions use [Horner's Method](https://en.wikipedia.org/wiki/Horner%27s_method) for calculation. These all rely on a `_evalpoly()` function.
> So if we can get log, pow, sin, atan in assembler (in that order) then most of the others are just dividing the result/adding a constant before calling.
I'm optimising the `_evalpoly()` function in assembly in the "floating point domain", and that will provide all of above, and then the other functions are, you point out simple conversions.

Once I'm happy with the functions relying on `_evalpoly()` and then following functions too from Hi-Tech C, I'll link it all together, and merge the PR.

> The gotcha is that newlib doesn't have any C in it at the moment, so the build/compile infrastructure isn't set up.

But then I don't have the necessary knowledge to link in the library and fix the build/compile infrastructure. Sorry.

aralbrec:
@feilipu Sorry I've been very busy the past few months, having little spare time and then became involved in doing some hw on the next which has some difficult deadlines.

In the original sccz80 float calling convention there is a concept of primary and secondary floating point register.  I don't know what dom's done with the 32-bit so I don't know if it still applies for that.  But this secondary register was implemented in the exx set in math48 which means sequences of float operations where made nearly stackless with the help of callee/fastcall.  If you look at the compiled code of a float calculation you will see it's fairly efficient with memory operations which contributed to making it faster than genmath.  Due to the 32-bit size of zsdcc floats and fastcall, there is efficiency in the zsdcc compiled code as well but not as much because zsdcc does not have a concept of primary and secondary float registers.  However zsdcc does get slightly faster code due to the 32-bit size on the compiler side but at the expense of accuracy compared to sccz80's 48-bit size.

I found while doing the math48 functions that it was fairly easy to rewrite the c code for some higher level functions in asm.  You may be able to borrow some from math48 but if not, you can leave them in c and maybe I can take a pass at them.  It is worthwhile to get core functions in asm as it does make a difference to code size which many targets are sensitive to.

I also saw mention some time back that you dropped a register preservation requirement on one of the integer multiply or divide functions?  The preservation requirement is usually there for a reason, either some library function(s) is optimized for it (and presumably user functions would be too) or it's needed by the fast integer library.  Keep in mind that if your float functions are using the integer math functions, they can be sped up again by selecting the fast integer math option in the target's configuration file.  So if speed is important that's what should be used rather than dropping a preservation requirement in the small version of integer math.

feilipu:
> Sorry I've been very busy the past few months, having little spare time and then became involved in doing some hw on the next which has some difficult deadlines.

@aralbrec no problem. I think many of us are looking forward to getting our Next too. Having a floating point library that uses the z80-zxn `mul de` function was one of the main inspirations for me to put the work into this PR.

I've adapted this library to match that of sccz80. LHS stack, RHS DEHL. Where there are multiple variables to be passed, I'm expecting to use the `-stdc` flag with sccz80 for compiling the library.
> I found while doing the math48 functions that it was fairly easy to rewrite the c code for some higher level functions in asm.

All the important intrinsic functions are now in assembly. I've found it a worthwhile simplification to work in the `float` domain, and just use mechanical byte shifting around the already debugged intrinsic functions.

> It is worthwhile to get core functions in asm as it does make a difference to code size which many targets are sensitive to.

Done. There are one or two utility functions that I might still do in assembly, and I think computing a `_fssqr(x)` can be written to be substantially faster than `x * x`.

> I also saw mention some time back that you dropped a register preservation requirement on one of the integer multiply or divide functions?

No, I reverted to the standard preservation option. There was a need to use the `af'` set, so now the z180 and z80-zxn have the same preservation as the z80 (in either fast or small implementation).

This library uses the z80 `l_mulu_64_32x32` function as the basis of a high double word `mulu_32h_32x32` function, only for the z80. This could be optimised, but I didn't do it for the z80.

But both z180 and z80-zxn use a special high double word multiply with rounding.

I'm going to start adding a mixture of Hi-Tech C and SDCC files into the library now that I've reorganised it slightly. Hope to get this done relatively quickly, so that testing and benchmarking can happen.

feilipu:
Added linkages for the assembly files, in `m32_asm.lst`, in as similar manner as is done for `m_asm.lst` for d48.

I've also added a `m32_sdcc` and `m32_sdcc` linkage for the C files, though as they should be compiled differently by each compiler with relevant setting, there is no differentiation in the source files.

Not sure how to handle SCCZ80 and the need for the `-fp-ieee` and `-stdc` switches, to match the current C interface. The `-stdc` switch only really affects those functions with 3 parameters, i.e. `polyf()`.

Somewhere there will need to be a `float.h` and `math.h` file created, probably by m4 (and not where they are now in `math/float/math32/c`), and the current linkages for the C files are probably rubbish.

But, without knowing how to best build C files into the new `-lm32` library, I'm at a loss of what to do next. In the `Makefile` there are only references to `z80asm` builds.

```
	$(Z80ASM) $(cpu) -xlib/sccz80/$(target) @target/$(target)/library/$(target)_sccz80.lst
	$(Z80ASM) $(cpu) -xlib/sdcc_ix/$(target) -D__SDCC -D__SDCC_IX @target/$(target)/library/$(target)_sdcc_ix.lst
	$(Z80ASM) $(cpu) --IXIY -xlib/sdcc_iy/$(target) -D__SDCC -D__SDCC_IY @target/$(target)/library/$(target)_sdcc_iy.lst
```

feilipu:
@aralbrec I'm going to hold off on building more library linkage, because I'm not sure that it is good use of my time currently. Once you get build support for C in the library, and there's a clear path to follow, I can continue.

I'm going to loop back to optimise the `_mulf()` function, as I just realised I have dead multiplies to purge. I'll switch my code to `mulu_16_8x8` multiplies, which are the natural size for z180, z80-zxn, and z80.
Then I can remove 4 dead multiplies (from the current 12), and associated byte shuffling, which impact ticks significantly. And I'll build an optimised `_sqr()` function at the same time, because the libraries need to take the square quite often.

feilipu:
Hmm. New `_fsmul` works fine for Z180, but I seem to have written a bodged up `mulu_de` for Z80.
To be repaired.

aralbrec:
@feilipu 

For C code, likely the way to go is to use the C compilers to translate the C code to asm.  And then the target libraries still use asm files in their library lists.  The C code translation would then be done ahead of the target library build in the build script.

Getting the right options on the compile line for the compilers (eg -fp-ieee for sccz80) and maybe getting some defines in to change the math.h header for a specific float library is probably something that has to be done in zcc.  I think a way to go is to look for "-lm32" and then add these sorts of options inside zcc as it parses the command line.

feilipu:
Migrated all z80 multiplications, over all multiplication sizes, to use the z80-zxn equivalent `z80_mulu_de` to maximise code reuse.

The `z80_mulu_de` has zero argument detection, leading zero detection, and is unrolled. With the exception of preserving `hl`, for equivalency with z80-zxn, it should be the fastest `16_8x8` multiply possible on a z80.

To calculate the 24 bit mantissa a special `mulu_32h_24x24` function has been built using 9 multiplies, the minimum number of `16_8x8` multiply terms. It is much more natural for the z80 to work in `16_8x8` multiplies than the Rabbit's `32_16x16` multiply, and it saves implementing dead multiplies in a larger function. It is a "correct" multiply, in that all terms are calculated and carry forward is considered. The lower 16 bits of the result are simply truncated, leaving a further 8 bits for mantissa rounding within the calling function. The resulting `mulu_32h_24x24` should be the fastest way to calculate an IEEE sized mantissa on a z80, z180, & z80-zxn.

Migrated all squaring (found in square root, trigonometric functions) to use the `sqrf()` or the equivalent assembly version. The `sqrf()` function has 6 multiplies in its `32h_24x24` mantissa calculation, instead of 9 multiplies in the normal `_fsmul` function, and also avoids the need to use the alternate register set.

Next up is `hypotf()`.

feilipu:
:checkered_flag: :checkered_flag: :checkered_flag: :checkered_flag: :checkered_flag: :checkered_flag: :checkered_flag: :checkered_flag:

libsrc/_DEVELOPMENT/math/float/math32/z80/f32_z80_mulu_de.asm
```diff
@@ -0,0 +1,117 @@
...
+   sla h
+   jr C,branch_15
+
+   sla h
+   jr C,branch_16
+
+   sla h
+   jr C,branch_17
+
+   sla h
```
ped7g: the result is already in DE, you can jump straight to `pop hl` instead, and skip both `sla h` and `ex de,hl` (for D=1 input) (also CF=0 btw)

libsrc/_DEVELOPMENT/math/float/math32/z80/f32_z80_mulu_de.asm
```diff
@@ -0,0 +1,117 @@
...
+
+.branch_17
+   add hl,hl
+   jr NC,exit
+   add hl,de
+
+.exit
+   ex de,hl
+   pop hl
+   or a
```
ped7g: why `or a`? if for carry flag, CF is already zero for all possible code paths here (after removing that last `sla h` above for D=1 case) (because either last `add hl,hl` was NC, or the following `add hl,de` did clear it)

Also if CF=0 is even important, then the `.zero` tail is not resetting it, so if any input is zero and CF=1, the routine will return DE=0, CF=1. But I think CF is not important at all?

libsrc/_DEVELOPMENT/math/float/math32/z80/f32_z80_mulu_de.asm
```diff
@@ -0,0 +1,117 @@
...
+   ret
+
+.zero
```
ped7g: to skim few more cycles, this can be broken into two tails, `zeroD` and `zeroE` (depending which one was zero), each of them doing either `ld e,d` or `ld d,e` + `ret`. ... now I even actually wonder how often you multiply by zero, if the initial test is worth it, as the `sla h` for leading zero bits takes roughly about 8 x (8+7) = 120T extra for zero input, but for non zero input the initial test is 30T, i.e. for 65025 out of 65536 cases there's +30T extra, remaining 511 cases save about 120T => i.e. roughly1950750T lost (non zero input) vs 61320T gained (when zero input). ... seems like I missed something obvious, and my reasoning is wrong somewhere, or you have hard-data from real life suggesting mul by zero is used a lot (but if then also mul-by-1 is a common thing, you can make that super fast by changing entry to `dec d / ret z / inc d / jr Z,zero`).

libsrc/_DEVELOPMENT/math/float/math32/z80/f32_z80_mulu_de.asm
feilipu: @ped7g thanks for taking the time to review. I appreciate it, and the feedback.

I was resetting carry, simply to align with the rest of the library code. I don't examine the carry at all. My thinking was muddy about whether this `mulu_de` was worth putting in the integer library, or not. Now, I believe it doesn't belong anywhere else. It is best to make it as fast as possible, since it is used exclusively by this library. It is the hottest code, so to speak. So, I'll remove `or a`.

Yes, as you say, neither the z80-zxn or z180 affect carry with their multiply instructions. You'll note that in the `z80_mulu_32_16x16` I have to capture the carry to preserve it past the last `mulu_de` for later use. This is not required in the z180 or z80-zxn versions of the code.

libsrc/_DEVELOPMENT/math/float/math32/z80/f32_z80_mulu_de.asm
feilipu: Yes, I agree. Actually I woke up thinking the same thing this morning, about splitting the tails.
Done.

But, more generally about whether zero detection is worth it or not? I've no idea, actually.

My thinking was coloured by two things. First, when initially using the rabbit style logic, there were a lot of dead multiplies, where one byte of the 16 bit pair would be set to zero. Implementing this on the z80 (z180, z80-zxn) meant a lot of needless multiplication. But, I've resolved this now, and it won't be an issue.

Second, a lot of integer like numbers have "short" mantissas. If we call a 24 bit mantissa made up of three bytes "abc", then quite often the b and c bytes end up being zero. I think this will be a common case. Handling these with zero detection will be a big win, making many multiplies very fast (because the result is zero). Of course the other side of the argument is that integers should be handled by the integer library, and that calculating a 10 iteration polynomial estimation for `logf()` (for example) will never have zeros in it, so carrying zero detection overhead for short mantissa floats is just wasteful.

I guess the only way to find out is to benchmark this. Another box to tick.

Just to comment, I don't see any need to specially detect a 1 byte. That will come forward no more often than any other value.

libsrc/_DEVELOPMENT/math/float/math32/z80/f32_z80_mulu_de.asm
feilipu: Thanks good input.
Done.

suborb:
@feilipu Congratulations, really good work!

Let me know when you're ok for me to merge and I'll get this into classic.

Benchmarking this in classic is probably easier since there's a natural ticks target and the cross platform maths libraries (genmath, math48, mbf32) can be easily switched in.

feilipu:
@suborb I think I am done. Have at it.

Note:
- [ ] sccz80 linkages hasn't been tested at all. You might need emergency surgery.
- [ ] z80-zxn has only been tested on a z180.
- [ ] not sure whether `-stdc` linkage will work for sccz80.

Very keen to see what happens on a z80 platform with multiply function, and how far behind this `mulu_de` mechanism is behind traditional shift and add for `32_32x32`. I'm hoping it is a close race, because the byte shifting overhead may be similar for both mechanisms, but who knows.

PR #1146, opened 2019-05-10
(math32) Start on the lowlevel sccz80 interface

suborb:
-Basic arithmetic and the .asm implemented functions are now working in classic/sccz80

Tasks (classic/common):

- [x] Change exports of c files to be m32_cosh 
- [x] Precompile the .c files to generate m32_cosh.asm etc and commit
- [x] sccz80 support functions (l_f32_*)
- [x] Complete sdcc support functions
- [x] Figure out how to handle __stdc functions for sccz80 wrt header files
- [x] Generate cm32_sdcc stubs
- [x] Generate  cm32_sccz80 stubs
- [x] Convert sdcc integer<->float conversions to .asm and take callee convention/names (our optimiser requires callee)
- [x] Update classic float.h constants to match IEEE values
- [x] Create z80 --IXIY library (if .C files compiled with sdcc)

Tasks (newlib):

- [ ] Update classic float.h constants to match IEEE values
- [ ] Implement support routines required by dtoa() (asm_fpclassify) 
- [ ] Update .lst files and include in newlib libraries.
- [ ] Generate newlib -lm32 library using stubs in math/float/math32/lm32
- ~[ ] Figure out how multiple maths libraries can be used in the same program~


suborb:
@feilipu Testing on a z80, it looks like we have few places of precision than I expected? print_bin() just dumps out the float in hex:

```
        float a = 20.0;
        float b = 10.0;
        float c = 1.0;

        print_bin(&c);
        print_bin(&b);
        print_bin(&a);
        c = a / a;
        print_bin(&c);
        print_bin(&b);
        print_bin(&a);
```
Gives:

```
feac 00 00 80 3f
feb0 00 00 20 41
feb4 00 00 a0 41

feac ff fd 7f 3f
feb0 00 00 20 41
feb4 00 00 a0 41
```

So the result should be 1.0 (0x3f800000) but ends up being (0x3f7ffdff) which is 0.999969, like wise for a/b gives us 0x3ffffdff which is 1.999939

feilipu:
@suborb I'll add these comments on the accuracy to the readme.md file, but it goes like this...

I'd expect to see at least the same answer as in the m48 code. I used that as my test tool, and did the same calculations using m48 and m32. I don't remember m32 getting different answers for divide (unlike the derived calculations, which well, see comments below).

BTW. I fixed a [carry bug](https://github.com/z88dk/z88dk/commit/9150124ffbb659a288634bfb1af0f143f4abf022) in the `_sqrf()` mantissa calculation last night. But this doesn't affect multiply / divide code.

For readme.md

The multiplication process should be "correct" and in fact more correct than the digi international method. Every carry term is calculated and brought into the result. Digi international code just ignored the c*f term of the 24-bit calculation of abc*cdf, because they determined it wasn't needed. m32 doesn't take that short cut. This also applies to the squaring process, which should also be "correct".

The division process relies on N-R estimation, and it follows exactly the same process as digi international do. There are some notes about the number of significant bits of calculation required to derive a correct IEEE 24-bit mantissa, and I believe that using the `32h_32x32` calculations this is achieved. I was seeing outcomes close to, and mostly the same as the m48 package. But both of these are not rounding to whole numbers (ever). 2.0 is always seen as 1.99999xx, and never as 2.00001xx. I've no clear idea why this is the case.

The addition / subtraction process should be "correct", and unless there are unresolved issues this result should be identical to m48 within the significant digits of IEEE 754.

The square root calculation relies on N-R and is an estimate only. With 2 iterations I think the estimate is pretty usable. 1 iteration is good for games and posting light sources, nothing else. It is pretty easy (at the expense of ticks) to make 3 iterations, which should be pretty damn good (as a technical outcome).

The rest of the derived functions rely on the polynomial expansion process and will only be as accurate as the coefficients that are fed into the process. I've just used those in the Hi-Tech C library code. I guess they got them right, but their code is not known for accuracy. Someone with a mathematical background might be interested to calculate better coefficients at some stage.

feilipu:
@suborb I might have a shot at the variant Newton-Raphson process using the Chebyshev polynomial initialisation. Now that I can work in the compact floating point domain, and have a fast `_polyf()` function available, this might be an easy way to get better accuracy, without an additional N-R iteration.

suborb:
I'm carrying on adding the sccz80 glue and converting the low-level routines to assembler and get things into a state where we can compare the performance of the various libraries.

I've just tried the same test with math48 and I'm getting the results I expect:  x / x = 1.0 (exactly) and 5 = sqrt(25.0), with math32 it's 0.999... and 4

I think not having these right is going to cause a lot of confusion/bugs. I'm guessing the issue is an artefact of the division algorithm - if there are no bugs in the implementation, then I feel that we should try out the more traditional algorithm.

feilipu:
Thanks, I think benchmarking this is the only way to see if it is actually a useful contribution.

I’ll look at the following things
digi international code review - did I miss something
`32h_32x32` to make it calculate and carry all the terms
N-R with Chebyshev initialisation

Away a few days this week, so I won’t be able to try all, but at least some.

feilipu:
Added a [third iteration to the `sqrtf()` function](https://github.com/z88dk/z88dk/pull/1148).

Now produces "more correct" outputs.
Passes the sqrt(25) = 5 test, exactly.
But starts to loose last significant digit correctness above sqrt(100).

feilipu:
Standing back from `_m32` for a bit, I'm tempted to say that it would be good to do it all again (and end up at the same place pretty much). They say the best code is written at least twice.

I think for the more complex functions, I should define the intrinsic arithmetic functions in terms of an internal "expanded IEEE" format that uses 6 bytes and a 32-bit mantissa (BCDEHL == SEMMMM), and use this internal number format to preserve mantissa accuracy but also make any higher functions written in assembly regular and easier to maintain.

What?  Well the inverse function is written obscurely, and is not orthogonal. Each iteration uses different techniques. But it does keep the 32-bit mantissa to the final rounding. OTOH, the inverse square root function is clean, and easy to read and maintain, but preserves only 24-bit mantissa between intrinsic function calls. I want both outcomes. Simplicity and 32-bit mantissa accuracy.

Also, as a later stage I want to have the complex functions written in assembly, and they need to have more accuracy than 24-bit mantissa allows. At least initially the `polyf()` function should be written in "expanded IEEE" format, even if its coefficients are stored and passed in IEEE standard format.

This "expanded IEEE"  format won't have better range than a normal IEEE float, and it shouldn't be exposed outside the library, but it will make internal calculations more accurate, and I think that is needed.

The overhead for 32-bit mantissa won't be too great, I feel. For `_fsmul`, I calculate 10 multiplies for 32-bits, vs 8 multiplies for 24-bits. Not outrageous. This 25% overhead impact will scale linearly across all of the library.

Just leaving these thoughts for comment...

suborb:
It's been painful, but we're now in the state where we can call sin() from and sccz80 and the sdcc compiled sin() function runs and produces a number vaguely in the right ballpark: sin(2)=0.902713, should be 0.909297

A snag with using the .c files is that two maths libraries can't co-exist at the same time (this was only ever possible with newlib) since those C files need ___fsdiv_callee etc

feilipu:
Well done. There’s so much that has to line up, just to get that far!

Now, the whole accuracy vs speed trade-off can be evaluated.

Fixing the divide routine will help, and is the first step on my side.
Then, getting more accuracy from the `poly()` function using a fuzed multiply-add will help too (a little).
And finally, understanding whether the coefficients being fed into the functions are accurate and enough terms are being used to generate accurate results is another thing to do.

suborb:
This should now be complete for classic. The last m32 rounding PR sorted out the ftoa() problems very nicely.

I'm not going to tackle newlib in this one - I'll open a new issue and then bits of it can be tackled separately.

PR #1147, opened 2019-05-11
zsdcc - update to r11214 - sdcc 3.9.0 stable release

feilipu:
This PR is to move the zsdcc version to [sdcc tag r11214, which is the sdcc release 3.9.0](https://sourceforge.net/p/sdcc/code/11214/)

The 3.9.0 release is a stable release, and contains only bug fixes relative to the current r10892 tagged zsdcc.


suborb:
I don't suppose you know why this PR contains already merged commits? It looks like it's GitHub being a bit daft given the actual diff.

feilipu:
[My git-fu is about 2/10, so everything below is just a guess]

When I merged the m32 PR, I squashed it down to a single commit so the history was just one entry.

Before I did this PR I “synced” my repo to the master ( the PR39 mentioned), then made the changes.
So I guess that git is thinking that all these changes still need to go into the master, although it knows the code is already there.

Options?
1. close the pr, trash my repo, and do a new pr from a clean repo.
2. close the pr, “something about branching”, do a new pr.
3. merge this pr, but edit the history during merge, so that the unnecessary junk commit info doesn’t appear. This would be the easiest, but I don’t know if it it is possible.

suborb:
Ah. You squashed. That explains it since history has been rewritten. 

We’ll keep having this weirdness then and have to squash everything coming from your repo. 

Might be best to regenerate your master branch and then raise prs from a feature branch. That way they can be squashed if desired and your master stays in sync.

feilipu:
Ok will do.

Issue #1118, opened 2019-03-03
ZSDCC - bug for segment assignment


feilipu:
Alan @EtchedPixels has found a bug in sdcc https://sourceforge.net/p/sdcc/bugs/2837/ that might affect ZSDCC 3.8.5.

Opening this issue to track resolution and path toward 3.9.0.

feilipu:
Partially fixed by [r11146](https://sourceforge.net/p/sdcc/code/11146/).
sdcc 3.9.0 r11214 contains the partial fix, as per #1149.
Doesn't look like anyone is rushing to do a complete fix.

feilipu:
Partially fixed, and not a high priority.
Close to clean up.

PR #1149, opened 2019-05-12
zsdcc - update to r11311 sdcc 3.9.1

feilipu:
This PR **was** to move the zsdcc version to [sdcc r11214](https://sourceforge.net/p/sdcc/code/11214/), which is the sdcc release 3.9.0.

The 3.9.0 release is a stable release, and contains only bug fixes relative to the current r10892 tagged zsdcc.

EDIT. This PR is to move the zsdcc version to [**sdcc r11311**](https://sourceforge.net/p/sdcc/code/11311/), which is within the sdcc release 3.9.1. A bug was identified and repaired in v3.9.1 relating to `__z88dk_fastcall`. Other improvements to the register allocator `ralloc2.cc` were made for the `iy` and `a` registers.

The only substantive changes to the patch file are where `SDCCglue.c` now contains the zsdcc patch in the main line, so included patch code needs to be removed from this patch. This patch code was referenced as a partial fix for #1118.


feilipu:
Filed an issue on SDCC, because of incorrect code generation.
https://sourceforge.net/p/sdcc/bugs/2912/

suborb:
@feilipu are you sure that's not our optimisation rules causing problems?

feilipu:
@suborb. No I built with clean SDCC to check the effect. I saw this also with the recommended version in our repo now too. So it’s been broken for some time. r9855 is ok of course.

suborb:
Does that suggest we should roll back to r9855 then?

feilipu:
I'm conflicted as to whether to roll back. But, I don't see a strong reason to right now.

This breakage is very "binary", not subtle. If the resulting assembly is broken, then it is no surprise and things stop right there. So if it were a common occurrence we'd have heard. So, rolling back now is probably not warranted.

But still, @spth, this still needs to be fixed before moving forward.

What is very surprising for me is that `m32_tanf.c` is just one, almost trivial, line of code.

```c
#include "m32_math.h"
float m32_tanf (float x) __z88dk_fastcall
{
    return m32_sinf(x)/m32_cosf(x);
}
```

suborb:
That's a good point, it is an obvious failure.

It does look like it's related to the fastcall - it also occurs with longs. It's not going to be a commonly encountered issue either - it seems to require both `m32_tanf()` and `m32_sinf()` to be fastcall.

Something in the register allocator seems to have changed as well - there seems to be preference to use bc, there's a fair bit of unnecessary swapping of bc and de going on.

feilipu:
Yes.
> Something in the register allocator seems to have changed as well - there seems to be preference to use bc, there's a fair bit of unnecessary swapping of bc and de going on.

I noticed that too. I've been comparing outputs for various versions for all the C in m32 library, and in no case is the new version better. It is not pretty, and I make it a regression in performance.

But, I figured that on balance probably a lot of other things in the compiler have gotten better, and as long as there is no obvious breakage then staying with the latest stable version is sensible from maintenance, and future thinking point of view.

feilipu:
@suborb, actually I was wrong. r9958 is not correct either.

This [sinf() function is definitely broken](https://github.com/z88dk/z88dk/blob/master/libsrc/_DEVELOPMENT/math/float/math32/c/asm/m32_sinf.asm#L283), because the `hl` low bytes of the variable are overwritten before being used. [atanf() also has the same problem](https://github.com/z88dk/z88dk/blob/master/libsrc/_DEVELOPMENT/math/float/math32/c/asm/m32_atanf.asm#L281).

The resulting code is better if the `--reserve-regs-iy` flag is used.
I'm just checking now whether the old ralloc function can be triggered too.

feilipu:
@suborb noting that r9958 has proven to have bugs too, perhaps this is worth a bit of testing to allow us to move over to r12281?

spth:
> I noticed that too. I've been comparing outputs for various versions for all the C in m32 library, and in no case is the new version better. It is not pretty, and I make it a regression in performance.

I've had a look at the one example function given in the bug report. Register allocation is now (rev. #11285) better for that one. Without --reserve-regs-iy, no variables are spilt.

Philipp

feilipu:
Philipp's doing more stuff for z80 in sdcc. We're now at r11288, with improvement to register handling.
Testing then move here...

feilipu:
Tracking #1207 with unsigned short values.

suborb:
I've now got win32 and macOS builds of this version which are ready to go.

As a note to future self, to remove all the win32 format warnings define: __USE_MINGW_ANSI_STDIO in src/common.h

feilipu:
I note that @spth already made a few more improvement to the `a` register handling in [r11923](https://sourceforge.net/p/sdcc/code/11293/), so I was wondering whether he's planning further improvements in the next few days or weeks?

If no further work is necessary I'd suggest to draw the line in the sand there at r11923.

So since we're already past the 3.9.0 stable version, and because of the fastcall bug in r9958, and the code generation bug in r10892, I was thinking probably moving to r11923 (or something where the release number matches the command line reported release number).

Or if sdcc is planning on a 3.9.2 in the next week or so, we could use that?

spth:
> I note that @spth already made a few more improvement to the `a` register handling in [r11293](https://sourceforge.net/p/sdcc/code/11293/), so I was wondering whether he's planning further improvements in the next few days or weeks?

I'll be quite busy during at least the next two weeks. If I find time for SDCC during that time it will probably go to the pdk13, pdk14, pdk15 and pic14 ports.

feilipu:
There is a [peep-hole optimiser bug](https://sourceforge.net/p/sdcc/bugs/2914/), fixed in [r11295](https://sourceforge.net/p/sdcc/code/11295/).
Although this doesn't affect z88dk, including it means that the "same" `sdcc` commit version can be used as is being used for `zsdcc`.

feilipu:
New sdcc bug registered [*2918](https://sourceforge.net/p/sdcc/bugs/2918/).

suborb:
I've updated the binaries so I'm going to merge this and then tomorrow we'll be on 11311.

Any idea why the version number is wrong @feilipu ?

feilipu:
I note that the `/trunk/sdcc/ChangeLog` file contains the release number which was also not updated. I don't know whether that is a cause, or an additional symptom. But they both report r11310, which is incorrect. But, it is inconsequential.
Thanks @spth for getting fixes so quickly.
🤞 that this is a good and stable platform for a while.

Issue #1155, opened 2019-05-21
(newlib) Update newlib to support math32


suborb:
The classic PR provides support for math32 in classic. However, newlib doesn't fully support math32 yet. The following is my understanding of what needs to be done:

- [x] Update .lst files to include the math32 core into each of the library builds
- [x] Add sections for code and rodata
- [x] `-lmath32` should have `asm_` bindings for all functions
- [x] Generate `math32.lib` for newlib
- [x] Enable `-lmath32_fast` linkage and add missing `math32_fast.lib` newlib
- [x] Create sccz80 and sdcc_ix variants of the `_m32_*.c` pre-assembles (not needed).
- [x] Update printf core to handle 4 byte floats for sccz80 (use CLIB_32BIT_FLOATS as per classic)
- [x] Update printf core to not convert 4 byte floats to math48 floats (use CLIB_32BIT_FLOATS as per classic)
- [x] Provide support routines needed for dtoa and hence printf() to work. The interface should be in `-lmath32`
- [ ] [???](https://github.com/z88dk/z88dk/issues/1155#issuecomment-516212490) Update scanf core to handle 4 byte floats for sccz80 (use CLIB_32BIT_FLOATS as per classic) __EDIT__ scanf for float is currently unsupported.
- [ ] [???](https://github.com/z88dk/z88dk/issues/1155#issuecomment-516212490) Update scanf core to not convert math48 floats to sdcc floats (use CLIB_32BIT_FLOATS as per classic) __EDIT__ scanf for float is currently unsupported.
- [x] Create math32 support routines for strtod - these should be in `-lmath32`
- [x] Update math.h for math32 values/functions. Decide on how to switch between them.
- [x] Check for any other float usage in library, ensure it uses asm_ interfaces and that they are present in `-lmath32`

feilipu:
Can't do everything, but will at least try to get the asm pieces working.

In #1157 
- The .lst files for all targets have some build info already, but this will be extended to all math32 asm functions.
- Add in sections specific to math32.
- Build the `-lmath32` asm bindings

.

feilipu:
BTW, I'm pretty impressed with the zsdcc generated functions in the classic lib.
I don't think it would  be possible to do a substantially better job by hand, unless one avoided usage of the index registers. But the saving in ticks would be outweighed by the increase in maintenance effort, imho.
[...]

PR #1159, opened 2019-05-22
m32 - fspoly index and rounding - do not merge

feilipu:
This PR will be replaced by a smaller one that maintains the current calling convention for `poly()` as defined in the IEEE math standard.

- [x] Reworked the polynomial evaluation function to find the highest index for the table in the first byte of the table. With the index calculation, a maximum of 64 coefficients are supported. The maximum table index used in the library is currently 9, so this won't be a limitation.
- [x] Reworked the rounding to match rest of the library.

And, you know, the output is pretty good, at least for logf and sinf. It agrees with the d48 calculations very well, except for the expf function that is very fast moving. I don't think that reworking coefficients is a priority, now that the underlying math is solid.

There may be breakage for sccz80. I've tried create the right cm32 stubs, but this needs checked.


suborb:
I think making the internal functions callee will have broken the bindings for both compilers - I removed the modifier in my PR to simplify things.

When a function is callee we have to do this trick (random function chosen):

```
extern char __LIB__ *utoa(uint16_t num,char *buf,int radix) __smallc;
extern char __LIB__ *utoa_callee(uint16_t num,char *buf,int radix) __smallc __z88dk_callee;
#define utoa(a,b,c) utoa_callee(a,b,c)
```

So that when we call directly we call the callee function, but when we take a pointer we take the address of the non-callee version. This means we need a cm32_XXXX_YYYY_callee.asm stub.

There's a similar thing that has to be done for __z88dk_fastcall as well.

However, it will be currently working since the .asm files haven't been regenerated - I left this as a manual step so there's no hard dependency on sdcc when building - just run the make in math32/c and commit the changes.

I also spotted that m32_atanf() is actually named m32_m32_atanf() which is probably due to my bad sedding.

feilipu:
Ah. OK. I didn't notice your active change of the linkage.  I just thought it was my sloppiness.
I'll unwind that out of this PR. Sorry.

Also, I'll leave it open so that it doesn't mess up any current work. It can be committed just when someone is working on the #1155.

libsrc/_DEVELOPMENT/math/float/math32/c/asm/m32_tanf.asm
```diff
@@ -274,15 +278,25 @@ _m32_tanf:
 	add	ix,sp
 	push	af
 	push	af
+	ld	a, l
+	ld	(), a
```
feilipu: Breakage.

libsrc/_DEVELOPMENT/math/float/math32/c/asm/m32_tanf.asm
```diff
@@ -274,15 +278,25 @@ _m32_tanf:
...
+	ld	a, h
+	ld	(), a
```
feilipu: Breakage.

feilipu:
Also noting that there is a significant speed up and accuracy improvement in replacing the many`1.0/x` cases in the c code by `m32_invf()`. And the `x*x` cases with `m32_sqrf()’.

A separate PR I think.

feilipu:
Closed because IEEE set  the calling convention for `poly()` and there's no reason to reinvent that wheel.

PR #1168, opened 2019-05-30
m32 - optimise 32h_24x24 mantissa calculation

feilipu:
In the process to achieve accuracy, I removed the optimisation present in the original digi international `32h_24x24` mantissa calculation, thinking it might have been the origin of the inaccuracy. Now that we're accurate we can return to look at this optimisation.

The optimisation  again... where the mantissa is 3 bytes "abc" and they are multiplied by 3 bytes "def", the lowest multiplication term "c*f" can be ignored as it doesn't influence the 32-bit result.

The optimisation produces a 1/8th improvement in the multiply and a 1/5th improvement in the square calculation, so is an easy and substantial performance gain.

The mantissa resulting from the optimised calculation is still within +-1 bit of correct (as it was previously), and tests remain passing.


suborb:
I'll regenerate the figures on this page: https://github.com/z88dk/z88dk/wiki/Classic--Maths-Libraries when this is merged.

feilipu:
Thanks. Looking forward to the results.

It will (hopefully) make a difference for the z80, taken together with the `zeroe` early exit modification earlier.

But it may not influence the z180 and z80-zxn figures too much, as the `mlt nn` and `mul de` is only a few cycles anyway. Perhaps the associated saving in byte shuffling and word additions will make it worthwhile too?

Anyway, I think this and #1166 are about where I will need to leave this work for the time being, as I'm not sure of the destination required for the newlib library #1155, so further work may end up being simply incorrect.

suborb:
Done: https://github.com/z88dk/z88dk/wiki/Classic--Maths-Libraries/_compare/06d758653e057c9aead5372a4be20cac4786b8c5...443ae823af76b7b4089dbdf34c40cbc05f01cb02

Improvements in the range 1-5%

PR #1172, opened 2019-06-02
Fixup error handling of output file creation.

QwazyWabbitWOS:
Some changes were needed in the way dzx7 handled null output_name.

### DIFF
--- a/src/zx7/dzx7.c
+++ b/src/zx7/dzx7.c
@@ -119,7 +119,7 @@ void write_byte(int value) {
 
 void write_bytes(int offset, int length) {
     int i;
-    if (offset > (int)(output_size + output_index)) {
+    if (offset > (int)(output_size+output_index)) {
         fprintf(stderr, "Error: Invalid data in input file %s\n", input_name);
         exit(1);
     }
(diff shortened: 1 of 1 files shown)


suborb:
There's some wonky whitespace changes going on....

But more fundamentally, under what circumstances/OS was malloc() failing? It's so early on in the run, that it feels unlikely to be a problem.

QwazyWabbitWOS:
Visual Studio 2019 reformats white space on edit, without warning. It's a readability thing, I suppose.

Return value of malloc should always be checked. It's just good coding practice. If a function can return failure, always test for failure before proceeding. If malloc fails, output_file is NULL and strcpy fails.
One never assumes the malloc family of calls will always succeed. One doesn't wait for some OS or conditions to induce a failure.
See -Wall -pedantic on GCC or clang.

suborb:
Ah, sorry I thought you'd experienced an actual problem and I was curious as to where/how it happened, if it's just a cleanup then ok.

If you can do the following I'll merge:

a) Fix up the white space
b) Exit at the point of allocation failure rather than at some point later

Note on a default Linux installation this error checking path won't be hit and it the program may fault when the memory is accessed should the machine be out of memory and the OOM killer fail to reap a process.

QwazyWabbitWOS:
Changes made as requested. 
If the OS kills the process on OOM then there's nothing the C language can do about it. Not checking the return value from malloc is undefined behavior and depending on the OS to save you is a programming error. But this change implements defined behavior without OS dependencies so I hope it's accepted in the spirit it was intended.

QwazyWabbitWOS:
Oh, _that_ wonky whitespace. I understand now. The editor inserted tabs. I will fix.

suborb:
Cool, thank you for the contribution.

PR #1181, opened 2019-06-06
m32 - mulu_de add config_private

feilipu:
add "config_private.inc" so that `_CLIB_OPT_IMATH` configuration switch works.


feilipu:
@suborb how does the switch work in classic?
This is clearly wrong.

feilipu:
Will have a another go at this...

PR #1184, opened 2019-06-08
(math) ldexp fixups

suborb:
Fixups for ldexp for math32, and implementations for genmath + mbf32

### DIFF
--- a/include/math/math_genmath.h
+++ b/include/math/math_genmath.h
@@ -97,7 +97,7 @@ extern double_t __LIB__ ceil(double_t) __smallc;
 
 /* Manipulation */
 extern double_t __LIB__ modf(double_t, double_t *) __smallc; /* decomposes a number into integer and fractional parts */
-#define ldexp(x,y) (pow(2.,(int)(y))*x)
+extern double_t __LIB__ ldexp(double_t x, int pw2) __smallc;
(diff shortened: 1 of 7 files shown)


feilipu:
I've been sitting looking at this now for about 90 minutes.
And I'm convinced that there's something I'm not quite understanding.

The frexp correction you've suggested, imho is not correct.
The mantissa in ieee is assumed 1.0 <= m < 1.999...
But frexp has to return mantissa 0.5 <= m < 0.999... 
so the original function loading exp `$7e` is correct.

And, but this doesn't make the Hi-Tech C originated C functions correct...

Thoughts?

suborb:
float ldexp (float x, int exp);

` ldexp` is meant to multiply x by 2^exp. 

To double x we need to multiply by 2^1. Which means just increasing the exponent by 1. Looking at the C snippet in the math32 ldexp file addition is what it does.

This fix gets the correct values for the alternate `exp()` function I sent you last night

feilipu:
Yes, granted. I was looking at a narrow view of what `ldexp` does.
So the `ldexp` fix is correct.

And, you aren't suggesting to change `frexp`, so that my understanding there is OK too.

I've got some other non - functional beautification to add, but I'll do it after this PR.

feilipu:
I've just merged these so I can push the other things too.

PR #1186, opened 2019-06-08
m32- add div2f & mul2f

feilipu:
An easy optimisation, for the relatively common case of multiplying or dividing by 2.


suborb:
I was thinking about calling ldexp from sccz80 to do this. Yet to take a look to see how easy this will be to do.

Issue #1187, opened 2019-06-08
Implicit Dependency on stdlib.h in ucpp?


QwazyWabbitWOS:
Visual Studio (versions tested: 2010, 2017, 2019) spews 'unknown pragma' warnings if <stdlib.h> is not explicitly included in the source modules assert.c, cpp.c, eval.c, lexer.c, macro.c. It looks like not including the header was an oversight as mem.c has it and nhash.c doesn't need it. I can post a PR if desired.

suborb:
That's weird, I can't find any `#pragma` within ucpp. Where is it reporting the warning?

QwazyWabbitWOS:
It's in the output stream of the VS IDE during the compilation process. The #pragmas are within Microsoft's headers and they're native to their libraries and correctly formatted. 
Sample output:

> 1>cpp.c
1>C:\Program Files (x86)\Windows Kits\10\Include\10.0.17763.0\ucrt\corecrt_wstdlib.h(362,9): warning C4068:  unknown pragma
1>C:\Program Files (x86)\Windows Kits\10\Include\10.0.17763.0\ucrt\corecrt_wstdlib.h(363,9): warning C4068:  unknown pragma
1>C:\Program Files (x86)\Windows Kits\10\Include\10.0.17763.0\ucrt\corecrt_wstdlib.h(364,9): warning C4068:  unknown pragma
1>C:\Program Files (x86)\Windows Kits\10\Include\10.0.17763.0\ucrt\corecrt_wstdlib.h(373,9): warning C4068:  unknown pragma
1>C:\Program Files (x86)\Windows Kits\10\Include\10.0.17763.0\ucrt\stdlib.h(304,9): warning C4068:  unknown pragma
1>C:\Program Files (x86)\Windows Kits\10\Include\10.0.17763.0\ucrt\stdlib.h(305,9): warning C4068:  unknown pragma
1>C:\Program Files (x86)\Windows Kits\10\Include\10.0.17763.0\ucrt\stdlib.h(339,9): warning C4068:  unknown pragma
1>C:\Program Files (x86)\Windows Kits\10\Include\10.0.17763.0\ucrt\stdlib.h(637,9): warning C4068:  unknown pragma
1>C:\Program Files (x86)\Windows Kits\10\Include\10.0.17763.0\ucrt\stdlib.h(638,9): warning C4068:  unknown pragma
[...]

PR #1188, opened 2019-06-09
PR for #1187

QwazyWabbitWOS:


QwazyWabbitWOS:
I reverted the VS project commit that immediately preceded. I wasn't sure if you wanted it. 
VS2015 is unsupported so eventually VS2019 will be required.

suborb:
Thanks, sorry for the delay in merging.

Regarding the vcproj - I don't use them, so I'll leave it for others to comment.

PR #1196, opened 2019-06-13
make - disable sdcc pdk13 port 

feilipu:


feilipu:
@zx70 I was just checking the discrepancy in Dockerfile vs Makefile regarding gameboy and mcs51.

`--disable-mcs51-port --disable-gbz80-port`

I think they should be added to the Makefile too, if they're not supported.
But, I'm not sure if they should be deleted from the Dockerfile?

PR #1197, opened 2019-06-14
m32 - use zsdcc r11276 to fix fastcall

feilipu:
There is an issue with the zsdcc r9958 version (old faithful), in that it overwrites the `hl` register parameter with `__z88dk_fastcall` functions with the stack adjustment, before `hl` can be saved. Discussed in #1149.

To be seen in `asinhf()`, `atanf()`, `acoshf()`, `sinf()`, and others.

This is fixed with the r11276 (and other earlier versions). But, somehow the code quality is otherwise worse. Perhaps there's some new peephole optimisation rules resulting.

### DIFF
--- a/libsrc/_DEVELOPMENT/math/float/math32/c/Makefile
+++ b/libsrc/_DEVELOPMENT/math/float/math32/c/Makefile
@@ -7,4 +7,4 @@ CFLAGS += -I.
 all: $(addprefix asm/,$(AFILES))
 
 asm/%.asm: %.c
-	zcc +z80 -clib=sdcc_ix --max-allocs-per-node200000 -SO3 $(CFLAGS) -a $^ -o $@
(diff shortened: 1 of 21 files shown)


suborb:
@feilipu does this resolve some of the wonky results?

suborb:
With regard to #1155 and the sdcc_ix vs sdcc_iy split, `__z88dk_fastcall` causing problems with sdcc and the slightly odd bc loads I'm thinking it might be better to use sccz80 to compile these - given that the routines are pretty "simple" and rely heavily on the intrinsics I suspect that sccz80 might do better in terms of size/speed.

Removing the use of ix simplifies the newlib build.

feilipu:
I don't think this influences the `logf()`, `expf()`, or `powf()` inaccuracy.

>  I'm thinking it might be better to use sccz80 to compile these routines.
Give it a go. Certainly zsdcc is not doing to well currently, either wrong code, or poor code.

PR #1199, opened 2019-06-15
m32 - fix polyf powf connections

feilipu:


libsrc/_DEVELOPMENT/math/float/math32/lm32/c/sccz80/atan2_callee.asm
```diff
@@ -0,0 +1,14 @@
...
+	PUBLIC	pow_callee
+	EXTERN	cm32_sccz80_atan2_callee
+
+	defc	pow_callee = cm32_sccz80_atan2_callee
+
+
+; SDCC bridge for Classic
+IF __CLASSIC
+PUBLIC _atan2_callee
+defc _atan2_callee = pow_callee
```
suborb: @feilipu are you sure?

feilipu: Ouch. copy->bad_paste.

suborb:
As an aside, I'm not convinced that log() is actually working.

I'm trying to track it down, but basically log(0.5) just hangs when called from sccz80, and from sdcc the result is always 0.1

feilipu:
> As an aside, I'm not convinced that log() is actually working.

I don't think it is `logf()`. Rather, it is `pow()`, that is not working for me. I've been sitting on it now for 5 hours, playing with the combinations.

Basically, in [m32_powf.c](https://github.com/z88dk/z88dk/blob/master/libsrc/_DEVELOPMENT/math/float/math32/c/m32_powf.c) any 2 of 3 work.

But not all three. And I've looked at the code, and moved things around, and nothing at all obvious...
These all work...
```c
return logf(x);
return x*y;
return logf(x)*y;
return expf(x*y);
```
But not...
```c
return expf(y* logf(x));
```
gives
```asm
l_m32_powf_00108:
	ld	e,(ix+6)
	ld	d,(ix+7)
	ld	l,(ix+4)
	ld	h,(ix+5)
	call	_m32_logf
	ld	c, l
	ld	l,(ix+10)
	ld	b,h
	ld	h,(ix+11)
	push	hl
	ld	l,(ix+8)
	ld	h,(ix+9)
	push	hl
	push	de
	push	bc
	call	___fsmul_callee
	call	_m32_expf
```
which I can't fault. But, it doesn't work.

feilipu:
Philippe has updated sdcc again, so I'll see if that improves things.

suborb:
With me it's definitely `logf()` that's causing issues. Replacing it with the original cephes code and I get sensible results for `logf()`. This is with the corrected prototype for `(m32_)ldexp()`

feilipu:
Ok. I'll assume that is where I need to look my end too.

suborb:
(I'm using sccz80/classic) 

I've got a simple loop from 0 -> 20 incrementing by 0.1 and calling `log()` with each value and then printing the hex and the floating point.

I've started off with a vanilla cephes `log()` and replacing it with bits of your optimised one. I'm running into problems with this line for the value: log(0.500000) we end up with the result 0xff80feb2 which i think is `FLT_MIN`.

```
y -= m32_div2f(z);
```

But also with this line:

```
y -= 0.5 * z;
```

In sccz80 the latter is replaced by an ldexp() twiddle. It doesn't consider negative numbers so `y += -0.5 * z` works.

I've got to do "daytime" things now, but I'll continue investigating later if you don't get anywhere.

Test program: 

[log.c.txt](https://github.com/z88dk/z88dk/files/3293024/log.c.txt)

Compile with `zcc +test log.c -fp-mode=ieee -lmath32`, run with `z88dk-ticks a.bin`

feilipu:
In hindsight.... it is obvious that the `dec` preserves carry, not sign... so the sign needs to be preserved.
Fixing now... doh.

feilipu:
Feels good to be able to include the `powf()` test. :tada:

feilipu:
I have fixed all I can find with, and I think that now there is still an issue with sccz80 somewhere.
The problems I was facing have all been resolved.

I think there is an issue because when I build the test program, I'm getting an odd number of pops following functions. As per below for example.
```sh
$ zcc +test --list @ftest80.lst -fp-mode=ieee -lmath32
$ z88dk-ticks a.bin
```

```
81    0074  CD 00 00    	call	pow
82    0077  C1          	pop	bc
83    0078  C1          	pop	bc
84    0079  C1          	pop	bc
85    007A  C1          	pop	bc
86    007B  C1          	pop	bc                           <- should be 4 pops
87    007C  CD 00 00    	call	l_plong
88    007F  21 10 00    	ld	hl,16	;const
89    0082  39          	add	hl,sp
90    0083  E5          	push	hl
91    0084  21 06 00    	ld	hl,6	;const
92    0087  39          	add	hl,sp
93    0088  CD 00 00    	call	l_glong
94    008B  D5          	push	de
95    008C  E5          	push	hl
96    008D  CD 00 00    	call	log
97    0090  C1          	pop	bc
98    0091  C1          	pop	bc
99    0092  C1          	pop	bc                           <- should be 2 pops
100   0093  CD 00 00    	call	l_plong
101   0096  21 0C 00    	ld	hl,12	;const
102   0099  39          	add	hl,sp
103   009A  E5          	push	hl
104   009B  21 06 00    	ld	hl,6	;const
105   009E  39          	add	hl,sp
106   009F  CD 00 00    	call	l_glong
107   00A2  D5          	push	de
108   00A3  E5          	push	hl
109   00A4  CD 00 00    	call	exp
110   00A7  C1          	pop	bc
111   00A8  C1          	pop	bc
112   00A9  C1          	pop	bc                           <- should be 2 pops
```

libsrc/_DEVELOPMENT/math/float/math32/z80/f32_fsldexp.asm
```diff
@@ -61,7 +61,7 @@ PUBLIC _m32_ldexpf
     add d
     ld d,a                      ; exponent returned
 
-    rl e                        ; get sign back
+    sla e                       ; restore sign to C
```
suborb: I think we still need to check the exponent, and if zero, jp to m32_zero_fastcall

feilipu: yes. that is a worthwhile check. not the root cause here. but good to add :+1:

suborb:
That's the right number of pops - it's storing the destination address on the stack which is popped into bc just before l_plong.

feilipu:
OK. Then I'm not sure where the problem is at.

I have one test program that runs to completion with libraries and linkage through sdcc, but it hangs after a few iterations with sccz80.

I am suspicious of poly() somehow, (after finding a few issue already) as that is the thing that will burn stack with multiple iterations.

feilipu:
yes. will send it now. also just fixing the check for zero in frexp.

feilipu:
[ftest.c.txt](https://github.com/z88dk/z88dk/files/3293313/ftest.c.txt)

suborb:
Thanks, for multiple argument functions with sccz80 you need to stick a `__stdc` decorator on them to get the arguments to be pushed right -> left

feilipu:
Ok. that will be why my program is breaking then.

libsrc/_DEVELOPMENT/math/float/math32/z80/f32_fsldexp.asm
feilipu: Put it after returning the sign the the right bit. That way fszero picks up the correct sign.

suborb: Isn't that checking that the (pw2 + exp) is zero? - i.e. the new value of the exponent. I think we need to check that incoming exponent is 0.

feilipu: Both, I'd say. Neither are permitted to succeed.

suborb: Good point!

I’ll steal this pattern for the other libraries.

feilipu:
I'll merge this soon as a check clears. Then we've a clean start for finding out why the sccz80 pow function produces entirely the wrong answer for `+test` but works perfectly for `+rc2014`.

suborb:
Cool. I know what I’ll be doing this evening then!

So rc2014 sccz80 works. Test sccz80 fails? That’s very odd - I think it’s callee in newlib, standard in classic?

suborb:
[Continuing on a closed PR since the conversation is here]

I've now added the sdcc intrinsics to the sccz80 math32.lib - `m32_powf()` etc needs them.

I've been running around in circles with this one, I have hunch due to a timestamp issue and old .o files lying around the place.

However I've finally been able to turn on the `pow()` test suite for math32.

feilipu:
A lot of bugs squashed, and other improvement made in this PR. Glad that we've got the accuracy issue beaten (by not squashing the `hl` register), and balancing the stack in the connections.

Still need to do the zsdcc newlib build, but that is for another day.

PR #1202, opened 2019-06-17
m32 - div2 better

feilipu:


libsrc/_DEVELOPMENT/math/float/math32/z80/f32_fsdiv2.asm
```diff
@@ -28,16 +28,14 @@ PUBLIC _m32_div2f
 .m32_fsdiv2_fastcall
     sla e                       ; get exponent in d
     rl d                        ; put sign in C
-    rr e                        ; save sign in e[7]
 
-    xor a
-    or d
+    inc d
+    dec d
     jp Z,m32_fszero_fastcall
```
ped7g: The `rl d` does already set ZF by new content of `d`, so you don't need to do `inc + dec` flag test (there was some other rotation with `e` before trashing ZF, but seems like you managed to [re]move that).

feilipu: Yes, the code originally relied on preservation of the C flag to store the sign bit through `dec d`. Later adding the check for zero exponent requires the sign bit to be put into c[7] for safe storage. But then later I remembered how I'd done it with `mul2f` and reverted to `inc d dec d` which makes preserving the sign obsolete.

There is still a penalty of 18 ticks for checking for zero, but this way shaves 8 ticks off, and preserves the a register.

PR #1206, opened 2019-06-19
m32 - rework zero exits

feilipu:
- try to support negative zero properly, by using sign bit properly.
- use `jr` in main-line code to save 3 cycles per non-taken test.

### DIFF
--- a/libsrc/_DEVELOPMENT/math/float/math32/lm32/c/sccz80/l_f32_ldexp.asm
+++ b/libsrc/_DEVELOPMENT/math/float/math32/lm32/c/sccz80/l_f32_ldexp.asm
@@ -6,26 +6,37 @@
 
         PUBLIC  l_f32_ldexp
 
-; Entry: a = adjustment for exponent
-;       Stack: float, ret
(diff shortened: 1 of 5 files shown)


suborb:
@feilipu just as a note, m32_compare_callee does not handle -0 correctly

feilipu:
@suborb I think they are supposed to [compare as equals](https://en.wikipedia.org/wiki/Signed_zero#Comparisons). I'll have a look to see if I can fix it.

feilipu:
Looking at the implemented decision tree for `m32_compare_callee`, it looks like a subtraction first, then evaluate signs. Is that a correct understanding?

Perhaps the decision tree needs to first look at exp and sign to capture special cases for infinities and zeros, and then work on mantissa for greater or lesser? It might be better to split the 4 byte representation out into sign, exponent and mantissa for this process.

Further note from wikipedia for my reference: Besides the more obvious results, IEEE 754 defines that −∞ = −∞, +∞ = +∞ and x ≠ NaN for any x (including NaN).

PR #1232, opened 2019-07-17
Add 8080 support

suborb:
This change is getting quite large, but it's mostly complete.

Supported for classic + sccz80 only

### Not Supported

* Long multiplication/division
* Numerous additional library routines

### sccz80 changes

* Added `-m8080 flag` and hopefully avoided all use for z80 opcodes when enabled
* Cleaned up the stack cleanup code a little since it was a tangly, spidery message

### z80asm changes

* Added `-m8080 flag` which enables:
* Emulation calls for the block move instructions (ldir, ldi, ldd, ldd)
* Mapping of `jr cc,MM` and `djnz MMM` into appropriate codes. Breaks if a `jr` has `ASMPC` in it (Accepted risk)
* @pauloscustodio I couldn't figure out how to get 8080 tests working so I've had to disable them.

### copt changes

* Added `#notcpu` qualifier for rules (we can add a `#cpu` later)

### ticks changes

* Added `-m8080` mode which disables use of z80 opcodes

### Testsuite changes

* Enabled a number of tests so that they run on the emulated 8080

### Classic changes

* Created `8080_crt0.lib` with appropriate code
* Created `cpm8080_clib.lib` and appropriate `-clib=8080` option for CP/M target

### Newlib changes

* Various tweaks to shared library files to detect `__CPU_8080__`


pauloscustodio:
> @pauloscustodio I couldn't figure out how to get 8080 tests working so I've had to disable them.
No problem, I'll have a look.

suborb:
Thanks Paulo.

I'm going to merge this as is and do additional work (Eg restoring longs) in a different PR.

pauloscustodio:
8080 tests in z80asm are now fixed.

PR #1240, opened 2019-07-19
m32 - some newlib helpers

feilipu:
Adding some newlib helpers, complying with the existing ABI of the float (double) in the `exx` register set.


feilipu:
This is `__dtoa_base10`, `__dtoa_base16`, `__dtoa_digits`, and `__dtoa_sgnabs` for mtath32. Getting there required adding `exp10f`, and `exp2f` and `log2f` came with them, plus a full implementation of `log10f`.

TBH, not sure whether the `_dtoa_` stuff works, as it will need to be linked to be used (which is not done yet).

PR #1263, opened 2019-08-11
yaz180 i2c - rework

feilipu:
Giving up on multi-bank driver support for i2c, and just making it work as user application drivers.
Now for testing.


feilipu:
Likely to be some more fixes but I've gotten git tangled. So better to do small PRs.

PR #1315, opened 2019-11-01
[scz180] - new platform

feilipu:
A new target based on Stephen C Cousins' z180 modular platforms.

- [SC126](https://smallcomputercentral.wordpress.com/sc126-z180-motherboard-rc2014/)
- [SC130](https://smallcomputercentral.wordpress.com/sc130-z180-motherboard/)

Intention is to support:

- [ ] ROM subtarget - which would be written to bare metal.
 - [ ] APP subtarget - which would rely on the [SCM](https://smallcomputercentral.wordpress.com/small-computer-monitor/) and its [entry points](https://smallcomputercentral.wordpress.com/small-computer-monitor/small-computer-monitor-v1-1/scm-v1-1-reference/).
__TO DO__ Later, since SCM isn't supported on SC130 yet. Banking not yet supported either.
 - [x] CPM subtarget - which relies on CP/M 2.2

TODO
 - [x] Configure the IO addresses
 - [x] Device drivers for the ASCI and 8255 IDE rework for low addresses.
 - [x] Device drivers for the SPI interface and SD card.
 - [x] Integrate diskio functions for both IDE and SD card. Found in [z88dk-lib](https://github.com/feilipu/z88dk-libraries/blob/master/ff/source/diskio_sd.c)
 - [ ] Drivers for SC126 delta vs SC130
__TO DO__ Later, difficult with no hardware to test.


include/fcntl.h
```diff
@@ -83,11 +83,11 @@ extern unsigned int   RND_BLOCKSIZE;
 struct RND_FILE {
 	char    name_prefix;   /* block name, including prefix char */
 	char    name[15];         /* file name */
-	u16_t	blocksize;
+	i16_t	blocksize;
 	unsigned char *blockptr;
 	long    size;             /* file size */
 	long    position;         /* current position in file */
-	u16_t	pos_in_block;
+	i16_t	pos_in_block;
```
feilipu: `(a)->pos_in_block=-1;`
causes a warning about type mismatch.
`blocksize` is never more than 1,000 so `i16_t`, to match `pos_in_block`, should be sufficient.
Right?

libsrc/_DEVELOPMENT/Winmake.bat
```diff
@@ -37,18 +37,20 @@ for %%t in (%targets%) do (
    if not "!temp!" == "%alltargets%" (
 
       set cpu=
+
+      if "%%t" == "scz180" (
```
feilipu: Flattened out the decision structure. It was getting unwieldy.
But, can't test this without a Windows build.

PR #1337, opened 2019-11-20
Videoton TV Computer - initial PR

dikdom:
This is the first iteration of the TVC support. The CAS file generation may be incorrect, I need to check with experts the block counting.
It prints to the console
importing time.h the clock ticks are well printed
joystick is read 
Simple programs can be compiled.


libsrc/stdio/tvc/fputc_cons.asm
```diff
@@ -0,0 +1,28 @@
...
+	PUBLIC  fputc_cons_native
+    INCLUDE "target/tvc/def/tvc.def"
+
+;
+; Entry:        stack contains a char
+;
+.fputc_cons_native
+    ld      hl,2
+    add     hl,sp
+    ld      a,(hl)
```
suborb: Does the value need to be in a and c? Just use `ld c,(hl)`?

libsrc/time/tvc/clock.asm
```diff
@@ -0,0 +1,24 @@
...
+;
+
+        SECTION code_clib
+        PUBLIC	clock
+        PUBLIC	_clock
+        include "target/tvc/def/tvc.def"
+.clock
+._clock
+;    INTINC is 0b1dh
+    di
```
suborb: Please use `call l_push_di` and `call l_pop_ei` - these functions preserve the interrupt enabled flag - i.e. `l_pop_ei` won't enable interrupts if they were disabled when `l_push_di` was called.

libsrc/stdio/tvc/fputc_cons.asm
dikdom: A simple copy-paste bug. Enough to have it in C, I will fix it.

ped7g: if you commit on the branch from where this PR was created, it will get incrementally added to this PR (as far as I can tell how it works for me on other github projects, didn't try PR for z88dk yet)

libsrc/time/tvc/clock.asm
dikdom: As a matter of fact there is no need to do ei/di as this counter is incremented in the TVC's interrupt routine. I fix this too by eliminating the ei/di pairs.

suborb:
Sorry, I've run out of time to try this out tonight but it's looks good. I've also managed to create a merge conflict (sorry) - could you fix it please?

Could you also create a wiki page linked from here: https://github.com/z88dk/z88dk/wiki/Platform

The format is basically:

- Hardware spec
- Features supported by z88dk
- Example compile line + how to load the software into an emulator
- Notes on any features of the libraries
- Links (including to the emulator)

dikdom:
Wiki updated, conflict fixed

suborb:
Nice wiki page, many thanks.

PR #1345, opened 2019-11-30
(rc2014) subtype hbios

feilipu:
Add a RC2014 `hbios` subtype for applications where the ROMWBW HBIOS is present.

This fits logically between the `acia` & `sio` subtypes for use on raw hardware and the `cpm` subtype which requires the CP/M BDOS functions to operate.

The `hbios` subtype uses the functions defined in [`arch/hbios`](https://github.com/z88dk/z88dk/tree/master/libsrc/_DEVELOPMENT/arch/hbios) to provide serial (and later diskio) functions.

The `hbios` subtype can be loaded into the ROMWBW monitor running on a RC2014, or it can be used when CP/M is running (but depending on the size of the application it may overwrite CP/M BIOS).


feilipu:
This is good now.

PR #1346, opened 2019-11-30
TVC console and keyboard handling

dikdom:
- getch and getk implemented in stdio
- various print and keyboard input related tvc specific functions implemented
- fixed ei/di handling around joy query


suborb:
Sorry, I was on mobile when I made the comments a couple of days ago and it started a review for my comments which is not what I actually wanted!

libsrc/tvc/tvc_fgets_cons.c
```diff
@@ -0,0 +1,30 @@
...
+    int pos=0;
+    char c;
+    asm_tvc_ed_cfix();
+    str[pos] = 0;
+    do {
+        c = asm_tvc_ed_getch();
+        if(c == TVC_CHAR_ESC) {
+            str[0] = 0x00;
+            break;
+        } else if (c == TVC_CHAR_RETURN) {
```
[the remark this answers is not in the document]
dikdom: Hm... gets() and fgets() behave differently. Nice.

dikdom:
Done.

suborb:
Thank you. I’m away at the moment so can’t try them out - I should get a chance at the weekend though.

include/tvc.h
```diff
@@ -85,4 +86,12 @@ extern int __LIB__ getk();
 #define tvc_clrscr asm_tvc_cls
 extern void __LIB__ asm_tvc_cls(void);
 
+
+/**
+ * Sets the video mode of TVC (2 colours, 4 colours, 16 colours), clears screen
+ * and initialize cursor positions.
+ */
+#define tvc_vmode asm_tvc_vmode
```
suborb: We tend to avoid having asm_XXXX() functions in the header file - functions named like that are for use by assembler programmers since they have register based input parameters.

When they are the same we create a file in the library that does something like this:

```
PUBLIC tvc_vmode
EXTERN asm_tvc_vmode

defc tvc_vmode = asm_tvc_vmode
```

include/tvc.h
```diff
@@ -75,7 +77,6 @@ extern int __LIB__ asm_tvc_kbd_status();
 extern int __LIB__ getk();
```
suborb: I don't understand why this is here since it's already defined in stdio.h

libsrc/target/tvc/asm_tvc_vmode.asm
```diff
@@ -0,0 +1,25 @@
...
+;
+
+	SECTION code_clib
+	PUBLIC  asm_tvc_vmode
+    INCLUDE "target/tvc/def/tvc.def"
+
+;
+; Entry:        0, 1, 2 - 2c, 4c, 16c mode
+;
+.asm_tvc_vmode
```
suborb: Regarding my comment above - it looks like this asm_ function has got C parameter parsing. In this case, it would be best named tvc_vmode and line 21 gains the asm_tvc_vmode label.

include/tvc.h
dikdom: Several cases I didn't understand the pattern. Thanks for the guides I will fix these issues soon.

libsrc/target/tvc/asm_tvc_vmode.asm
dikdom: asm_ files refactored.

include/tvc.h
dikdom: removed

include/tvc.h
```diff
@@ -60,38 +72,61 @@ extern char __LIB__ fputc_cons_native(int character);
  * earlier this call will block (func $A1)
  */
 #define tvc_kbd_chin     fgetc_cons
-extern int __LIB__ fgetc_cons();
+int __LIB__ fgetc_cons();
```
suborb: Is this one needed? cf getk() comment previously

libsrc/target/tvc/tvc_get_bordercolor.asm
```diff
@@ -0,0 +1,22 @@
...
+    INCLUDE "target/tvc/def/tvc.def"
+
+;
+; Entry:        none
+; In order to work properly U0 must be mapped in.
+;
+.tvc_get_bordercolor
+._tvc_get_bordercolor
+
+    ld      a, (BORDER)
```
suborb: I normally save a byte and do a ld hl,(BORDER); ld h,0

libsrc/target/tvc/tvc_set_ink.c
```diff
@@ -0,0 +1,23 @@
+#include <tvc.h>
+#include <stdint.h>
+
+void tvc_set_ink(color_or_index c) {
+    uint8_t *memP;
+    int res = tvc_get_vmode();
+    switch (res) {
+        case 0:  // 2 color
+        case 1:  // 4 color
+            memP = (uint8_t *)0x0B4D;
```
suborb: You can actually do this which generates slightly more compact code:
    
        *((uint8_t *)0x0b4d) = c.paletteIndex

include/tvc.h
dikdom: Nope, removed.

libsrc/target/tvc/tvc_get_bordercolor.asm
dikdom: ... and few cycle as well. Refactored.

libsrc/target/tvc/tvc_set_ink.c
dikdom: Done.

dikdom:
Requested changes are done.

suborb:
Thank you for your work on this one - it's really appreciated!

PR #1363, opened 2019-12-23
[newlib] - add hbios target

feilipu:
The [ROMWBW](https://github.com/wwarthen/RomWBW) [HBIOS](https://github.com/wwarthen/RomWBW/tree/master/Source/HBIOS/Config) supports a number of different machines, including SBC 1/2, Zeta 1/2, N8, Mark IV, RC2014, SCZ180, and Easy Z80.

The HBIOS support in both RC2014 and SCZ180 platforms, has been pulled out to a separate target for HBIOS, similar to the newlib CP/M target, but limited to the "hardware abstraction layer" drivers only.  This allows HBIOS support for more machines, without a proliferation of new targets.

At this stage, I don't think it would be more than a collection of HAL drivers, because there's no need to have a CCP or BDOS equivalent in the target.

The HBIOS target includes unified access to the disk, serial, video interfaces via the [described API](https://github.com/wwarthen/RomWBW/blob/master/Doc/RomWBW%20Architecture.pdf).


feilipu:
Still some issues with `fscanf()` across the platforms, but this is good enough for now.

PR #1383, opened 2020-01-25
(classic) Graphics reorganisation

suborb:
Rather a big reorganisation.

Split the graphics library into a number of subsets:

- narrow (i.e. <= 256x256)
  - gencon: 1x1
  - gencon: 2x2
  - gencon: 2x3
  - text (2x2)
  - text6 (2x3)
- wide
- portable (C code, good for 8080/gbz80)

And compile/assemble for each target. This way we don't need to constantly clean or miss that and end up with the wrong code in a library.


suborb:
@zx70 can I get your eye over this one - it's a big change in a delicate area - I've tested at least one of each variant but I'm bound to have missed something...

zx70:
> @zx70 can I get your eye over this one - it's a big change in a delicate area - I've tested at least one of each variant but I'm bound to have missed something...

I'll have a look  :)

zx70:
First impression is positive, it all builds and most of the stuff seeems to be running correctly.
I see stencil.c crashing in low resolution (-lgfx) on the ZX81, I wonder if it happened also with the old version.

suborb:
Thanks, I'm sure there's a few bits hanging around, but lets merge!

PR #1384, opened 2020-01-26
rc2014 - LUT Module support

feilipu:
Support for a LUT Module, to provide fast multiplies for z80, and also sin, log, ln, crc, etc tables.

Based on a [design discussion on the RC2014 Forum](https://groups.google.com/d/msg/rc2014-z80/Oly0f1qwZv8/2e4kFiBCEwAJ).

@suborb I'm not sure how to get functions in the device library to take precedence over functions in the math library. Specifically the LUT implementation of the `16_8x8` multiply should take precedence. I've played with different options but, short of rewiring the whole `math32` library, I can't get the device library to take.


suborb:
Normally putting it on the zcc command line after the maths library will mean it’s placed before the maths library on the z80asm command line. 

As a result the routines in the lut library would be picked up first.

Run with -Cl-v and see what the order is.

feilipu:
The (my) problem is I’ve put the functionality into ~/target/rc2014/device/lut rather than an external library. So I guess the linked math library is taking precedence.

I could work around this issue by supplying a specific 16_8x8 multiply for each target. Not a bad idea, because it would allow one math32 library and different platform specifics.

It would reduce the four math32 targets down to one, with the multiply based on hardware available (z80, z180, z80n, or lut). But, more work. But cleanup of options. 

Classic would have to supply the multiply routine then too.

feilipu:
Thinking about this for a few hours. `math32` is too complicated the way it is today. There are only 3 files specific to each hardware target and they can be built by the target. Then there can be just one `math32` library, which can be linked with just one configuration `--math32`.

Then separately we can consider to keep or lose the `--math32_fast` option, particularly when the only platform it is useful for will soon have a "`--math32_lut`" option too. I'm of a mind to lose software table option, or at least take it out of play as a standard alternative.

I'll do a separate PR for that change, before this one.

suborb:
A few hours of not thinking from me...

The other way to do this to this is to add a -pragma-redirect to the command line and point the label to a different one. Because this is part of the crt it’s processed first so there’s no linking order issues.

feilipu:
Marie Kondo wandered past, and chucked everything out, with PR #1385. I think this looks much better.

I've not played around with the classic builds, so up to you if you'd like to clean up too.

suborb:
I'd quite like to clean up the classic ones, the path of least resistance is to move those 3 routines from the math32 library into the crt0 libraries. I think this would drop the lookup table l_mulu_de.

feilipu:
Yes agreed to drop the table `l_mulu_de`, and then to obsolete / remove the `__CLIB_OPT_FMATH` option. Removing the table multiply also simplifies the larger multiplies for z80, as only one version of the code (using the register preserving `l_mulu_de`, equivalent to the z80n code) remains. Another "Kondo moment".

I chose to leave the 3x large multiply files within the library, and just move the build linkage over to the target. Probably the same path will work for classic too.

feilipu:
I'll merge this, and open a new issue for simplifying classic `math32`.

PR #1405, opened 2020-02-18
Feature/m100

zx70:


suborb:
Target number 100!

libsrc/target/m100/stdio/generic_console.asm
```diff
@@ -0,0 +1,62 @@
...
+generic_console_set_paper:
+generic_console_set_attribute:
+	ret
+
+
+generic_console_set_ink:
+	ret
+
+generic_console_scrollup:
+	call $45ED	; SCROLL_ALTLCD
```
suborb: scrollup needs to preserve bc,de - does SCROLL_ALTLCD do that?

zx70: Looks like a valid hint.  I'm still improving the whole thing,  e.g. locations vary depending on the ROM versions, etc..

zx70:
https://github.com/z88dk/z88dk/commit/b3029f1422cbf3638cc74e5a331ad95ec4fb0a6c

rpn.c is now reasonably ok,  tic.c is still bad when scrolling.

PR #1410, opened 2020-02-23
newlib - FreeRTOS includes

feilipu:
As the FreeRTOS system provides multiple header files at this stage the only way to get it supported is to build it into the library.

FreeRTOS uses the MIT licence so there is AFAIK no impairment or implication by including these files in the repository.

No C files are included at this stage, because the library can be added later and will be target specific (timer functions).

This initial PR is currently using ordinary calling, but will probably migrate to fastcall / callee when a bit better developed.


feilipu:
Need to note that tonight I've got FreeRTOS successfully running on YAZ180.
Using these [z88dk-libraries](https://github.com/feilipu/z88dk-libraries/tree/master/freertos).

This PR needs to add some specific configurations for other Z180 platforms, including HBIOS on SCZ180. I'd also like to add the ZXN target too, before it is closed.

But... with running code things are so much easier. :smile:

feilipu:
Now SCZ180 with HBIOS works (mostly).

Still an issue of clashing with HBIOS bank swapping, but it is getting close to correct.

feilipu:
Now with the TPA bank test correctly set, it works successfully on the SCZ180 running the RomWBW HBIOS.

I don't have a Timer Module for RC2014 (to be remedied at some stage), and I think understanding the ZXN memory mapping will be a bit of a challenge, so I'll merge this PR now.

The freeRTOS port is quite transferable to any other Z80 platform. Besides the usual freeRTOS API configurations, I've kept the platform dependencies in just one file. Within that file there are 4 macros needed

- set the Tick rate to be used by Timer macros
- configure the Timer and make sure the ISR is always available
- reset / reload the Timer if needed (as on z180)
- stop the Timer (not really necessary)

The libraries are built at z88dk-libs/freertos, and installed the usual way.
There are no header files installed, as they're included in this PR.

PR #1421, opened 2020-03-14
zsdcc - v4.0.0 bugfixes r11556

feilipu:
I hate to be doing another update so quickly, but there are some sdcc bugfixes that relieve workarounds with freertos, and improve 16-bit literal handling.

I've done some back to back code review, and I can't spot any other differences.

The freertos issues are sdcc bugs [#2817](https://sourceforge.net/p/sdcc/bugs/2817/) and [#2973](https://sourceforge.net/p/sdcc/bugs/2973/). Both of these issues can be worked around simply by removing the `const` decoration from some of the declarations. But then the z80 code deviates from the mainline code. There are no resulting assembly code differences.

Also an improved code for addition of a 16-bit literal was added in [r11550](https://sourceforge.net/p/sdcc/code/11550/).

from this
```asm
ld	a,(ix-5)
add	a,0x27
ld	l, a
ld	a,(ix-4)
adc	a,0x00
ld	h, a
ld	a, (hl)
```
to this
``` asm
ld	l,(ix-5)
ld	h,(ix-4)
ld	de,0x0027
add	hl, de
ld	a, (hl)
```
If it is ok to merge, please generate new windows/mac builds.


suborb:
I’ll try to get the builds done today and merge this when I’m done.

suborb:
I've rebuilt the binaries, so merging...

PR #1422, opened 2020-03-15
newlib targets - startup consistency / formatting

feilipu:
- Rework some of the newlib targets for consistency.
- Remove Page 0 config from subtype startups where a bios is already present.

Checking and testing before merging.


feilipu:
Tested across RC2014, SCZ180, YAZ180. I think this does nothing except align all the startup code (as it should).

PR #1423, opened 2020-03-17
RC2014 - Integer LUT math

feilipu:
As a companion to the floating point LUT math, some integer math.

Testing and larger multiplies before merging.


feilipu:
`I_lut_mulu_32_32x32` passes Pi to 800 places.

feilipu:
sdcc doesn't call 32_16x16 when multiplying `uint16_t`x`uint16_t` into a `uint32_t` result.
Add demotion to 32_16x16 to cover this case.

feilipu:
Calculating [pi to 800 decimal places](https://github.com/z88dk/z88dk/tree/master/libsrc/_DEVELOPMENT/EXAMPLES/benchmarks/pi) using `small_mulu_32_32x32`  takes `12'00"`, on RC2014.
Using `lut_mulu_32_32x32` takes `8'20"`. Identical results.
```sh
zcc +rc2014 -subtype=cpm -clib=sdcc_iy -SO3 --max-allocs-per-node400000 -DSTATIC -DPRINTF pi.c -o pi -create-app
```

Issue #1424, opened 2020-03-23
Can't build ZXNext examples


Dovgalyuk:
e.g., z88dk/libsrc/_DEVELOPMENT/EXAMPLES/zxn/sprites/mixed-sw-hw/main.c

It seems that the compiler can't find zxn includes. How to specify path to them?

Xalior:
This example works for me using the compile line example given in the main.c - have you set up your ZCCCFG as shown on https://github.com/z88dk/z88dk/wiki/installation ?

aralbrec:
Yes that is probably the problem.  But there is another problem with that compile line in that the zx next now treats the sna format as a legacy format so when loading an sna, the computer is set up as a 48k or 128k spectrum with all of the next hardware switched off.

Xalior:
I was going to open a pull request for that @aralbrec ;-)

Dovgalyuk:
$ echo $PATH
......:/home/pasha/ZXNext/z88dk/bin
$ echo $ZCCCFG
/home/pasha/ZXNext/z88dk/lib/config


test.c:
 #include <arch/zxn.h>
 int main()
 {
    return 0;
 }


$ zcc +zxn test.c
Error at file 'test.c' line 207: syntax error
.... and many more errors

aralbrec:
I downloaded the latest build for windows and that also runs fine.  Unfortunately I can't check the linux package until tonight.

Do you mind trying:

`which zcc`

to verify the right zcc is running?  If you installed z88dk previously, that older version may be earlier in your path.  Also, there are other projects that use a decade-old version of zcc in their installs that may also be earlier in the path.

Other than that, the next step would be to see those errors.

Can you:

`zcc +zxn -v test.c > foo.txt`

and report the result either download link or cut & paste here?  You may have to redirect stderr to a file too.

Dovgalyuk:
which zcc
this told me that I have zcc in /usr/local/bin
It seems that I installed it for some reason, when started playing with zcc.

After removing all z88dk files from /usr/local I'm still get that error messages.
[error.txt](https://github.com/z88dk/z88dk/files/4373304/error.txt)

Dovgalyuk:
Here are also the temporaries:

[tmpXXiStVNo.asm.txt](https://github.com/z88dk/z88dk/files/4373979/tmpXXiStVNo.asm.txt)
[...]

PR #1425, opened 2020-03-23
Update SpecNext build instructions in examples

Xalior:
The use of the .sna format now turns off much of the advanced hardware
found in the Spectrum Next by default (for legacy compatibility reasons)
therefore apps wishing to use these features should now use the extension
.SNX or, preferably, the Next specific .NEX format.

I've taken the liberty of updating all the examples in _DEVELOPMENT except
those that specifically reference the "bigsna" format.

This pull request is a product of discussion https://github.com/z88dk/z88dk/issues/1424#issuecomment-602651540

-Dx


suborb:
Thanks!

Issue #1428, opened 2020-03-28
Peephole suggestion


obiwanjacobi:
While debugging other stuff I saw a couple of instances of code like this:

```asm
ld hl, 0000
add sp, hl
```

I don't know how often this occurs, but it should be easy to optimize?
[2c]

feilipu:
I'm doing other stuff now, but please try the branch in #1429 to see if it helps this issue.
Perhaps there's some side-effects I've missed?

feilipu:
Just looking a bit deeper, I've not seen any examples where `ld hl, 0000` comes forward.
Usually I'm seeing `ld hl,0x0000`.

Do you have a piece of code that generates this form?

obiwanjacobi:
Uhh, right. I mixed it up - sorry.

```asm
ld      hl, $0000
add     hl, sp
```

Which is fine...
[This sitting at home is getting to me ;-)]

Again, sorry.

PR #1429, opened 2020-03-28
sdcc - peephole rule 146

feilipu:
New sdcc peephole rule based on #1428.


suborb:
@feilipu I'm going to close this one since I'm pretty confident `add sp,hl` can't be generated

PR #1430, opened 2020-03-28
noise management to wyz library effects

jsmolina:
Adding noise management in effects to wyz lib.

Original code by thEpOpE 

It works ok: example of it on 
https://github.com/jsmolina/wyz-z88dk-integration/releases/download/0.1/a.tap

(3 = hihat, 2=snare, 1=bass drum)

Caveats: it detects if noise channel  is present in effects, but you cannot mix two kind of effects. In case there is no noise channel, you must set 00, like in 
https://github.com/jsmolina/wyz-z88dk-integration/blob/master/effects.asm#L29-L30


suborb:
Where's the original modification from?

jsmolina:
It was thEpOpE's improvement to 'La Churrera' MK1 engine, so effects become much more powerful. 

I'm mistaking something in the code but I'm almost there...  (thEpOpE said it might be related to IX register and z88dk way of passing parameters to functions)

jsmolina:
@suborb  WDYT?

libsrc/psg/wyz/wyzplay.asm
```diff
@@ -980,19 +998,26 @@ RS_CANALC:
     LD      (IX+4),B
     LD      (IX+5),C
     LD      (IX+10),D
+poneruidomixer:		;este punto es comun a los 3 canales
+    ld a,e	;cargamos la máscara para el mixer
+    and (ix+7)	;lo aplicamos desde el buffer de registros del ay
```
suborb: @jsmolina we need to ensure that the top two bits of register 7 are 10b or MSX machines start to explode.

jsmolina: I've been talking with original author (thEpOpE), and looks like


1. Mixer (IX+7) it's set by WYZ tracker
2. Initially, 'E' has a common mask (%11110111)
this mask only nullifies one bit. C channel bit.

For channel B, register is rotated twice, cyclically. 
So the bit that is kicked our on the right is added from the left.

For channel A, register is rotated once.

So, for channel A: register E = %11110111
in channel B: register E= %11101111
in channel A: register E = %11011111

this mask is applied in three instructions
```
poneruidomixer:    ;este punto es comun a los 3 canales
      ld a,e  ;cargamos la máscara para el mixer
      and (ix+7)  ;lo aplicamos desde el buffer de registros del ay
      ld (ix+7),a  ;dejamos el resultado en el buffer de registros del ay
```

The problem are as you said with 2 first bits (bit 7 and bit 6). Rotations are cyclic, top bits are never altered.
These two high bits are enabling/disabling the two pins IN/OUT that AY has, so they must be always preserved, as ZX Spectrum also uses them.

Player also considers them here:
```
FIN_NOPLAYER:
            LD      A,10111000B
            LD      [PSG_REG+7],A
            RET
```

suborb: Great comment, thank you

PR #1436, opened 2020-04-02
newlib - demote uint16_t inputs in mulu_32_32x32

feilipu:
For both compilers (for z80) multiplying two `uint16_t` variables will either;

- `__mulint_callee` - truncate the result to `uint16_t` (if no casting), or
- `__mulong_callee` - if a `uint32_t` result is required and one of the inputs is cast to indicate this, then issue a `l_mul_32_32x32`.

This does PR checks whether the inputs to `__mulong_callee` are `uint16_t`, and if so issues  a `l_small_mul_32_16x16`, which is substantially faster.

This would match the z180 and z80n solutions.

@suborb classic z80 `~/z88dk/libsrc/z80_crt0s/z80/sccz80/l_long_mult.asm` doesn't `defc` to `l_small_mul_32_32x32`. I've left that untouched, but it might be worth moving over to newlib too?

__EDIT__ the reason I've put this up for review is that the `l_small_` functions are small, and this modification pulls the `32_16x16` code into every build using `32_32x32`, which may not be acceptably small.

### DIFF
--- a/libsrc/_DEVELOPMENT/math/integer/small/l_small_mul_32_32x32.asm
+++ b/libsrc/_DEVELOPMENT/math/integer/small/l_small_mul_32_32x32.asm
@@ -1,77 +1,105 @@
 
 ; 2000 djm
 ; 2007 aralbrec - use bcbc' rather than bytes indexed by ix per djm suggestion
-
-INCLUDE "config_private.inc"
+; 2020 feilipu - demote uint16_t
 
 SECTION code_clib
 SECTION code_math
 
+EXTERN l_small_mul_32_16x16
+
 PUBLIC l_small_mul_32_32x32, l0_small_mul_32_32x32
+
+l0_small_mul_32_16x16:
+
+    ; multiplication of two 16-bit numbers into a 32-bit product
+    ;
+    ; enter : hl'= 16-bit multiplier   = y
+    ;         hl = 16-bit multiplicand = x
+    ;
+    ; exit  : dehl = 32-bit product
+    ;         carry reset
+    ;
+    ; uses  : af, hl, bc', de', hl'
+
+    push hl
+    exx
+    pop de
+    jp l_small_mul_32_16x16
 
 l_small_mul_32_32x32:
 
    ; multiplication of two 32-bit numbers into a 32-bit product
    ;
-   ; enter : dehl = 32-bit multiplicand
-   ;         dehl'= 32-bit multiplicand (more leading zeroes = better performance)
+   ; enter : dehl = 32-bit multiplicand (more leading zeroes = better performance)
+   ;         dehl'= 32-bit multiplicand
    ;
    ; exit  : dehl = 32-bit product
    ;         carry reset
    ;
    ; uses  : af, bc, de, hl, bc', de', hl'
 
-   xor a
-   push hl
-   exx
-   ld c,l
-   ld b,h
-   pop hl
-   push de
-   ex de,hl
-   ld l,a
-   ld h,a
-   exx
-   pop bc
-   ld l,a
-   ld h,a
+
+    xor a
+    or e
+    or d
+
+    exx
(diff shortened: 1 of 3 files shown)


feilipu:
@suborb just looking at the `__CPU_INTEL__` version of [`l_small_mul_32_16x16`](https://github.com/z88dk/z88dk/blob/170fdb62c3ef066340502797c4b78bb1a8555876/libsrc/_DEVELOPMENT/math/integer/small/l_small_mul_32_16x16.asm#L40), and I think there's a problem, because the carry from `rl d` is being overwritten by popping the carry from `add hl,hl`. Thoughts?

suborb:
It looks like I overlooked the multiplication delegation, I'm not sure why. The intention for the classic crt0 routines is to delegate to `l_mulu_` routines. By default this binds to the `l_small_` implementations. The fast implementations are switched to at compile time using `-lfastmath`.

And yes, the Intel one was wrong - it's never called automatically since `l_long_mult` for 8080 is an implementation taken from ACK. However, this might be a good optimisation to do for 8080 since there's a lot of memory access in the 8080 flavour.

In regards this PR, this looks like it adds 36/37 (I lost count!) bytes. I suspect that this is an optimisation that will be hit quite frequently - what sort of improvement are you seeing?

feilipu:
> It looks like I overlooked the multiplication delegation, I'm not sure why.

Possibly because the `l_small_mulu_32_32x32` isn’t easily convertible to `__CPU_INTEL___`? As it is based on alternate register optimisations. Your ACK version doesn’t have this issue.

Perhaps it would be “nicer” just to put them as alternates in the one `l_small_mulu_32_32x32` file?

> And yes, the Intel `l_small_mul_32_16x16` was wrong - it's never called automatically since `l_long_mult` for 8080 is an implementation taken from ACK.

The `l_small_32_16x16` was not implemented until recently (because it wasn’t ever needed by the compilers).

> However, this might be a good optimisation to do for 8080 since there's a lot of memory access in the 8080 flavour.

I’m sure it would help.

> In regards this PR, this looks like it adds 36/37 (I lost count!) bytes. I suspect that this is an optimisation that will be hit quite frequently - what sort of improvement are you seeing?

I’ve not put it through benchmarks formally for z80, as I was playing with LUT multiplication when I noticed the compilers don’t automatically recognise that a `int16` multiplied by an `int16` stored into a `int32` should produce a 32 bit result.

For the LUT version it was more than twice as fast for `int16` multiplication. For the z80 version, it must be about >3x faster, as it has about half the code, to traverse for half the iterations.

I don’t know whether it is a C standard or not, but I was quite surprised to learn that it is not sufficient to make the result type of a 16x16 multiply 32 bits to get a 32 bit answer. The multiplication itself needs to be promoted by a cast to 32 bits. A gotcha.

feilipu:
I did some benchmarking using Pi to 800 places.
Results below look quite useful, for the addition of about 40 bytes to the binary.

__`PI.C`__

__sccz80/small__

cycle count  = 5246791210 (March 2, 2017)
cycle count  = 5240675831 (April 4, 2020)
cycle count  = 4018350135 (April 4, 2020, l_long_mult-demote)

__zsdcc/small__

cycle count  = 5278798872 (March 2, 2017, ZSDCC #9833)
cycle count  = 5296259552 (April 4, 2020, ZSDCC #11556)
cycle count  = 4067517071 (April 4, 2020,  ZSDCC #11556, l_long_mult-demote)

__`PI_LDIV.C`__

__sccz80/small__

cycle count  = 3810732458 (March 2, 2017)
cycle count  = 2576381983 (April 4, 2020, l_long_mult-demote)

__zsdcc/small__

cycle count  = 3827247920 (March 2, 2017, ZSDCC #9833) 
cycle count  = 2609489119 (April 4, 2020, ZSDCC #11556, l_long_mult-demote)

PR #1447, opened 2020-04-16
spectrum native added docs and zx_paper, zx_ink and zx_setattr

Baltasarq:
Changelog:
- Added documentation
- Added `zx_setpaper`, `zx_setink`, and `zx_setattr`.


include/spectrum.h
```diff
@@ -317,10 +317,28 @@ extern void __LIB__   in_MouseKempSetPos_callee(uint xcoord, uint ycoord) __smal
...
 // Quickly set the whole screen color attributes
+// Param colour must be in the form i | p, where
+// i can be any of: INK_BLACK, INK_BLUE,... to INK_WHITE
+// p can be any of: PAPER_BLACK, PAPER_BLUE,... to PAPER_WHITE
 extern void  __LIB__  zx_colour(uchar colour) __z88dk_fastcall;
+// Change the paper & ink attrs from now on
+// Param colour must be in the form i | p, where
+// i can be any of: INK_BLACK, INK_BLUE,... to INK_WHITE
+// p can be any of: PAPER_BLACK, PAPER_BLUE,... to PAPER_WHITE
+#define zx_setattr(colour) fputc_cons(16); fputc_cons(48 + (colour & 0x0f)); fputc_cons(17); fputc_cons(48 + (colour >> 3))
```
suborb, line 335: I wonder if this might be better as a library function - it depends on how often it's called of course but it's quite a bulky macro.

Baltasarq: Are you afraid of incrementing the binary size of the user's program?

Baltasarq: Maybe we can stick to `zx_setpaper` and `zx_setink` and forget about `zx_setattr()`...

suborb: Yes, it's about 20/30 bytes per instantiation I guess, so if it's used a lot then having it as a library function is better.

Probably easiest to stick with zx_setpaper and zx_setink

PR #1448, opened 2020-04-16
Added zx_print(s) so you don't need to include stdio.h

Baltasarq:
Changelog:
- Added `zx_print(s)` to spectrum.h so you don't need to include stdio.h

### DIFF
--- a/include/spectrum.h
+++ b/include/spectrum.h
@@ -335,8 +335,7 @@ extern uint  __LIB__    zx_screenstr_callee(uchar row, uchar col) __smallc __z88
 #define zx_topleft()              fputc_cons(22); fputc_cons(0); fputc_cons(0);
 
 // Print a string on the screen
-extern int printk(char *fmt, ...);
-#define zx_print(s) printk(s)
+extern int zx_printf(char *fmt, ...);
 
 // In the following, screen address refers to a pixel address within the display file while
 // attribute address refers to the attributes area.
--- a/include/spectrum.h
+++ b/include/spectrum.h
@@ -334,6 +334,10 @@ extern uint  __LIB__    zx_screenstr_callee(uchar row, uchar col) __smallc __z88
 #define zx_setcursorpos(a,b)      fputc_cons(22); fputc_cons(a); fputc_cons(b);
 #define zx_topleft()              fputc_cons(22); fputc_cons(0); fputc_cons(0);
 
+// Print a string on the screen
+extern int printk(char *fmt, ...);
+#define zx_print(s) printk(s)
+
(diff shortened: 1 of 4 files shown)


include/spectrum.h
```diff
@@ -334,6 +334,10 @@ extern uint  __LIB__    zx_screenstr_callee(uchar row, uchar col) __smallc __z88
 #define zx_setcursorpos(a,b)      fputc_cons(22); fputc_cons(a); fputc_cons(b);
 #define zx_topleft()              fputc_cons(22); fputc_cons(0); fputc_cons(0);
 
+// Print a string on the screen
+extern int printk(char *fmt, ...);
+#define zx_print(s) printk(s)
```
suborb, line 339: This won't work for the case `zx_print("%d %d",12,12);` for example

include/spectrum.h
```diff
@@ -334,6 +334,10 @@ extern uint  __LIB__    zx_screenstr_callee(uchar row, uchar col) __smallc __z88
 #define zx_setcursorpos(a,b)      fputc_cons(22); fputc_cons(a); fputc_cons(b);
 #define zx_topleft()              fputc_cons(22); fputc_cons(0); fputc_cons(0);
 
+// Print a string on the screen
+extern int printk(char *fmt, ...);
```
suborb, line 338: This would need to be `extern int __LIB__ printk(char *fmt)`

I think it's probably better to just do this using a library alias:

```
extern int zx_printf(char *fmt,...)
```

And then in the library create two new files somewhere in libsrc/target/zx/ that looks like this

SCCZ80:
```
PUBLIC zx_printf
EXTERN printk
defc zx_printf = printk
```

ZSDCC:
```
PUBLIC _zx_printf
EXTERN _printk
defc _zx_printf = _printk
```

include/spectrum.h
suborb: An alternative alias would be to puts_cons(x) which takes a single string and just writes it out to the console. It's a lot smaller than printk as well.

include/spectrum.h
Baltasarq: Okay, I modified `spectrum.h` and created `zx_printf.asm` in `libsrc/target/zx`, with the contents for _sccz80_ you wrote above.
But... how do I create the _ZSDCC_ version?

suborb: Just create _zx_printf.asm and copy the ZSDCC version into that.

If you haven't already you'll need to add reference to both files to libsrc/target/zx/spectrum.lst

suborb:
Thank you!

PR #1449, opened 2020-04-19
benchmarks - April 2020

feilipu:
It has been 3 years since the last set of published benchmarks, so let's see what's happened.

 - generally binary sizes haven't changed much, and performance is mostly better.
 - sccz80 can now compile and run `fasta.c` and `n-body.c`.
 - sccz80 has gotten much better, and is much closer to sdcc.
 - some marks are better than HiTech-C (but not for the same binary size).
 - `long` demotion helps (eg `pi_ldiv.c`).
 - `math32` helps float results in some cases.

It would be good to have a review of some the numbers, particularly where they've changed significantly.

__EDIT__ I don't know if there's a summary results table in this repository. I think the tables are in the old place, so there's nowhere to summarise.


libsrc/_DEVELOPMENT/EXAMPLES/benchmarks/fasta/z88dk-new/readme.txt
```diff
@@ -9,11 +9,14 @@ VERIFY CORRECT RESULT
 To verify the correct result compile for the zx spectrum target
```
suborb, line 9: CP/M target

suborb:
Thank you for running these again!

I can believe the sccz80 results have improved: after Alvin ran them last time I took a look at the generated code and added a bunch of alternate generating paths.

I'll do a sanity check of a few of them.

I wonder if it's worthwhile promoting these results to the wiki? - Having them buried in the source code doesn't really aid visibility.

libsrc/_DEVELOPMENT/EXAMPLES/benchmarks/fasta/z88dk-new/readme.txt
feilipu: I used CP/M to test for correctness on real target, and this slipped through. Will fix, along with another typo I noticed elsewhere.

feilipu:
Yes the old place wiki has a summary. That at least should be updated.

But I don't think I have write access to do it, afaik.

Might be also worth promotion of improvement and contrast in news section of wiki.

feilipu:
There's only one result pointing to further work, where sccz80/classic is substantially worse than sccz80/new.  I'll point it out later, if not obvious.

I double checked those results, so unless I have messed up with procedure, there's something amiss.

suborb:
Yeah I spotted that one when glancing through. I’ve lost it now though. 

If it’s fp then the normalisation in genmath is probably the cause.

suborb:
I’d like to migrate off the old wiki. Classic is now off there, I’ve seen you’ve done bits of newlib as well. Just need to figure out what’s left and reorganise it a bit.

suborb:
Ah, it's the n-body one. Yes, that issue is still valid. We had it in the maths benchmarks here: https://github.com/z88dk/z88dk/wiki/Classic--Maths-Libraries#n-body

feilipu:
I’ll try to get the wiki pages sorted shortly. Perhaps even write a bit of a promo highlights text.

PR #1465, opened 2020-05-08
classic - cp/m sequential read/write

feilipu:
Working on adding sequential read and write to the CP/M BDOS `read()` and `write()` functions.

- PIP is the benchmark.
- I've an asm program that can beat PIP (by copying what it does).
- classic lib using the Hitech-C derived functions is slower.

These modifications are to speed up large consecutive reads and writes with no offset.

- The offset calculation is moved outside the while loop because the offset can't change within a call.
- If there is no offset, and the read/write is longer than a sector, then sequential mode is used.
- The function already prepares for random read/write by doing `_putoffset()`, I think.

__EDIT__ a more conservative set of changes produces the same outcome.

### DIFF
--- a/include/cpm.h
+++ b/include/cpm.h
@@ -28,7 +28,7 @@
 /* Size of CPM Sector */
 #define SECSIZE  128
 
-/* Falgs for fcp->use */
+/* Flags for fcp->use */
 #define U_READ  1               /* file open for reading */
 #define U_WRITE 2               /* file open for writing */
 #define U_RDWR  3               /* open for read and write */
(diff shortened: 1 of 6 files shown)


feilipu:
Testing these small changes shows an improvement for the 1MB file copy from 53 seconds to 51 seconds, in direct comparison. I've not found any simple way to do better.

I think there's too much code between the `fread()` call and the eventual `read()` (and also for write case) to approach assembly calls to BDOS (39 seconds), or PIP (42 seconds). It is what it is.

For review.

zx70:
I'm merging it basing on the interesting code you are proposing.
Please take some more time to check how it works with filter programs (e.g. a tool to convert whole a file to uppercase, of similar stuff dealing working on a single byte).
Seek() fpos() and similar functions could also be affected,  but controlling at block level is a priority (hoping it will work on the many targets).

PR #1472, opened 2020-05-14
Fix #1451 - link very slow in Windows

### LINKED ISSUE #1451: (z80asm) Linking very slow on windows
There's a forum thread here: http://specnext.com/forum/viewtopic.php?f=17&t=1667

Which I think shows that linking is extremely slow when using newlib and windows. If my supposition is correct (is it @pauloscustodio?) then fiixing it for windows should make *nix linking even faster.

>Running on a virtualised windows 10 (i5-6260U), the command supplied runs in about 14 seconds.
>
>Compiling with -clib=new (so using sccz80) takes about 9 seconds.
>
>Compiling with zcc +zxn -clib=classic hexdump.c -o hexdump -subtype=dot -create-app  -lesxdos -v - so using the classic library to create a dot file compiles in 3.8 seconds. 
>
>Running on my ancient MacBook (i7-3557U), the times are: 3.8, 1.8 and 0.75 seconds respectively which is altogether much more acceptable.
>
> Looking at the log as it's printing to screen, it appears that most of the time is being spent at the linking stage. What I suspect is happening is that z80asm is searching for library routines by opening the library file, scanning through for the routine, closing the file and doing it again for the next library routine. On *nix this behaviour goes un-noticed, but because (in my limited experience) Windows is horrendously slow at file IO and newlib is decomposed into lots of small routines it becomes really noticeable.

pauloscustodio:
Load object files and libraries to memory and do all searching for symbols in memory instead of opening and closing files for each new symbol to search for.

Fix is not clean because of interference between assembly and linking stages - these should be independent, but are not... maybe fix later...


suborb:
What time improvement did you see?

pauloscustodio:
About half-time linking, see https://github.com/z88dk/z88dk/pull/1472/commits/a70942ad17f46086277994f31e3edbde1d2735f6

PR #1486, opened 2020-06-01
Fix explore addresses and feature colors if interactive debugging

PaintedBlck:
Dear developpers,

please find a first **_pull request for z88dk-ticks_**. It goes without saying that **_this pull request is only a polite proposition_** based on experience and conform with yesterday's question and the received hints of @suborb  

The fixes:
- When asking for a **_memory dump, the addresses in the left column_** where off with -1 starting from the second line on. This has been fixed (I have rewritten the memory dump part of the function cmd_examine.
- When running in interactive mode, with trace on, there were **_two disassembly lines of the same instruction_**. This has been corrected. Please note that this is not corrected for the non-interactive debugging mode (stdin or stdout going through a pipe).

The features:
- When running in non interactive mode, all output is like it was before. Running the test-suits (that use z88dk-ticks to my understanding) gives only positive results, **_what makes me think that the non-interactive layout is unchanged._**
- When running in **_interactive mode, I have changed the output of the reg command_**. I have regrouped and re-aligned, and some parts of the text have a modest color in order to provide an interface that, to my opinion, is faster to use.
- When running in i**_nteractive mode, the help output_** has been colored slightly.
- When running in **_interactive mode, the "prompt" has been changed_** so to leave an extra line between previous output and the new prompt, to permit to search faster where previous results start and stop for each command. The prompt has also been colored and extended in order to form kind of a "line" between previous results and the actual command and results.

I have tested the entire compilation and tests on Linux Blackarch and on Mac OSx. I have also compiled some test programs for the Z80, and everything seems to function on the ticks-front.

For what concerns Windows, I have not been able to create the correct compilation environment to compile the entire z88dk. But I'm pretty sure that (in the correct environment) it will compile and run without problems, since
- Windows 10 TH2 has a command prompt that is compatible with all ANSI escape codes (which I have tested more extensively today under Windows than the limited use I make of escape codes in this proposal for ticks),
- isatty() is deprecated, but still accepted by Windows (in the future, this should be replaced with _isatty() in the Windows version.

I'm looking forward to your comments regarding these little corrections and changes, and I'm planning to invest more time in the z88dk development suite. **_A further proposition would be to tackle the compiler warnings like indent warnings, type-casting warnings_**, ...

Kind regards,
PaintedBlck

### DIFF
--- a/src/ticks/debugger.c
+++ b/src/ticks/debugger.c
@@ -3,6 +3,7 @@
 #include <string.h>
 #include <stdlib.h>
 #include <ctype.h>
+#include <unistd.h>                         // For declarations of isatty()
 
 #include "utlist.h"
 
@@ -11,7 +12,13 @@
 
 #define HISTORY_FILE ".ticks_history.txt"
 
+#define FNT_CLR "\x1B[34m"              // Color (blue on black)
+#define FNT_BLD "\x1B[34;1m"            // Bold color (bright blue on black)
+#define FNT_BCK "\x1B[34;7m"            // Reverse color (white on blue)
+#define FNT_RST "\x1B[0m"               // Reset colors to default value (gray on black)
 
+#define CLR_REG
+#define CLR_ADDR
 
 typedef enum {
     BREAK_PC,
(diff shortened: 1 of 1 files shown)


src/ticks/debugger.c
```diff
@@ -3,6 +3,7 @@
 #include <string.h>
 #include <stdlib.h>
 #include <ctype.h>
+#include <unistd.h>                         // For declarations of isatty()
```
suborb, line 6: For WIN32 we can `#include <io.h>` which defines `isatty()` rather than `unistd.h`. This should work on both mingw and MSVC

src/ticks/debugger.c
```diff
@@ -516,31 +588,37 @@ static int cmd_examine(int argc, char **argv)
...
+
+                if ( i % 16 == 0 )                              // Handle line prefix
+                    if (interact_with_tty)
+                        printf(FNT_CLR"%04X"FNT_RST":   ", addr);
+                    else
+                        printf("%04X:   ", addr);               // Non-color output for non-active tty
+
+                printf("%02X ", b);                             // Hex dump of actual byte
+
+                if (i % 16 == 15) printf("   %s\n", abuf);      // Suffix line with ASCII dump
```
suborb, line 611: I really like this feature, I don't know why I didn't do it originally!

suborb:
I really like these changes, thank you so much - I'll do the `<io.h>` fixup.

image

PaintedBlck:
Thank you so much for the positive comments. There’s more where this came from. I’ll keep in touch with further propositions.

Kind regards

PR #1488, opened 2020-06-02
Fix - Resolved warnings (buffer overflow, uninitialized variables)

PaintedBlck:
Dear development team,

I took the liberty to resolve all **_compiler warnings for 'appmake'_**.

Each warning has been looked into. In case of a real problem, the problem is solved. In case of a non-existing problem (most of the cases), the compiler flag is disabled for the concerning statement by means of a number of #pragma statements around the problem area.

I hope these changes can get your approval. In case they do, I'll continue correcting other compiler warnings.

Kind regards,
PaintedBlck


src/appmake/ace-tap.c
```diff
@@ -128,7 +128,14 @@ int acetap_exec(char* target)
                 strncpy(name, blockname, 10);
             } else {
                 strcpy(name, blockname);
+#if __GNUC__ >= 8
```
suborb, line 131: I have to admit that I don't like this since it only removes the diagnostic for one compiler.

 I think we'd be better off by using a snprintf() format string to right pad with spaces.

src/appmake/appmake.c
```diff
@@ -202,7 +202,7 @@ long parameter_search(const char *filen,const  char *ext,const char *target)
      /* Successfully opened the file so search through it.. */
     while ( fgets(buffer,LINEMAX,fp) != NULL ) {
         if (strncmp(buffer,target,strlen(target)) == 0 && isspace(buffer[strlen(target)])) {
-            sscanf(buffer,"%*s%*s%*[ $]%lx",&val);
+            sscanf(buffer,"%*s%*s%*[ $]%lx", (long unsigned int *) &val );
```
suborb, line 205: I have a feeling this is a hangover from dos/32 bit - I think a better fix is probably `"%" PRIx32 ""`

src/appmake/appmake.c
```diff
@@ -1220,7 +1220,14 @@ void mb_enumerate_banks(FILE *fmap, char *binname, struct banked_memory *memory,
 
                         snprintf(section_name, sizeof(section_name), "%s", &symbol_name[2]);
                         section_name[len - 7] = 0;
+#if __GNUC__ >= 8
```
suborb, line 1223: Probably better to resize the buffer and then we cover all compilers?

src/appmake/ace-tap.c
PaintedBlck: I agree. The point is that I'm not a 'Warning' lover, since warnings are there to prevent problems, and letting them exist gives me an uncomfortable feeling. I like compiles to be crisp.

Thanks for the hint. I'll look into the snprintf() format solution.

Grtz,
PB

src/appmake/appmake.c
PaintedBlck: I'm not sure what you mean because all definitions of PRIx32 that I have found in C implementations are x or lx. Since both x and lx can only scan unsigned numbers, there must be a typecast between &var (being pointer to signed) and the result of the scan (being pointer to unsigned).

Since Z80 works with 16-bit addresses, all 16 bits of a scan with x or lx will be filled with the scanned number. So, if we want to represent an error as -1 in the same var, we need at least 32 bits, or a long type.

So, I would like to propose to stick with a temporary type casting to  (long unsigned int *) since this is cross C version and cross compiler compatible to my knowledge.

Grtz,
PB

suborb: Sorry, that was my confusion, I thought it was a size issue rather than a signedness issue.

PR #1489, opened 2020-06-02
Fix - ticks - memory.c

PaintedBlck:
**Warning solved**:
     zx_banks[] was one element too small for loop in zx_init()

**Solution**:
     Definition changed to zx_banks[8]

**Question**:
     Do you prefer a pull request per module (ticks, appmake, ...) or one pull
     request for a larger number cross-module fixes?

Kind regards,
PB


suborb:
Thanks, nice spot.

I think I prefer one per module otherwise things can get a bit messy. I did a cleanup of a few of the other programs since you gave me the bug!

PR #1490, opened 2020-06-03
Fix - appmake - Solved warnings (buffer overflow, wrong variable type, string padding, ...)

PaintedBlck:
Hi,

I have **_Solved all warning_** of the **_appmake_** module

I have **_taken into account your remarks in the previous pull request for appmake_**.
- some **_buffer sizes have been adjusted_** to prevent warnings and/or buffer overflows
- sign **_warnings have been solved with (long unsigned int *) typecasting_**
- some c files had their execute bit set to '1', since git saves the execute bit, **_I have reset the execute bit of these files to '0'_**.

**_Please check in mz.c (line 805) if padding a filename with ascii CR (\r, \15) characters is really the goal of the original code_**. The changed code sticks to padding with CR, but prevents a buffer overrun on the stack (variable name is on the stack).

So far as I'm concerned, this pull request is ready for a merge in the z88dk master branch.

Grtz,
PB


src/appmake/appmake.c
```diff
@@ -202,7 +202,7 @@ long parameter_search(const char *filen,const  char *ext,const char *target)
      /* Successfully opened the file so search through it.. */
     while ( fgets(buffer,LINEMAX,fp) != NULL ) {
         if (strncmp(buffer,target,strlen(target)) == 0 && isspace(buffer[strlen(target)])) {
-            sscanf(buffer,"%*s%*s%*[ $]%lx",&val);
+            sscanf(buffer,"%*s%*s%*[ $]%lx", (long unsigned int *) &val);
```
PaintedBlck, line 205: See pull request text: sticking with a typecast for the pointer to val seems the most elegant and compatible with different C versions and all compilers.

src/appmake/c7420.c
```diff
@@ -179,8 +176,8 @@ int c7420_exec(char* target)
         /****************/
         /* BASIC loader */
         /****************/
-        sprintf(lsbbuf, "%i", (int)pos % 256); /* no more than 3 characters long */
-        sprintf(msbbuf, "%i", (int)pos / 256); /* no more than 3 characters long */
+        sprintf(lsbbuf, "%i", (uint8_t) (pos % 256) ); /* no more than 3 characters long */
+        sprintf(msbbuf, "%i", (uint8_t) (pos / 256) ); /* no more than 3 characters long */
```
PaintedBlck, line 180: This change in type cast doesn't change anything, since both numbers are at most 3 characters long (values 0..255) which fits exactly in uint8_T.

suborb:
Thank you for your work on this.

I downloaded a mzt from elsewhere and the header is:

```
hexdump -C RAMTEST.mzf
00000000  01 52 41 4d 54 45 53 54  2e 4f 42 4a 0d 0d 0d 0d  |.RAMTEST.OBJ....|
00000010  0d 0d 46 00 00 d0 00 d0  00 00 00 00 00 00 00 00  |..F.............|
```

So, it looks like padding with 13d, 15o is the right thing to do.

src/appmake/mz.c
```diff
@@ -792,28 +792,26 @@ int mz_exec(char *target)
...
-		if (strlen(blockname) > 16 ) {
-			strncpy(name,blockname,16);
-		} else {
-			strcpy(name,blockname);
-			//strncat(name,"\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0",17-strlen(blockname));
-			strncat(name,"\15\15\15\15\15\15\15\15\15\15\15\15\15\15\15\15\15",17-strlen(blockname));
-		}
+
+        // To Do - It seems very weird to pad the filename with CR's (ascii char 15 octal)
+        memset(name, '\r', sizeof(name)-1);
```
suborb, line 805: For true portability this should be '\15' rather than \r to cope with non-ascii host machines.

PR #1498, opened 2020-06-08
Fix-basck - Solved warnings

PaintedBlck:
Dear development team,

this is a pull request for the support **_module basck_**.

The **_following problems were solved_**:

- **_'./build.sh -c' did not erase certain support module directories_**. Cause was the Makefile in the 'support' directory that did not clean the 'basck', 'debugger' and 'rcmx000' directories.
- basck.c: erased **_trailing whitespace_** on a number of lines
- basck.c: **_dangling else_** warning
- basck.c: warning about **_unclear statement 'if (img[res-1] & 0xC7 == 0xC7)'_**. Added parentheses
- basck.c: replaced **_return value of functions dlbl and clbl_** with no return value (void)
- basck.c: warnings due to **_typecasting problems_** in printf statements (%04x needs unsigned int and was given long int values).
- basck.c: **_misleading indent warning_**
        - Solution was to keep the actual program logic and to shift an if block 4 spaces to the left.
        - **_TO DO: check if the actual program logic is correct_**. My opinion is that the 'if block' at line 2641 should contain the 'if block' at line 2642 AND the 'if block' at line 2654, which it doesn't, so it isn't. But I'm far from sure.

I hope these changes can have your approval. Since the misleading indent warning is beyond my actual knowledge of the basck module, **_I propose to merge all my changes in the master branch, and to take the time to verify the program logic starting at line 2641_**. My changes haven't changed the actual program logic.

Kind regards,
PB


suborb:
This is one for @zx70 to check.

PR #1506, opened 2020-06-13
Tiki100 fixing

Frodevan:
Did some tweaking. I have not set up a build-environment, so it's nice if someone can test that all inkludes are ok and that scrolling a line up in stdio works as expected before merging.


suborb:
@Frodevan - thanks. It looks like the build is failing on this one: https://travis-ci.org/github/z88dk/z88dk/builds/697907163 (I'm sure there used to be notifications on PRs to show this, but it's gone missing).

Possibly there's some confusion between `gr_scroll` and `gr_vscroll`?

Frodevan:
Yeah, noticed the typo after you mentioned it. Now it should be right.

suborb:
@Frodevan sorry this isn't working.

Colour isn't working:

image

And this is the result after scrolling one line:

image

suborb:
As a note, if you can compile programs for the tiki100 (and can't build the libraries), you should be able to just copy the files into a local directory and add them to your project, eg:

```
zcc +cpm --generic-console -subtype=tiki100 ansivt52.c generic_console.asm -create-app
```

Should use the symbols defined in the local copy of `generic_console.asm` rather than the version in the library.

Frodevan:
I'll look into getting zcc up and running locally like you mentioned tomorrow, just tried to build the whole z88dk in Mingw32, but there's a whole forest of external dependencies I don't have set up.

In the mean time, I pushed some corrections for the typos that would be involved in producing the faults you mentioned. Some small overlooked details, etc., etc...

suborb:
Building the libraries on windows should be fairly easy these days - I went through the exercise on a clean machine a while back and came up with this recipe: https://github.com/z88dk/z88dk/wiki/Classic-Building-the-libraries

Basically all that's needed is git-bash and Make. 

Building the binaries and all the tests is a bit of a flood of dependencies though - mostly perl ones for the z80asm tests.

I'll test your changes either tonight or tomorrow

libsrc/target/tiki100/gr_setpalette.asm
```diff
@@ -63,39 +65,42 @@ set_loop:
...
-        DI
-        OUT ($14),A             ; Palette register (prepare the color to be loaded)
-        OUT ($14),A             ; Palette register (do it again to be sure)
-	ld	a,(hl)
+
+	PUSH	DE
+	LD	A,E
+	DI
+	OUT	($14),A	     ; Palette register (prepare the color to be loaded)
+	ld	a,(PORT_0C_COPY)
```
suborb, line 82: I feel like I'm missing something - why not just replace the hl constant in the original with PORT_0C_COPY and save a byte on the new code?

libsrc/target/tiki100/stdio/generic_console.asm
```diff
@@ -216,22 +217,21 @@ done:
 
 
 generic_console_scrollup:
+	push	af
```
suborb, line 220: Only bc and de need to be saved here

libsrc/target/tiki100/stdio/generic_console.asm
```diff
@@ -216,22 +217,21 @@ done:
...
+	call	gr_vscroll
 	call	swapgfxbk
-	ld	hl,1024
-	ld	de,0
-	ld	bc, 32768 - 1024
-	ldir
-	ld	h,d
-	ld	l,e
-	inc	de
+	ld	hl,$7C00
```
suborb, line 226: I think this address is wrong?

image

suborb:
I've lifted the scrolling since that works well - quite an impressive speedup!

libsrc/target/tiki100/gr_setpalette.asm
Frodevan: Forgot to add it, will do

Frodevan: The idea was really just to use the constant instead of a value directly, and eventually make the code more readable. When you mention it, using hl does indeed save a byte and a clock-cycle.

Frodevan:
Also, one thing to consider...

ld hl,**
ld de,**

6 bytes, 20+2 clocks

ld hl,**
ld d,h
ld e,l
inc de

6 bytes, 24+4 clocks

Frodevan:
The race-condition fix now follow the exact same order of IO that I use in my own libraries.

suborb:
Working well - thanks. I'll merge once the build goes through.

PR #1508, opened 2020-06-13
(sccz80) _Float16 support

suborb:
Add support to sccz80 for _Float16. This is available as a first class type and follows C11 extension ISO/IEC TS 18661-3:2015 (hopefully).

Take the opportunity to do a bit of cleanup of sccz80, removing dead code and trying to contain codegen code into codegen.c rather than have knowledge all over the place.

This PR also contains @feilipu work on the library needed.

To be done elsewhere:

- Conversion to from bbc, mbf, genmath formats


lib/config/alias.inc
```diff
@@ -18,6 +18,6 @@ ALIAS   --math-bbc              -Cc-fp-mode=z88     -lbbc_math
 #
 # Math library aliases (classic + newlib)
 #
-ALIAS   --math16                -Cc-fp-mode=ieee16  -lmath16 -Cc-D__MATH_MATH16 -D__MATH_MATH16 -pragma-define:CLIB_16BIT_FLOAT=1
```
suborb, line 21: @feilipu I think this can be removed - all that's needed now is -lmath16?

include/sys/types.h
```diff
@@ -29,7 +29,11 @@ typedef double double_t;
 
 #ifndef _HALF_FLOAT_T
 #define _HALF_FLOAT_T
-typedef short half_t;           /* IEEE16 half float type */
+#ifdef __SDCC
```
suborb, line 32: I think this is probably best as `#ifndef SCCZ80` - I'm playing around with VScode and using clang as the IntelliSense engine.

libsrc/_DEVELOPMENT/EXAMPLES/benchmarks/mandelbrot/z88dk-new/mandelbrot.c
```diff
@@ -84,25 +90,26 @@ int main (int argc, char **argv)
...
 
-    for(y=0;y<h;++y) 
+    for(y=0;y<h;++y)
     {
         for(x=0;x<w;++x)
         {
             Zr = Zi = Tr = Ti = 0.0;
-            Cr = (2.0*x/w - 1.5); Ci=(2.0*y/h - 1.0);
-        
+            Cr = (2.0*x/w - (DOUBLE)1.5);
```
suborb, line 98: This case is no longer needed

suborb:
Merging this one since there's more work to come but at the moment it is usable.

lib/config/alias.inc
feilipu: I'd like to leave the alias for consistency, but I agree all the definitions can depart.

include/sys/types.h
feilipu: Done. Added `f32_t` and `f16_t`, to match the integer shortcuts. Not sure what to do about 48 and 64 bit math, though. Another day.

libsrc/_DEVELOPMENT/EXAMPLES/benchmarks/mandelbrot/z88dk-new/mandelbrot.c
feilipu: Done.

lib/config/alias.inc
feilipu: Actually, it is nice to be able to choose whether code should use the normal path or the 16 bit path, based on whether the library is included. Will keep the definitions, unless there's an issue?

include/sys/types.h
feilipu: Took the `f32_t` and `f16_t` shortcuts back out. `f32_t` can't work consistently.

PR #1509, opened 2020-06-15
Fix - Appmake - Travis build - fread return code warnings

PaintedBlck:
As demanded by @suborb for module appmake

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,6 +34,7 @@ matrix:
 #        - MATRIX_EVAL="CC=clang && CXX=clang++"
 
 before_install:
+  - git checkout test
   - eval "${MATRIX_EVAL}"
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew update                  ; fi
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew install cpanm dos2unix ccache ragel ; fi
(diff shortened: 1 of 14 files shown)


PaintedBlck:
There seems to be a problem with the tests that I have to look into. But I wanted to give the signal that I'm busy changing the appmake module withing the frame of the warnings generated by the travis build.

suborb:
Demanded is a bit strong! More a suggestion if you were interested 😀

It's really appreciated - these things compile so quickly and in general I don't often compile them so I've never caught the itch.

PaintedBlck:
It's probably more my failing language of Shakespeare than the fact that I wanted to imply pressure from your side ;-) 

All 'fread' statements are handled for the **_module appmake_**, except for the **_one at the bottom of 'mz2500.c' that will need a closer look_**. I can do that one also, but that will be for later. First, I'd like to do the easy ones in the other modules. They give a better return on investment :-)

**_The actual changes have been tested with a travis build that succeeded_**.

'PB assistance' for all your itches, big and small...

Grtz,
PB

suborb:
Thank you, I'll merge and then tweak to use `myexit()` so we log with an error message.

suborb:
Argh, that was silly of me - the wrong parameter is being passed to fclose() in places!

PR #1511, opened 2020-06-17
zsdcc - sdcc peep.c patches

feilipu:
Many of the peephole patches contained in r9958 were omitted from the update to r11556.
This attempts to correct that issue, raised #1147.

I've tried to migrate the r9958 patches to r11556 sensibly. Verification is requested.

The biggest change is to revert to using the `argContPrec()` function to identify arguments. 

Some further relevant changes in r11556 code over r9958.

- The definition for `ISINST(l, i)` has been improved, so that the `\t` characters can be omitted, and ordering of matching is not required. `#define ISINST(l, i) (!STRNCASECMP((l), (i), sizeof(i) - 1) && (!(l)[sizeof(i) - 1] || isspace((unsigned char)((l)[sizeof(i) - 1]))))`
- The eZ80 is now supported by sdcc, so code had to be reorganised to suit that CPU.
- There was an issue that some eZ80 instructions were applied to the Z180 (which don't exist there), and a few others were missing. These have been added.


feilipu:
I've been trying to test this PR by using `-SO2` compilation options, but it is having no effect. I'm not seeing the sdcc peep hole rules being used.

Is there a way to generate guarantee that the sdcc `peep.c` peephole will be used, for comparison with the zsdcc peephole outcomes?

feilipu:
There have been substantial improvements to the `z80/peep.c` in sdcc through to r11690. The initial patch set will no longer work (or possibly even be relevant).

Need to redraft this PR to match the current state of sdcc.

feilipu:
Starting to think the best way to resolve this is to break any remaining pieces into feature requests for sdcc upstream.

feilipu:
The r9958 patch set remains available, but the work remains to isolate issues from that set that are not addressed by current r12419 (and later) builds.

Closing this because it is hanging around. I can be reopened if needed.

PR #1516, opened 2020-07-02
Fix copt - Travis build - Unused functions, uninitialized variables

PaintedBlck:
This patch solves all Travis build warnings for the copt module.


suborb:
Thanks - I'm always in two minds what to do with imported sources: leave or modify. Given that these regex files haven't had an update in years it looks like modify is the best option!

PR #1517, opened 2020-07-02
Fix sccz80 - Travis build - Variable set but not used

PaintedBlck:
This patch solves all Travis build warnings for the sccz80 module.


suborb:
Sorry, that one crept back in during the _Float16 work

PR #1519, opened 2020-07-03
Fix ticks - Travis build - Ignored return value

PaintedBlck:
Solved fread return value unused warnings. A Travis build gives no further errors, nor does the build on Linux.

To do this, I needed the exit_log function in ticks.c. I have inserted the code for this function in the file ticks.c just after the #defines and global data definitions. The used code is the same is the code used in appmake.c

Maybe, it would be better to split off this kind of functions in a common source. To be considered.


src/ticks/ticks.c
```diff
@@ -582,14 +583,34 @@ static const uint8_t mirror_table[] = {
     0x3, 0xB, 0x7, 0xF   /* 12-15 */
 };
 
+void exit_log(int code, char *fmt, ...)
+{
+    va_list  ap;
+    
+    char *program_name = (strrchr(__FILE__, '/') ? strrchr(__FILE__, '/') + 1 : __FILE__);
```
suborb, line 590: This will report the program_name as `ticks.c`, on windows (if compiled with MSVC) it would report the full path to the file, for example: `c:\z88dk\src\ticks\ticks.c`.

Given that, it might be best just to hardcode "ticks" as program_name.

PaintedBlck: Indeed, we'll go hardcore ;-)

I'll change and do an other pull request.

PaintedBlck: Promblem with already applied changes in master repo should be solved.

PR #1520, opened 2020-07-04
Fix appmake - Travis build - Ignored return value, potential buffer overflow

PaintedBlck:
I have first _**re-pulled in the up-to-date z88dk/z88dk/master**_ in order not to show false changes that are already in the repo. This will hopefully solve the problem of already changed lines.

This patch _**solves all Travis build warnings for the appmake module**_.

The fread warning (copy bin to dsk loop in mz2500.c) was an indication that there was no error checking in this loop. The rest of the file contained very stringent error checking. So, I have rewritten the loop at the end of the file to have error checking combined with using the return value of fread.

I have also _**swapped the size of the items to fread and the number of items to fread**_, in order to achieve correct error checking.

The **_functionality has not changed_** (I have checked the results before and after the changes).


src/appmake/cpc.c
```diff
@@ -394,7 +394,7 @@ int cpc_exec(char* target)
            * exit_log(1,"Couldn't determine size of file\n"); }
            * len=ftell(source); fseek(source,0,SEEK_SET);
            */
-        if (1 != fread(srchead, 128, 1, source)) { fclose(source); exit_log(1, "Could not required data from <%s>\n",filename); }
+        if (1 != fread(srchead, 128, 1, source)) { fclose(source); exit_log(1, "Could not read required data from <%s>\n",filename); }
```
suborb, line 397: @PaintedBlck thank you, that was a silly copy paste error *1000 by me.

PaintedBlck: ;-)

PR #1521, opened 2020-07-04
Urgent fix - build.sh - Impossible to use parameters -b nor -l

PaintedBlck:
Hi,

launching `./build.sh -c -b -l` gives as a result that -b is an invalid option. Likewise _**'`./build.sh -c -b -l`' gives as result that -l (small L) is an invalid option**_.

Besides above problems, _**I still see that -nb and -nl are usable parameters**_: `Usage: ./build.sh [-e][-k][-c][-t][-nb][-nl]`.

```
./build.sh: illegal option -- b
Usage: ./build.sh [-e][-k][-c][-t][-nb][-nl]

-k\tKeep building ignoring errors
-c\tClean before building
-t\tRun tests
-e\tBuild examples
-p\tTARGET Build specified targets
-b\tDon't build binaries
-l\tDon't build libraries

Default is to build binaries and libraries
```
I have _**fixed and changed build.sh**_ so to be _**backward compatible with the idea behind the one in z88dk/s88dk/master**_. But _**besides the urgent fix**_, I have also:
- added some logic to _**prevent strange cases**_ to happen: `./build.sh -b -l`
- added some logic (_**extra option: '-C'**_) that also _**cleans the bin directory**_
- added some logic to _**prevent adding the bin directory multiple times to $PATH**_
- _**changed the '\t' in the help text**_ so to have a nicer layout on the console (see below)

_**build.sh has been thoroughly tested**_ with a large number of parameter combinations (even '-c -C') and seems to behave correctly. The Travis build _**gives the same results as the one for the actual master branch**_.

The proposed help screen is:
```
Usage: ./build.sh [-b][-c][-C][-e][-h][-k][-l][-p][-t]

  -b    Don't build binaries
  -c    Clean build environment
  -C    Clean build environment and binaries (including bin/*)
        Chosing this option makes a manual rebuild of zsdcc and
        szdcpp necessary in the win32 environment
  -e    Build examples
  -h    This help information
  -k    Keep building ignoring errors
  -l    Don't build libraries
  -p    TARGET Build specified targets
  -t    Run tests

Default is to build binaries and libraries
```

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -4,97 +4,157 @@
 # Build z88dk on unix systems
 #
 
+function show_help_and_exit {
+
+  if [[ -n $1 ]]; then rc=$1; else rc=0; fi
+
+  echo ""
+  echo "Usage: $0 [-b][-c][-C][-e][-h][-k][-l][-p][-t]"
+  echo ""
+  echo "  -b    Don't build binaries"
+  echo "  -c    Clean build environment"
+  echo "  -C    Clean build environment and binaries (including bin/*)"
+  echo "        Chosing this option makes a manual rebuild of zsdcc and"
+  echo "        szdcpp necessary in the win32 environment"
+  echo "  -e    Build examples"
+  echo "  -h    This help information"
+  echo "  -k    Keep building ignoring errors"
+  echo "  -l    Don't build libraries"
+  echo "  -p    TARGET Build specified targets"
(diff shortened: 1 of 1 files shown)


suborb:
Oh, that was dumb of me (it's not been a great week to say the least). Thank you once again.

PR #1523, opened 2020-07-05
Fix zcc - Travis build - Invalid ISO C90 comment format, unused return values

PaintedBlck:
Hi,

Just continuing to _**fix the usual warnings for module zcc**_.

Note that also _**src/common/dirname.c had to be changed**_ due to the ISO C90 compiler setting: only two comments put in 'old' style /* */, so _**no danger of causing problems in other modules**_.

Remark: since the function exit_log is not used in zcc.c by the author, _**I have adapted my changes to the style of the rest of the zcc.c source**_ (fprintf followed by exit)

Grtz.


suborb:
Thank you once more

Issue #1543, opened 2020-07-26
newlib - maths library using Am9511A 


feilipu:
Feeling inspired by this whole Am9511A maths thing for the DAI, I laid out this Am9511A Module for the RC2014 today.

![Am9511A Module](https://user-images.githubusercontent.com/3955592/88478829-9f1cec80-cf8e-11ea-9fbb-6e87d249a01c.png)

The Module supports using interrupt mode, with the Z80 NMI. But since the NMI jump address `0x66` clashes with CP/M it is not really useful, and I don't plan to implement interrupt access initially.

The library will use the same linkage as math32, but the back end will be supported by either "demand wait" or "polling" mode access to the Am9511A. Bindings to IEEE-754 32-bit floating point will be provided, so that the use of the APU is transparent.

suborb:
Nice idea hiding it behind math32 format - if we can create 8080 variants of the converters/utility routines we should be able to get it running on the dai as well.

feilipu:
Stretch goal.
As it is modular, multi-APU config should be possible (but totally impractical).

suborb:
FWIW, I've added an AM9511a emulator taken from an Enterprise128 emulator.

I've not hooked it in, it fudges the clock cycles but it should make testing this library a bit easier when it's done.

feilipu:
> FWIW, I've added an AM9511a emulator into ticks.

Looks very useful.

The emulator uses a [fixed multiplier](https://github.com/z88dk/z88dk/blob/master/src/ticks/am9511.c#L612) for APU:CPU ticks. It would be a useful extension to make the ticks multiplier configurable, depending on the F_CPU and implementation. The APU Module above has a 1:3 multiplier off a 7.3MHz CPU. I guess the DAI is 1:1 off a 2MHz CPU. The YAZ180 is 1:16 FWIW.

I'll need to read the command register for the result flags (zero, overflow, NaN, etc) as part of the result unloading and proper conversion to IEEE-754. That was omitted from the [current routines](https://github.com/z88dk/z88dk/blob/master/libsrc/_DEVELOPMENT/math/float/math32/z80/f32_fam9511.asm), because of being lazy. And the `_APU_F_BUSY` would be important to have available to break out of a "demand wait" or "polling" loop. But, I guess that for the emulator adding the correct number of ticks and returning not busy flag would be nearly the same as implementing a demand wait? Predicting what the CPU Could be doing during 7,000 ticks of polling is anyone's guess.

[...]

PR #1547, opened 2020-08-05
(am9511) Am9511 APU Module

feilipu:
Support for the AMD Am9511A APU, in the form of a Module for the RC2014, and for the DAI.
Will also support YAZ180, in due course.

Issue #1543

Designed to look like IEEE-754, so differences in hardware floating point format are masked.


suborb:
I really like the idea of using IEEE as the user/compiler data type for 32 bit maths libraries - it allows sdcc to get in on supporting different maths libraries fun.

Regarding this PR, the DAI uses a memory mapped AM9511 and is also an 8080 so I think will need a different library to this one - probably a variant of daimath32 delegating to the ROM calls.

feilipu:
> Regarding this PR, the DAI uses a memory mapped AM9511.

I couldn't find any reference to the DAI Am9511A interface. Is there a URL somewhere?
But in general, memory mapped should be pretty easy to manage, just a different 16-bit I/O address (and I've tried to be 16-bit compliant already).

> and is also an 8080 so I think will need a different library to this one.

Tried initially to be 8080 compliant, but there were then too many things going on to make progress. Something to come back to once z80 is working. Most of the code is in the `push` `pop` (format conversion) routines. So it will be just be rewriting them.

 - probably a variant of daimath32 delegating to the ROM calls.

Yes. that's a good idea.

suborb:
> I couldn't find any reference to the DAI Am9511A interface. Is there a URL somewhere?

Just this disassembly: https://electrickery.hosting.philpem.me.uk/comp/dai/doc/FW_1E0-1EF.pdf starting around address $ECCC - it uses (memory) addresses $fb00, $fb01

> But in general, memory mapped should be pretty easy to manage, just a different 16-bit I/O address (and I've tried to be 16-bit compliant already).

I did a little trick on TMS9918 library to handle this sort of thing, eg: https://github.com/z88dk/z88dk/blob/master/libsrc/video/tms9918/gen_set_vdp_reg.asm#L31 However that only works if the code is being compiled for each target which wouldn't be the case with maths libraries.

> Tried initially to be 8080 compliant, but there were then too many things going on to make progress. Something to come back to once z80 is working. Most of the code is in the `push` `pop` (format conversion) routines. So it will be just be rewriting them.

8080 is surprisingly tough at times, a bit of a paradigm shift from dealing with registers to dealing with memory buffers.

feilipu:
>> But in general, memory mapped should be pretty easy to manage, just a different 16-bit I/O address (and I've tried to be 16-bit compliant already).

> I did a little trick on TMS9918 library to handle this sort of thing, eg: https://github.com/z88dk/z88dk/blob/master/libsrc/video/tms9918/gen_set_vdp_reg.asm#L31
However that only works if the code is being compiled for each target which wouldn't be the case with maths libraries.

I was thinking that the `pushf`& `popf` routines to load and unload data could be stored in the `target/device/am9511` and be specific for each target that had the APU.

The problem with that idea is that the APU commands are done now inline. I'd have to move the command loading to a target specific routine too. But, not a great deal of complexity to do that.

feilipu:
Whetstone results compared to `math32`.
`math32`: 91sec.
`am9511`: 28sec.
```sh
B0% whet9511

N=      0 J=      0 K=      0
X1=  1.0000e+00 X2= -1.0000e+00
X3= -1.0000e+00 X4= -1.0000e+00

N=    120 J=    140 K=    120
X1= -6.8343e-02 X2= -4.6263e-01
X3= -7.2971e-01 X4= -1.1240e+00

N=    140 J=    120 K=    120
X1= -5.5320e-02 X2= -4.4739e-01
X3= -7.1092e-01 X4= -1.1030e+00

N=   3450 J=      1 K=      1
X1=  1.0000e+00 X2= -1.0000e+00
X3= -1.0000e+00 X4= -1.0000e+00

N=   2100 J=      1 K=      2
X1=  6.0000e+00 X2=  6.0000e+00
X3= -7.1092e-01 X4= -1.1030e+00

N=    320 J=      1 K=      2
X1=  4.9039e-01 X2=  4.9039e-01
X3=  4.9037e-01 X4=  4.9037e-01

N=   8990 J=      1 K=      2
X1=  1.0000e+00 X2=  1.0000e+00
X3=  9.9994e-01 X4=  9.9994e-01

N=   6160 J=      1 K=      2
X1=  3.0000e+00 X2=  2.0000e+00
X3=  3.0000e+00 X4= -1.1030e+00

N=      0 J=      2 K=      3
X1=  1.0000e+00 X2= -1.0000e+00
X3= -1.0000e+00 X4= -1.0000e+00

N=    930 J=      2 K=      3
X1=  8.3469e-01 X2=  8.3469e-01
X3=  8.3469e-01 X4=  8.3469e-01


B0% whetstn

N=      0 J=      0 K=      0
X1=  1.0000e+00 X2= -1.0000e+00
X3= -1.0000e+00 X4= -1.0000e+00

N=    120 J=    140 K=    120
X1= -6.8344e-02 X2= -4.6263e-01
X3= -7.2971e-01 X4= -1.1240e+00

N=    140 J=    120 K=    120
X1= -5.5314e-02 X2= -4.4744e-01
X3= -7.1096e-01 X4= -1.1031e+00

N=   3450 J=      1 K=      1
X1=  1.0000e+00 X2= -1.0000e+00
X3= -1.0000e+00 X4= -1.0000e+00

N=   2100 J=      1 K=      2
X1=  6.0000e+00 X2=  6.0000e+00
X3= -7.1096e-01 X4= -1.1031e+00

N=    320 J=      1 K=      2
X1=  4.9115e-01 X2=  4.9115e-01
X3=  4.9114e-01 X4=  4.9114e-01

N=   8990 J=      1 K=      2
X1=  1.0000e+00 X2=  1.0000e+00
X3=  9.9994e-01 X4=  9.9994e-01

N=   6160 J=      1 K=      2
X1=  3.0000e+00 X2=  2.0000e+00
X3=  3.0000e+00 X4= -1.1031e+00

N=      0 J=      2 K=      3
X1=  1.0000e+00 X2= -1.0000e+00
X3= -1.0000e+00 X4= -1.0000e+00

N=    930 J=      2 K=      3
X1=  8.3463e-01 X2=  8.3463e-01
X3=  8.3463e-01 X4=  8.3463e-01
```

suborb:
Am I correct in thinking that if we define and export `__IO_APU_CONTROL`, `__IO_APU_DATA` and the other ports in `lib/target/test/classic/test_crt0.asm` we'll be able to use from ticks? Or do we need to define the AM9511 operations as well?

feilipu:
I think the APU operations will need to be defined as well. Each operation consists of loading a command Byte into the `__IO_APU_CONTROL` register, then magic happens, and the result can be popped off when the magic is complete. So, somehow ticks would have to know what magic needed to be done (as per the emulation you've added).

To be brutally honest, the only novelty in this library is the `pushx` and `popx` routines for the APU. Everything else is just scammed from `math32`. Which means that the library can be pretty easily massaged to support different access methods (like the DAI, and YAZ180).

An extension to this is to do the `long` integer arithmetic with the APU too. I'll get around to that shortly. Should also make a substantial improvement for the RC2014 and the DAI (especially).

suborb:
I'll have a fiddle with +test/ticks I've finished the longlong stuff - I'm almost there, however I've managed to break the math32 and math16 tests (division by 2) when I rearranged the classic crt0 code. It's going to be fun discovering what house of cards I disturbed there.

> An extension to this is to do the long integer arithmetic with the APU too.

I did try to use the (8080) DAI routines for the 8080 arithmetic but things broke in a mysterious way so I left it for later investigation - but APU acceleration would be much better.

>  Should also make a substantial improvement for the RC2014 and the DAI (especially).

Sadly Mame doesn't emulate the AM9511 so that will be a real hardware test only.

PR #1554, opened 2020-08-15
Automation of Dhrystone measurements

ttsiodras:
These two Makefiles allow for very easy computation of the Dhrystone
results:

    $ cd z88dk/libsrc/_DEVELOPMENT/EXAMPLES/benchmarks/
    $ cd dhrystone21/z88dk-classic
    $ make

    [-] Compiling Dhrystone code...
    [-] Measuring T-states (cycles):
    [-] From 014b to 02b4
    [-] Computing Dhrystones...
    [-] DMIPS = .18446

    $ make clean
    $ cd ../z88dk-new
    $ make

    [-] Compiling Dhrystone code...
    [-] Measuring T-states (cycles):
    [-] From 01c0 to 0345
    [-] Computing Dhrystones...
    [-] DMIPS = .17738

    $ make clean
    $ git status
    $
    $ # All clean

In addition to z88dk itself, the only dependencies used by the Makefiles
are "cat", "bash", "tr", "awk" and "bc" - traditional UNIX tools that exist
out of the box in all modern distributions.

### DIFF
--- a/libsrc/_DEVELOPMENT/EXAMPLES/benchmarks/dhrystone21/z88dk-classic/Makefile
+++ b/libsrc/_DEVELOPMENT/EXAMPLES/benchmarks/dhrystone21/z88dk-classic/Makefile
@@ -1,0 +1,27 @@
+SRC:=$(wildcard *.[ch])
+BIN:=dhry.bin
+
+Q=@
+ifeq ($V,1)
+Q=
+endif
+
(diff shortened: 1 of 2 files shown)


suborb:
Thank you - I like the ideas behind this.

When I get a chance I'll extend and make it usable for the other benchmarks we run.

PR #1563, opened 2020-09-04
fix rs232 if1 implementation for zx

desertkun:
Current implementation of `rs232if1.lib` does not work at all, calling `rs232_init` or `rs232_params` leads to crash. Tested this on 48k and 128k spectrum on FUSE emulator (do not have real hardware at this time).

This is my very first assembly code, I basically moved things over from `rs232plus.lib`, what I gathered is current implementation of `rs232if1` is a leftover of disassembly with little thought into it. I also put little thought in actual timing, since FUSE's implantation simply responds to read events. But because I haven't touched the timing part of the old assembly, it should work fine. Hope there is anyone around with real speccy that can try this.

Here are "proofs" this fix works with my custom FUSE peripheral that handles port `0xF7`
image

Code to test this:
```C
#include <stdio.h>
#include <z88.h>
#include <rs232.h>

int main()
{

    printf("\x0C"); // clear screen
    printf("Hello, world!\n");
    if (rs232_init() != RS_ERR_OK) {
        printf("Error\n");
        return(2);
    }
    printf("RS232 init OK\n");
    if (rs232_params(RS_BAUD_9600, RS_PAR_NONE) != RS_ERR_OK) {
        printf("Error\n");
        return(2);
    }

    printf("RS232 params OK\n");
    uint8_t test = 0x00;
    if (rs232_get(&test) == RS_ERR_OK)
    {
        printf("GET 1 OK: 0x%1x\n", test);
    }
    if (rs232_get(&test) == RS_ERR_OK)
    {
        printf("GET 2 OK: 0x%1x\n", test);
    }

    rs232_put(0xAA);
    rs232_put(0xAA);
    rs232_put(0xAA);
    rs232_put(0xAA);
    rs232_put(0xAA);

    printf("SEND OK\n");

    return 0;
}
```

Here's my custom FUST port handler capturing this:
image

### DIFF
--- a/libsrc/target/zx/rs232/if1/rs232_init.c
+++ b/libsrc/target/zx/rs232/if1/rs232_init.c
@@ -17,12 +17,16 @@
 u8_t rs232_init()
 {
 #asm
-        rst     8
-        defb    $31             ; Create the IF1 system variables area
-        ;xor     a
-        ;ld      ($5cc7),a       ; Reset SER-FL to clean the input buffer
-        ;ld      a,(23624)
-        ;ld      (23750),a       ; Set IOBORD to the current border colour (hide flashing)
+    PUBLIC BAUD
+    PUBLIC SERFL
+
+    BAUD:
+        ; Default speed: 2400 baud
+        defw $36
+    SERFL:
+        ; flag + data byte about an eventual 2nd character already transmitted
+        defw 0
+
         ld      hl,RS_ERR_OK
 #endasm
 }
--- a/libsrc/target/zx/rs232/if1/rs232_get.c
+++ b/libsrc/target/zx/rs232/if1/rs232_get.c
@@ -20,6 +20,9 @@
 uint8_t rs232_get(uint8_t *char) __naked __z88dk_fastcall
 {
 #asm
+        EXTERN BAUD
+        EXTERN SERFL
+
         push	hl
 ;	rst	8
 ;	defb	$1d  		; Calls the Hook Code here
@@ -41,7 +44,7 @@ uint8_t rs232_get(uint8_t *char) __naked __z88dk_fastcall
 ; (Hook Code: $1D)
 
 ;; BCHAN-IN
-L0B88:  LD      HL,$5CC7        ; sv SER_FL
+L0B88:  LD      HL, SERFL        ; sv SER_FL
         LD      A,(HL)          ; Is the second-character received flag set?
         AND     A               ;
         JR      Z,L0B95         ; forward to REC-BYTE (if Z set)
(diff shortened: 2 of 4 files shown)


suborb:
Originally this code just used hook codes to talk to the serial port. Then in https://github.com/z88dk/z88dk/pull/1010 the get/put were replaced and reported as working.

After that PR it was left as half-ROM, half-user code, this PR looks like we're now fully user code which given the variance in if1 ROMs has to be a good thing.

libsrc/target/zx/rs232/if1/rs232_init.c
```diff
@@ -17,12 +17,16 @@
 u8_t rs232_init()
 {
 #asm
-        rst     8
-        defb    $31             ; Create the IF1 system variables area
-        ;xor     a
-        ;ld      ($5cc7),a       ; Reset SER-FL to clean the input buffer
-        ;ld      a,(23624)
-        ;ld      (23750),a       ; Set IOBORD to the current border colour (hide flashing)
+    PUBLIC BAUD
```
suborb, line 20: This looks like a call to `rs232_init()` will execute `nop ; ld (hl),0 ; nop` which is not really desirable!

To make it ROMable (as well), it can be rearranged:

```
SECTION data_clib

BAUD:  defw $36
SERFL: defw 0

SECTION code_clib
   ld hl,RS_ERR_OK
```

desertkun: Thanks, like I said, new to this. Will update the PR.

desertkun:
> After that PR it was left as half-ROM, half-user code, this PR looks like we're now fully user code which given the variance in if1 ROMs has to be a good thing.

Could you please elaborate on that? Is that relevant to those magical addresses mean that I've replaced? Thanks.

suborb:
> > After that PR it was left as half-ROM, half-user code, this PR looks like we're now fully user code which given the variance in if1 ROMs has to be a good thing.
> 
> Could you please elaborate on that? Is that relevant to those magical addresses mean that I've replaced? Thanks.

Yes, the magic numbers were system variable addresses - see p45 of this: https://www.sinclair.hu/speccyalista/konyvtar/kezikonyvek/ZXInterface1_Microdrive_Manual.pdf

desertkun:
I guess I need to further investigate this, I never actually had an interface1 cartridge, did not even realize it's a separate cartridge with its own ROM just up until that moment.  Thanks for your time. I'll reopen this if anything.

suborb:
No worries, I'm still keen on merging this since the current ROM dependency I think just sets up the variables, though I'd need to check the disassembly (https://www.tablix.org/~avian/spectrum/rom/if1_2.htm) to confirm.

desertkun:
@suborb I have found the IF1 shadow rom and plugged it in, and fed appropriate `--no-rs232-handshake --rs232-tx <file> --rs232-rx <file>` to fuse, and can confirm that the original code works as well, so I'd rather make no changes in things I do not understand all that much.

PR #1566, opened 2020-09-08
fix sizeof(“concat" “strings”)

desertkun:
Existing sizeof cannot handle C-style `"concatenated" "strings"`, fails like so:
```
src/main:4:26: error: Missing token, expecting ) got "
src/main:4:26: error: Missing token, expecting ; got "
src/main:4:29: fatal error: Expected ';'
Compilation aborted
```

Caught this trying to compile [mbedtls](https://github.com/ARMmbed/mbedtls/blob/e8582ba0f3858d59ad7321ae56c95e2f4e05e657/include/mbedtls/asn1.h#L121).

I also added a regression.

### DIFF
--- a/src/sccz80/const.c
+++ b/src/sccz80/const.c
@@ -584,10 +584,19 @@ void size_of(LVALUE* lval)
         }
     } else if (cmatch('"')) { /* Check size of string */
         length = 1; /* Always at least one */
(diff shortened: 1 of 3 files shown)


suborb:
Thank you for this.

PR #1569, opened 2020-09-17
proposal: add __z88dk_shortcall_hl declspec

desertkun:
I am trying to implement couple of modules for [spectranet](http://spectrum.alioth.net/doc/index.php/Spectranet:_Tutorial_7#Installing_ROM_modules_into_the_Spectranet.27s_flash) which have the following calling convention:
```asm
LD HL, 0xFE01
RST 0x28
```
which I almost like `__z88dk_shortcall` but the latter does this instead
```asm
RST 0x28
DEFW 0xFE01
```
which has it uses cases, but not good for me. This change adds `__z88dk_shortcall_hl` declspec, which has exact parameters (rst, addr), but instead of defining a word it loads the addr into `HL` register.

With this declspec I could declare such a function
```
static void a_function(char a) __z88dk_shortcall_hl(0x28, 0xAB01);
```
And have the corresponding assembly ready inside of the internals of spectranet internals, and C will do the the rst required.

I also made sure that `__z88dk_shortcall_hl` and `__z88dk_fastcall` are impossible to have togethere since they both use HL register.

I have added a regression for this,  and if merged, will update the [wiki](https://github.com/z88dk/z88dk/wiki/CallingConventions#far-convention-modifiers). I understand that this use case is fairly unique, and that's fine I could keep my copy around, but maybe someone else could ever need this.

### DIFF
--- a/src/sccz80/codegen.c
+++ b/src/sccz80/codegen.c
@@ -1116,12 +1116,31 @@ void gen_call(int arg_count, const char *name, SYMBOL *sym)
 
 void gen_shortcall(Type *functype, int rst, int value) 
 {
-    if (functype->return_type->kind == KIND_LONGLONG) {
-        ol("ld\tbc,__i64_acc");
-        push("bc");
+    if ((functype->flags & SHORTCALL_HL) == SHORTCALL_HL) {
+        if ((functype->flags & FASTCALL) == FASTCALL) {
+            // preserve HL from FASTCALL into BC because hl is going to erase it
+            outfmt("\tld bc,\thl\n");
+        }
+        vconst(value);
+        outfmt("\trst\t%d\n", rst);
+    } else {
+        if (functype->return_type->kind == KIND_LONGLONG) {
+            ol("ld\tbc,__i64_acc");
+            push("bc");
+        }
+        outfmt("\trst\t%d\n",rst);
+        outfmt("\t%s\t%d\n", value < 0x100 ? "defb" : "defw", value);
     }
-    outfmt("\trst\t%d\n",rst);
-    outfmt("\t%s\t%d\n", value < 0x100 ? "defb" : "defw", value);
+}
+
+void gen_hl_call(Type *functype, int module, int address)
+{
+    if ((functype->flags & FASTCALL) == FASTCALL) {
+        // preserve HL from FASTCALL into BC because hl is going to erase it
+        outfmt("\tld bc,\thl\n");
+    }
+    vconst(module);
+    outfmt("\tcall\t%d\n", address);
 }
 
 void gen_bankedcall(SYMBOL *sym)
--- a/src/sccz80/declparse.c
+++ b/src/sccz80/declparse.c
@@ -597,6 +597,7 @@ static void parse_trailing_modifiers(Type *type)
     if ( c_use_r2l_calling_convention == NO ) {
         type->flags |= SMALLC;
     }
+    char shortcall_hl = 0;
     while (1) {
         if (amatch("__z88dk_fastcall") || amatch("__FASTCALL__")) {
             if( type->parameters && array_len(type->parameters) != 1 ) {
@@ -627,6 +628,31 @@ static void parse_trailing_modifiers(Type *type)
             continue;
         } else if ( amatch("__banked")) {
             type->flags |= BANKED;
+        } else if ( amatch("__z88dk_hl_call")) {
+            double module, addr;
+            Kind  valtype;
+
+            needchar('(');
+            if (constexpr(&module, &valtype, 0) == 0 ) {
+                errorfmt("Expecting a module address",1);
+            } else {
+                if ( module < 0 || module > 65535 ) {
+                    errorfmt("Module address value is out of range (%x)",1, (int)module);
+                }
+                needchar(',');
+                if ( constexpr(&addr,&valtype,0) == 0 ) {
+                    errorfmt("Expecting a call address",1);
+                } else {
+                    if ( addr < 0 || addr > 65535 ) {
+                        errorfmt("Short call value is out of range (%x)",1, (int)addr);
+                    }
+                    type->flags |= HL_CALL;
+                    type->funcattrs.hlcall_module = module;
+                    type->funcattrs.hlcall_addr = addr;
+                }
+            }
+            needchar(')');
+
         } else if ( amatch("__nonbanked")) {
             type->flags &= ~BANKED;
         } else if ( amatch("__z88dk_sdccdecl")) {
@@ -644,7 +670,8 @@ static void parse_trailing_modifiers(Type *type)
             }
             type->funcattrs.params_offset = val;
             needchar(')');
-        } else if ( amatch("__z88dk_shortcall")) { // __z88dk_shortcall(rstnumber, value)
+        } else if ( amatch("__z88dk_shortcall") || (shortcall_hl = amatch("__z88dk_shortcall_hl"))) { // __z88dk_shortcall(rstnumber, value)
+            uint32_t hl = shortcall_hl ? SHORTCALL_HL : 0;
             double   rstnumber,val;
             Kind  valtype;
 
@@ -662,12 +689,14 @@ static void parse_trailing_modifiers(Type *type)
                     if ( val < 0 || val > 65535 ) {
                         errorfmt("Short call value is out of range (%x)",1, (int)val);
                     }
-                    type->flags |= SHORTCALL;
+                    type->flags |= SHORTCALL | hl;
                     type->funcattrs.shortcall_rst = rstnumber;
                     type->funcattrs.shortcall_value = val;
                 }
             }
             needchar(')');
+        } else if ( amatch("__z88dk_shortcall_hl")) {
+            type->flags |= SHORTCALL_HL;
         } else if (amatch("__preserves_regs")) {
             int c;
             needchar('(');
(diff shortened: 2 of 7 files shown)


suborb:
Sorry, I've not had a chance to look at this - I will do soon and appreciate the time you've put into this.

I'm just wondering if this is actually the right approach for interfacing to Spectranet? I don't know the API but I'm surprised that parameters in registers (beyond the call number) aren't needed for the API calls. 

My gut feeling is that it's going to more portable to create stub library routines that calls "rst $28" regardless of whether parameters are required - at least then the application code could be compiled with sdcc or the library swapped out for alternate TCP stacks (for example, the Next ESP)

desertkun:
First of all, I would like to appreciate your response and effort paying attention into this.

On the subject, that is not a function call, but rather a module call.  Spectranet has an extra 4k-banked 112k of user ROM which are mapped into 0x2000 - 0x2FFF. The way this works, is you place your function code on top of the module, push the module id onto HL and then do `CALL MODULECALL` when you are calling outsize of page (page out) or `RST MODULECALL_NOPAGE` when you're calling one module from another, in same memory `0x2xxx` space. And lastly, arguments for a function are pushed onto stack, and then a module is called, and the module then does `JP` to a correct C function inside the module, that unwinds the stack.

What I am trying to do here, is to make C pretend a module is a normal function, which can be called from C. My goal is to get something as big as [axtls](http://axtls.sourceforge.net/) split it into dozens of 4k-modules, fill whole ROM with it and hopefully stretch the limits of poor machine enough to get SSL running. Then, each use from another module will get this declspec.

suborb:
Ah, now I understand.

For some reason there's been a lot of banking discussions floating around lately and I'm always interested in how people solve it.

I guess with this model you're effectively planning on compiling each API function down into a binary per page (ignoring common internal functions for a moment)? This will mean there's copies of the crt0 routines (eg l_long_mult etc) in each binary, however without a "common" ROM bank I can't really see an alternative to it unless you can reserve some RAM to hold them.

As an aside I think you can achieve what you're doing with the existing `__z88dk_shortcall` and the zcc option `-custom-copt-rules` and a rule file that looks this:

```
        rst     0x28
        defw    %1
=
        ld      hl,%1   ;const
        rst     0x28
```

src/sccz80/define.h
```diff
@@ -130,6 +130,7 @@ struct type_s {
         int   params_offset;
         uint8_t  shortcall_rst;
         uint16_t shortcall_value;
+        char  shortcall_hl;
```
suborb, line 133: I think adding an extra enum to symbol_flags is the best way to do this (and then obviously the change ripples through)

src/sccz80/declparse.c
```diff
@@ -592,6 +592,33 @@ static Type *parse_type(void)
     return type;
 }
 
+static inline void parse_trailing_shortcall(Type *type, char hl) {
```
suborb, line 595: Ideally this should just be named `parse_shortcall()`

src/sccz80/codegen.c
```diff
@@ -1114,14 +1114,19 @@ void gen_call(int arg_count, const char *name, SYMBOL *sym)
...
 }
 
-void gen_shortcall(Type *functype, int rst, int value) 
+void gen_shortcall(Type *functype, int rst, int value, char hl)
 {
-    if (functype->return_type->kind == KIND_LONGLONG) {
-        ol("ld\tbc,__i64_acc");
-        push("bc");
+    if (hl) {
+        outfmt("\tld hl,\t%d\n", value);
```
suborb, line 1120: Call `vconst(val)` - it will might up an opportunity for optimisation by the peep hole rules.

src/sccz80/codegen.c
```diff
@@ -1114,14 +1114,19 @@ void gen_call(int arg_count, const char *name, SYMBOL *sym)
     ot("call\t"); outname(name, dopref(sym)); nl();
 }
 
-void gen_shortcall(Type *functype, int rst, int value) 
+void gen_shortcall(Type *functype, int rst, int value, char hl)
```
suborb, line 1117: No need to pass in `hl` here, you can pull that parameter off `functype`

testsuite/results/shortcall.opt
```diff
@@ -40,6 +40,22 @@
...
 
 
+._func_hl
+	ld	hl,1	;const
+	ld	de,0
+	push	de
+	push	hl
+	ld	hl,2	;const
+	push	hl
+	ld hl,	2000
```
suborb, line 50: The `vconst()` comment above will "fix" the indentation here.

desertkun:
Thanks for the review, I will fix the comments soon. On other comment on `crt0` you made, yes, you are right, I was surpized to see that my <4k builds were actually coming out as 5+k. Could you please give me some tips on how ot make the linker think those builtins are located at some memory? And how to build crt0 separately as a binary blob? Say I'll force the user to allocate specific version of crt0 at specific location as part of the actual program in normal memory (`0x5B00+`) and the actuall program will need those anyway.

suborb:
To share the routines, you'll need to compile with a custom crt0 which redirects the runtime routines to known addresses. I've not tried it out, but:

    zcc +zx -crt0 module_crt0.asm module.c

Where module_crt0 has the following contents (you'll need to add which routines are used):

```
PUBLIC l_gint
defc l_gint = 24576
....
```

You can find out which routines are needed by seeing which are missing when you try to create the binary.

The RAM blob will need to be created with a jump block:

```
ramblob_crt0.asm:

org 24576

jp l_gint
jp ...
```

And then compile with:

   zcc +zx -crt0 ramblob_crt0.asm emptyfile.asm

As I mentioned, I've not tried it so I may not have it completely right, but hopefully you get the gist.

desertkun:
Thank you for explanation. Sorry if I am abusing this discussion, as I move along, I run into issues.

When I call, simple function, e.g. `uint8_t bump(uint8_t a) <declspec>; uint8_t ret = bump(5);` it ends up pushing 4 extra words down the stack for spectranet purposes before actually calling the implementation.

green: a and return address; red: spectranet stuff down the stack


Is there a way to tell the implementation (aka `uint8_t bump_impl(uint8_t a) {`) that stack arguments and return address are offset by 8 bytes? Unwinding the stack and winding it back seems way too painful.

suborb:
You'll want to use `__z88dk_params_offset(8)` declspec on the `bump` prototype.

The inter-bank calling in z88dk was originally using it, but for non library use it was a bit cumbersome so I switched over to saving the "extra" parameters on a secondary stack fairly recently.

desertkun:
I have finally figured this out. Unfortunately `__z88dk_params_offset` won't do since I actually need three kinds of calls to the same function: a direct call within 4k module, external call from regular programm (`CALL MODULECALL`) or call from another module (`RST 0x28`), all of which push 0, 8 and 4 bytes of garbage up the stack before the actual call.

So I have to rely on `__z88dk_fastcall` with only single parameter-pointer to a struct now, the only problem is, `HL` has to be occupied by by `__z88dk_fastcall`. Currently I had to hack the thing like so:
```asm
LD BC, HL
LD HL, %address%
RST 0x28 / CALL MODULECALL

; my module handler
PUSH BC
; some code
POP HL
JP %function%
```
(copy HL to to BC so I can later load BC back into HL and perform the jump to the function body.

I see that by the spec of `__z88dk_fastcall`, the registers are HL/DE. Looking on the code and cannot find the code that sets/reads the HL. The idea is, add an argument that would specify the register required, e.g. 

```c
void my_func(struct args_t* args) __z88dk_hl_call(0xAF01, 0x3FF8) __z88dk_fastcall('bc') __smallc;
```

So my current angle of attack is:
1. Add an optional argument to `__z88dk_fastcall`, needed to preserve HL for two guys below
2. Add new `__z88dk_hl_call(module, addr)` (would generate `LD HL, module; CALL addr`, needed to call spectranet from user space)
2. Add new `__z88dk_rst_call(module, rst)` (would generate `LD HL, module; RST rst`, needed to call spectranet form other module)

What do you think?

suborb:
Yes that’s the problem with the stack offset code. 

You can obviously come down to 2 cases - 0 offset and 8 offset. 

Thus the 8 offset could be written as a wrapper around the real function which should. 

It’s all really messy though - using a secondary stack to hold the stuffed variables means that everything is 0 offset which really simplifies things. There’s a bit more cost to the calling but imo the cost is worth it for the simplification.

desertkun:
That's what I tried originally, did not work for whatever reason, so I tried with fastcall which worked great, and only one signature for everone. The problem with HL still stands, so I woulder if you could bring me some clues so I could override it to use DE or BC instead.

IMG_20201003_164319_456

suborb:
So the fastcall convention comes out of hl being the primary register in sccz80 - all values are loaded into hl so it was natural to pass the parameter in the value.

bc is treated as a scratch register - it's used in function calls to populate the address where int64 values should be returned.

Given that you've implemented `gen_rst_call()` (or whatever it's called) then inserting `ld b,h ; ld c,l` is reasonable to do there. On the receiving side the passed parameter is pushed onto the stack in declparse.c::declfunc() at around 1788. So with an appropriate annotation you can handle that side as well.

desertkun:
Ok, here's the gist:
1. `__z88dk_shortcall` is not modified
2. `__z88dk_shortcall_hl` is a flag, meant to follow `__z88dk_shortcall` aka `__z88dk_shortcall(xxxx, xx) __z88dk_shortcall_hl` that would mean that instead of defw, a LD HL should be used
3. new `__z88dk_hl_call(module, address)` that'd generate `LD HL, %module%; CALL %address%`
4. `__z88dk_preserve_hl2bc` flag that could be specified on either of above, that'd before loading `HL`, would preserve original `HL` into `BC`.

If these changes make no sense for general use, that's fine. But if you see potential (maybe other spectranet users), then I would bother writing regression.

suborb:
I'm not sure that an explicit `__z88dk_preserve_hl2bc` is required in the general case?

- `__z88dk_hl_call() __z88dk_fastcall` implies `__z88dk_preserve_hl2bc` in order to pass (and receive) the parameter properly.

Likewise:

- `__z88dk_shortcall_hl() __z88dk_fastcall` implies  `__z88dk_preserve_hl2bc` in order to pass (and receive) the parameter properly.

Which means that flag should be set automatically? Not setting it would result in broken code being generated which is not at all desirable.

I can't see the handling on the `declfunc()` side to receive `bc` as the `__z88dk_fastcall` parameter - I might have missed it though.

desertkun:
Thanks on that hint; makes sense, does it make sense to produce a warning that HL is copied over because of a conflict?

About the BC thing, yeah, that's because it is expected to be handled by the user when the module is called (and not a function, the module could even not call the function in question).

suborb:
I think a warning would generate far too much noise - an example implementation and docs on https://github.com/z88dk/z88dk/wiki/CallingConventions would be enough - the bc thing can be explained away there as well.

desertkun:
I have updated the `shortcall.c` testsuite. Does that constitute an example? As for the wiki, I have updated that as well.

suborb:
Sorry @desertkun, this one completely slipped my mind - I'll get in this weekend

PR #1647, opened 2020-11-28
[m100] basic support for generating option ROMs

KP1533TM2:
Hello.
This version is now able to produce basic option ROM software for TRS-80 Model 100/102 computers.
Option ROMs can be used to add extra software to those machines without much RAM penalty.
They can be called from machine's BASIC typically with `call 63012` command. Some software packages released on option ROMs also drop a small file in RAM as a simple launcher.

At the moment only console I/O is supported for option ROMs. I haven't looked into more advanced features (i.e. graphics) but if they use firmware calls, they now should be easily adaptable.

The main hurdle of making option ROM in either machine language or in C is the awkward bank-switching of M100. System ROM aka 'firmware' occupies same address space 0x0000...0x7FFF as the optional ROM and machine switches between ROMs back and forth when needed. Thus, using firmware calls in option ROM is tricky.
[Luckily this trickiness has already been solved at least once for machine code by Stephen Adolph.](http://bitchin100.com/wiki/index.php?title=OPTROM_Switching&oldid=2419)
I merely translated Stephen's template into Zilog mnemonics and made a CRT out of it.
There's now also ROMCALL macro used during clib compilation, which allows to build 2 versions of m100 clib from the same source set - one calls firmware from RAM, and another one uses template-provided firmware-calling RST 6.

CRT_ORG_DATA and CRT_ORG_BSS are pointed to ALTLCD area of RAM, which leaves C program with only around 320 bytes of RAM, including temporary math variables etc. ALTLCD normally keeps the previous text display page, and storing stuff there shouldn't interfere with anything else on the machine.

I suppose for more memory, it's either possible to access it directly, or use firmware functions to avoid interfering with OS and other files in RAM.

Both A.CO and option ROM targets now use 'native' stdio routines since 'generic' ones didn't seem to handle newlines at all on this machine.

Apart from anything already mentioned A.CO target shouldn't have changed.

**Usage:**
`zcc +m100 -subtype=default ...` compiles `A.CO` target for RAM
`zcc +m100 -subtype=optrom ...` compiles for option ROM

Снимок экрана_2020-11-28_21-39-35

m100

### DIFF
--- a/lib/target/m100/classic/optrom.asm
+++ b/lib/target/m100/classic/optrom.asm
@@ -90,15 +90,14 @@ stdcall:
 	
 	push	hl
 	push	de
-	
-	defb	$38	; ldsi $06, 8085 undocumented opcode
-	defb	$06
 
-	defb	$ed ; lhlx undocumented 8085 opcode
+	ldsi	$06	
+
+	lhlx
 	dec	hl
 	dec	hl
 	ex	de, hl
-	defb	$ed ; lhlx undocumented 8085 opcode
+	lhlx
 	
 	pop	de
 	ex	(sp), hl
@@ -108,9 +107,8 @@ intcall:
 	push	hl
 	push	de
 	
-	defb	$38	; ldsi $04, 8085 undocumented opcode
-	defb	$04
-	defb	$ed ; lhlx undocumented 8085 opcode
+	ldsi	$04
+	lhlx
 
 	dec	hl
 	dec	hl
@@ -119,7 +117,8 @@ intcall:
 
 	push	hl	
 	ld	hl, OPON
-	defb	$d9	; shlx, 8085 undocumented opcode
+
+	shlx
 	pop	hl
 	pop	de
 	ex	(sp), hl	
--- a/lib/target/m100/classic/optrom.asm
+++ b/lib/target/m100/classic/optrom.asm
@@ -1,0 +1,161 @@
+;
+;	Startup for m100 optrom mostly based on Stephen Adolph optrom demo:
+;	http://bitchin100.com/wiki/index.php?title=OPTROM_Switching&oldid=2419
+;	Alexei Gordeev, Nov 2020
+;
+
+	; guaranteed safe place to put some variables
+	defc ALTLCD_RAM = $FCC0
+	defc ALTLCD_LEN = 320        
+        
+        defc OPON = $FAA4
+        defc CRT_ORG_DATA = ALTLCD_RAM
+        defc CRT_ORG_CODE = $0000
+
+IFNDEF CRT_ORG_BSS
+        defc CRT_ORG_BSS = ALTLCD_RAM ; Ram variables will be kept in safe RAM area by default
+ENDIF
+        defc    __crt_org_bss = CRT_ORG_BSS
+
+	org	  CRT_ORG_CODE
+rst0:
+	jp	program		; RST0 standard entry
+
+	defs	$08-ASMPC
+rst1:
+	ret			; RST1 not used
+	defs	$10-ASMPC
+rst2:
+	ret			; RST2 not used
+
+	defs	$18-ASMPC
+rst3:
+	ret			; RST3 not used
+
+	defs	$20-ASMPC
+rst4:
+	ret			; RST4 not used
+
+	defs	$24-ASMPC
+trap:
+	di			; TRAP
+	call	intcall
+
+	defs	$28-ASMPC
+rst5:
+	ret			; RST5 not used
+
+	defs	$2c-ASMPC
+rst55:
+	di
+	call	intcall
+
+	defs	$30-ASMPC
+rst6:
+	jp	stdcall		;RST 6 used as short call to a Standard ROM routine.
+
+	defs	$34-ASMPC
+rst65:
+	di	;Replaces the 6.5 interrupt and sets up a call to 6.5 in Standard ROM
+	call	intcall
+
+	defs	$38-ASMPC
+rst7:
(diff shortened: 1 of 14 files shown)


lib/target/m100/classic/optrom.asm
```diff
@@ -0,0 +1,161 @@
...
+	
+	pop	de
+	ex	(sp), hl
+	jp	stdon
+
+intcall:
+	push	hl
+	push	de
+	
+	defb	$38	; ldsi $04, 8085 undocumented opcode
```
suborb, line 111: These undocumented opcodes are supported by z80asm with the -m8085 flag

lib/target/m100/def/romcalls.def
```diff
@@ -0,0 +1,32 @@
...
+; And if we're calling from normal RAM, then we simply do CALL ROM_ADDR
+
+; With ROMCALL macro, all ROM calls (except RSTs) can be abstracted into
+; ROMCALL; DEFW ROM_ADDR
+; ...and be conditionally compiled depending on the subtype we use.
+
+#define ROMCALL defb ROMCALL_OP
+
+IF __OPTROM__
+ROMCALL_OP EQU $F7 ; rst 6 if calling from OPTROM
```
suborb, line 17: I think, that if you compile the library with EXTERN ROMCALL_OP and then defc ROMCALL_OP (and PUBLIC it) in the crt file, then we'll only need a single library, z80asm should resolve the value at link time.

KP1533TM2:
Ah, I see, EXTERN was the magic word all along! I was struggling for a couple of days trying to figure out how to do a single library, thank you!
Both suggestions have worked, I'll update my repository soon.

In the meantime I noticed fgets() is acting funny, it shows 2 cursors, the underscore and the system one, like that: `_⬛`, and it doesn't recognize ASCII 0x08 backspace as a control sequence and stores it in buffer as a normal character.

suborb:
> In the meantime I noticed fgets() is acting funny, it shows 2 cursors, the underscore and the system one, like that: _⬛, and it doesn't recognize ASCII 0x08 backspace as a control sequence and stores it in buffer as a normal character.

There's flags for those as well, in case you've not found them here they are.

In m100_crt0.asm, set the following (it's currently 12):

```
defc    CRT_KEY_DEL = 8
```

And:

```
defc CLIB_DISABLE_FGETS_CURSOR  = 1
```
To disable the software cursor in `fgets()`.

suborb:
> Both A.CO and option ROM targets now use 'native' stdio routines since 'generic' ones didn't seem to handle newlines at all on this machine.

That's really odd, the driver is just pokes $f63a/$f639 with the XY coordinates and calls `rst $20` so it should be doing the same thing as the native driver. I noticed you changed the screen size, so that might have been the cause of the problem if you were using long lines.

KP1533TM2:
The problem with generic console that I experience is that `\n` or `\r` doesn't have any effect at all, i.e.
`printf("Hello world\n"); fgetc_cons()` would result in cursor still sitting at the end of the line, and now on the new line. Appending more `\n` or `\r\n` to the end of the string also has no effect.
Perhaps for explicit positioning CURPOS ($42C7) ROM call should be used, or its effect replicated in generic console routine. Although I suspect it might be slower because it seems like the firmware keeps track of cursor position anyway.

Regarding CONSOLE_ROWS/CONSOLE_COLUMNS, I thought those defines should represent actual console size; aren't they?

suborb:
> The problem with generic console that I experience is that \n or \r doesn't have any effect at all, i.e.
printf("Hello world\n"); fgetc_cons() would result in cursor still sitting at the end of the line, and now on the new line. Appending more \n or \r\n to the end of the string also has no effect.

Ah yes, I understand now. The coordinates aren't updated until the next character is printed so the hardware cursor ends up at the end of the previous line. Let me have a think about how to solve that.

> Regarding CONSOLE_ROWS/CONSOLE_COLUMNS, I thought those definitions should represent actual screen size; aren't they?

Yes they should be, I've seen odd effects when the number of columns defined doesn't match up with the hardware so I assumed that that was the problem.

KP1533TM2:
Looking briefly at M100 ROM disassembly, it performs some kind of an escape sequence when positioning with ROM routine.
I restored `TAR__fputc_cons_generic` definition back to see how it works, made a small program with different printfs and it indeed seems to position text properly, although leaving firmware cursor behind sometimes.
However, it disregards the previous (BASIC console) positioning and always starts at the top left. Looks a bit messy :)
Снимок экрана_2020-11-29_18-47-39

CURPOS address should be $427C, I'll correct that. Oops.

suborb:
I'll clean up generic console a bit when this is merged - I think we can do a few things to make it nicer - the benefit of having gencon running is of course portability.

Could I ask you to do some updates to the wiki page: https://github.com/z88dk/z88dk/wiki/Platform---M100  - it's a bit brief at the moment, but if you could add some information on new flag and how to load into an emulator that will save a lot of questions in the future.

KP1533TM2:
Sure, no problem.
I'm trying to make graphics routines work at the moment. It runs on emulator (although line drawing is a somewhat glitchy), I'm having a bit of unexpected difficulties with my real M100.

KP1533TM2:
IMG_20201129_233314

I guess I don't have much more to add at the moment in terms of code.
Line drawing and dot plotting seems to work for both subtypes. Unplotting a circle doesn't seem to be implemented.

KP1533TM2:
I've updated the page. Looks good to me :)

suborb:
Thank you so much for the code, words and for testing on real hardware!

PR #1648, opened 2020-11-28
Changed banked_call to _banked_call.

hackneyc:
It seems the banked call code for SDCC tries to call the wrong function which causes a compilation error when using __banked from C to call another banked C function. Adding the _ (underscore) to the banked_call function resolves this issue.

### DIFF
--- a/libsrc/_DEVELOPMENT/sdcc_opt.1
+++ b/libsrc/_DEVELOPMENT/sdcc_opt.1
@@ -558,23 +558,23 @@ __xinit_%0
 =
 	ld	hl,sp + %2
 
-	call	banked_call
+	call	_banked_call
 	DEFW	%1
 	DEFW 0
 =
-	call	banked_call
+	call	_banked_call
 	defq	%1
 
 	call	___sdcc_bcall
 	DEFW	%1
 	DEFW	%2
 =
-	call	banked_call
+	call	_banked_call
 	defq	%1
 
 	ld	e,%1
 	ld	hl,%2
 	call	___sdcc_bcall_ehl
 =
-	call	banked_call
+	call	_banked_call
 	defq	%2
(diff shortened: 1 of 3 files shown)


suborb:
Now I'm starting to wonder if I tested banking with sdcc..

I think this is the incorrect change, the symbol called by sdcc is `banked_call` which is the name of the implemented function in the targets that support it.

The problem lies at the top of the sdcc_opt.X file where there's an `EXTERN _banked_call` which should actually be `EXTERN banked_call`

hackneyc:
Could it be the EXTERN in the same file which is incorrect? It is EXTERN _banked_call so there is a mismatch between the asm code generated and the EXTERN reference. Which caused a compilation error in the C code even when using -c to compile only.
I assumed the EXTERN was correct and changed the calls accordingly. If other targets use banked_call instead of _banked_call it is the EXTERN which is incorrect.

hackneyc:
I can make that change and test and update the pull if that's ok?

suborb:
@hackneyc that would be great, thanks!

hackneyc:
While I have you. Where should banked_call be implemented for ZX? I have ported code from the other compiler and tested it, I assume it needs to be included in some library somewhere?

suborb:
Classic has an implementation here: https://github.com/z88dk/z88dk/blob/master/libsrc/target/zx/banked_call.asm

hackneyc:
That's the one I grabbed. I'm using it with sdcc_iy which doesn't seem to have its own implementation. I was wondering if sdcc_iy should have it's own and if so where it should go?

suborb:
I guess there's two possible places:

* _DEVELOPMENT/arch/zx/misc/z80
* _DEVELOPMENT/target/zx/driver/banking

I don't do much work on newlib so I've not got particularly strong opinions either way, I'm probably leaning slightly towards the target/ since it'll be easier to have a different implementation for the ZX Next.

suborb:
@hackneyc This looks complete I think, unless there's anything you want to add?

hackneyc:
@suborb ,good to go from my side.

PR #1658, opened 2020-12-20
ucpp: fix line numbering with \r\n

desertkun:
I've finally figured out behavior for #1653 the ucpp preprocessor is incorrectly handling `\r\n` endlines on Windows. I know this compiler does not support windows, but it works with WSL and I would like to be able to compile windows-edited files under WSL.

The fix is simple, when we hit `\r\n` (figured out by looking ahead of buffer one char), ignore the current one (`\r`) and process just the next one `\n`.

Tested both with windows-edited files (`\r\n`) and linux ones (`\n`), the following file (`a.h` is an empty file)

```
#include "a.h"
/*
 * comment
 */
int b = 20;
```

is treated like so all the times

```
#line 1 "main.c"
#line 1 "a.h"

#line 2 "main.c"
 


int b = 20;
```

### DIFF
--- a/src/ucpp/cpp.c
+++ b/src/ucpp/cpp.c
@@ -335,7 +335,6 @@ static void reinit_lexer_state(struct lexer_state *ls, int wb)
 	ls->input = 0;
 	ls->ebuf = ls->pbuf = 0;
 	ls->nlka = 0;
-	ls->macfile = 0;
 	ls->discard = 1;
 	ls->last = 0;		/* we suppose '\n' is not 0 */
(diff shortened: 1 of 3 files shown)


suborb:
Thank you for cracking this one. Hopefully this solves the random issues that we've had over the years. It looks like you have a nice reproducible case as well, which makes me wonder if some of the issues in the past have been due non-consistent conventions.

BTW, there is a windows native binary being produced so you don't have to use WSL (though under WSL2 it probably is quicker than win32).

PR #1682, opened 2021-01-22
Prevent asmpp.pl from converting the quoted file name for the "BINARY" directive to ASCII hex and small change to asmstyle.pl

hackneyc:
asmpp.pl
Converts all strings, i.e. values in double quotes to ASCII hex. This causes an assembler error for the "BINARY" directive as the file name is converted to a comma separated list of hex values. The change added is to skip this conversion for the "BINARY" directive. Probably a better change would be to include the actual binary file and convert it to "DB  ..." but this is beyond my Perl skills. Also, explicit check for IF, IFDEF. IFNDEF, ELSE, and ENDIF in column 0. Currently they are treated as labels instead of directives.

asmstyle.pl
1) Updated to treat white space at the beginning of a line followed by a ; the same is a ; at the beginning of a line
2) { and } are now treated like an opcode with no parameters used with defvars
3) Added support for formatting defvars structures.
4) Added support for formatting macros as supported by asmpp.pl.


pauloscustodio:
Thank you for the PR.

PR #1684, opened 2021-01-26
fix CONSOLE_COLUMNS definition

nippur72:
this seems a copy&paste error to me it should be `CONSOLE_COLUMNS` not `CONSOLE_ROWS`


suborb:
Yup, thank you. Normally they are defined together, but this is more correct.

Thanks.

PR #1685, opened 2021-01-26
Fix Enterprise joystick handling

dp304:
The `joystick()` function on Enterprise didn't work (it always returned the input parameter).

This patch fixes it.


suborb:
Thanks again.

PR #1691, opened 2021-01-30
zsdcc - bump to 12036 (bugfix)

feilipu:
Thinking that a new z88dk release is due soon, this small bump to sdcc adds fixes for a stack frame optimisation bug, and to improve a rule.

2020-01-29 Philipp Klaus Krause
	* src/z80/peep.c:
	  Fix bug #3173 (this is probably causing bug #3171 and #3166 too... ?)
	* src/z80/peeph.def:
	  Split rule for add from adc, sbc rule due to differences in flag handling.


suborb:
And this time I've updated tarballs and binaries a lot quicker.

PR #1720, opened 2021-03-18
Df collision doc 1

derekfountain:
This is another article in my C programmer's guide for the ZX Spectrum - sprite collision detection techniques. One article, lots of associated images, and some minor updates to other articles.

It's the first in 3 years. AA used to proof read these before pulling, but if that's not possible these days then maybe someone else could cast an eye over it, or just accept it as it is? I'm confident it's sound and accurate, and will fix any issues as they're pointed out to me.


suborb:
Thanks @derekfountain - it's really appreciated.

I'll have a read through and then merge

feilipu:
I should refresh the browser before clicking.

PR #1763, opened 2021-04-20
[laser500] include extra keys in inKey.asm

nippur72:
in the target `laser500` there is a `CLIB_LASER500_SCAN_EXTRA_ROWS` flag disabled by default that prevents reading the extended keys (cursor keys, fkeys etc) with the keyboard input routines. 

That flag was added for some reason because early MAME versions didn't read the keyboard properly. But now MAME works correctly and we also have the much more accurate [web emulator](https://nippur72.github.io/laser500emu) so there is no reason why this flag should be enforced by default or have it in the first place.

This PR removes it.

I also fixed a reference to the old target name "vz700" which is now `laser500`.

### DIFF
--- a/lib/target/laser500/classic/laser500_crt0.asm
+++ b/lib/target/laser500/classic/laser500_crt0.asm
@@ -1,9 +1,8 @@
 ;
-;	Startup for V-Tech VZ-350/500/700?
+;	Startup for V-Tech Laser 350/500/700
 ;
 
-	module	vz700_crt0 
-
+	module	laser500_crt0
 
 ;--------
 ; Include zcc_opt.def to find out some info
--- a/lib/target/laser500/classic/laser500_crt0.asm
+++ b/lib/target/laser500/classic/laser500_crt0.asm
@@ -24,7 +24,6 @@
 	defc	CONSOLE_COLUMNS = 40
 	defc	CONSOLE_ROWS = 24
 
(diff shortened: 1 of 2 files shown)


suborb:
Thank you @nippur72 - are there any firmware updates that we need to update the address for?

nippur72:
> Thank you @nippur72 - are there any firmware updates that we need to update the addresses for?

are you talking about `lm80c` target, right? Yes there are but I am going to create PRs myslef as long as I'm able to build z88dk on my local host. BTW, is there a guide or forum post other than the `README.md` ? I'm trying to build it in Windows with "Windows Subsystem for Linux" (aka WSL).

suborb:
> are you talking about `lm80c` target, right? Yes there are but I am going to create PRs myslef as long as I'm able to build z88dk on my local host. BTW, is there a guide or forum post other than the `README.md` ? I'm trying to build it in Windows with "Windows Subsystem for Linux" (aka WSL).

Ah yes, sorry that's my own confusion. Apologies.

There's a wiki page here: https://github.com/z88dk/z88dk/wiki/installation#linux--unix which is the recipe for building everything on Linux. Alternatively, if you just want to build the libraries on windows, then this is the recipe: https://github.com/z88dk/z88dk/wiki/Classic-Building-the-libraries

Issue #1768, opened 2021-04-29
(sdcc) Codegen issues


suborb:
Reported on the forum: https://z88dk.org/forum/viewtopic.php?f=12&t=11519

```
#include <stdio.h>

typedef unsigned long   uint32;
typedef unsigned short  addr16;
typedef unsigned short  uint16;
typedef unsigned char   uint8;

uint8* stack = 0;
addr16 stackTop = 0;

static uint32 stackPeek() {
    uint8* location = stack + stackTop;
    return *((uint32*)location);
}

static uint32 stackPop() {
    uint32 value = stackPeek();
    stackTop += 4;
    return value;
}

void main() {
    stackPop();
    printf ("%x ", stackTop);
    stackPop();
    printf ("%x ", stackTop);
    stackPop();
    printf ("%x ", stackTop);
    stackPop();
    printf ("%x ", stackTop);
    stackPop();
    printf ("%x ", stackTop);
}
```

Compile with: `zcc +zx -vn -clib=sdcc_iy -startup=0 Bug1.c -o bug -create-app`

Expected to see "4 8 c 10 14" but got "404 808 c0c 1010 1414"

Note: using --max-allocs-per-node0 options fixes the problem

Second issue:

```
#include <stdio.h>

typedef unsigned long   uint32;
typedef unsigned short  addr16;
typedef unsigned short  uint16;
typedef unsigned char   uint8;

uint8* memory = 0;

void memWrite(uint32 sectorNumber){
    uint16 value = 200;
    uint32* byte51202 = (uint32*)(memory + 51202);
    *byte51202 = sectorNumber;
    memory[51210]=100;
    memory[51212]=value;
}

void main() {
    memWrite(0);
    printf ("%u ", memory[51210]);
    printf ("%u ", memory[51212]);
}
```

Compile with: `zcc +zx -vn -clib=sdcc_iy -startup=0 Bug2.c -o bug -create-app`

Expected to see "100 200" but got "0 0"

Using --max-allocs-per-node0 option did not help, output was "0 200"

suborb:
Second one looks like a --reserve-regs-iy bug:

```
dom@heliotrope:~/z88dk:250% zcc +test stk.c -compiler=sdcc --reserve-regs-iy
dom@heliotrope:~/z88dk:251% z88dk-ticks a.bin
0 0
Ticks: 28129
dom@heliotrope:~/z88dk:252% zcc +test stk.c -compiler=sdcc
[...]

PR #1769, opened 2021-04-29
zsdcc - update r12250 Version 4.1.4

feilipu:
I think this is all that is needed now.
AFK currently on loan machine.
Hopefully addresses #1768.


feilipu:
Grr. Forgot to switch branch to master before commiting the math32 stuff.
Anyway. Shouldn't be a problem to leave it, except a little messy.

suborb:
Thank you for doing this, I've now built all the flavours so it's merge time!

PR #1770, opened 2021-04-29
Command line pragma cleanup

suborb:
Add some pragmas/rename some pragmas:

* `#pragma define CRT_ENABLE_COMMANDLINE = 0` - disable command line arguments for targets that support it
* `#pragma define CRT_COMMANDLINE_REDIRECTION=0` - disable command line redirection. If not specified, it will be enabled if `CRT_ENABLE_STDIO=1` 
* `#pragma define CLIB_OPEN_MAX=xx` - set the number of fcbs for CPM fileio

Old pragma mappings:

* nostreams -> CRT_COMMANDLINE_REDIRECTION = 0
* nofileio -> CLIB_OPEN_MAX=0


lib/crt/classic/crt_cpm_fcntl.asm
```diff
@@ -0,0 +1,21 @@
...
+    IF !DEFINED_CLIB_FCBS_MAX
+        DEFC    CLIB_FCBS_MAX = CLIB_FOPEN_MAX
+    ENDIF
+    PUBLIC  __CPM_FCBS_MAX
+    defc    __CPM_FCBS_MAX = CLIB_FCBS_MAX
+
+
+    SECTION bss_crt
+
+IF CLIB_DISABLE_FILEIO = 0
```
suborb, line 13: Lets be consistent and use ENABLE from now on

suborb:
Need to update wiki before merge...

PR #1782, opened 2021-05-10
newlib - align before section (rather than after)

feilipu:
Currently `align` is done after the `section` rather than before.

This PR moves the `align xxx` before the `section_align_xxx` statement.

IIRC there was some discussion about which was the right order, but IMHO it makes more sense to have each section starting with the right alignment, and to ensure that each following smaller section is also starting with the right alignment offset.

If this is not correct assumption, please just close this PR with a note for future reference.

### DIFF
--- a/libsrc/_DEVELOPMENT/target/crt_memory_model_z180.bak
+++ b/libsrc/_DEVELOPMENT/target/crt_memory_model_z180.bak
@@ -36,29 +36,29 @@ section code_lib
 section code_compiler
 section code_user
 
-section rodata_align_256
 align 256
+section rodata_align_256
 
-section rodata_align_128
 align 128
+section rodata_align_128
 
-section rodata_align_64
 align 64
+section rodata_align_64
 
-section rodata_align_32
 align 32
+section rodata_align_32
 
-section rodata_align_16
 align 16
+section rodata_align_16
 
-section rodata_align_8
 align 8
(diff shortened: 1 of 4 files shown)


suborb:
I thought that anything placed in the section would be placed after the align, so the data itself would be aligned even if the section name itself wasn't. So this comes down to a preference/presentation issue rather than a functionality issue.

In which case, I think having the `__head` of the sections be aligned correctly does seem more logical.

feilipu:
> I thought that anything placed in the section would be placed after the align, so the data itself would be aligned even if the section name itself wasn't.

Yes, perhaps that was the discussion point I was distantly remembering?

But, if `align` after `section` applies alignment within the prior `section`, it does make me question which alignment (for example) the contents of `section rodata_align_256` would have with the `align 256` before the section name but `align 128` after the section name?

IDK, but although the PR suggestion looks clear, perhaps it is not correct?

feilipu:
@pauloscustodio Do you have a thought about this?

But, *if* an `align` after `section` *applies alignment to the prior section*, it does make me question which alignment (for example) the contents of `section rodata_align_256` would have with the `align 256` *before* the section name but `align 128` *after* the section name?

feilipu:
Just closing this as it is hanging around.
If it turns out to be sensible, then it can be reopened.

PR #1788, opened 2021-05-13
Resolved "Unescaped left brace..."

iratahack:
Unescaped left brace in regex is deprecated here (and will be fatal in Perl 5.32), passed through in regex; marked by <-- HERE in m/^\s*({ <-- HERE |})/ at ./asmstyle.pl line 97."


pauloscustodio:
@iratahack , thank you for the fix.

PR #1793, opened 2021-05-21
Remove tabs after opcodes and directives with no arguments.

iratahack:
asmstyle.pl would insert tabs after opcodes and directives even if they had no arguments. This update prevents the insertion of these tabs.
Updated Makefile to install asmstyle.pl in $PREFIX/bin


pauloscustodio:
Thanks for contributing, @iratahack

PR #1798, opened 2021-05-28
Build infrastructure fixes.

iratahack:
This started because I only wanted to build the zx target so I was passing `-p zx` to the build script. I found this caused problems building SDCC due to the use of `TARGETS ?= all` in the SDCC Makefile. I have added the following changes...

1) Ensure `$MAKE` is used everywhere instead of `make`. For this work the code which sets `$MAKE` based on the OS type has been moved towards the beginning of the script.
2) If `do_clean` or `do_clean_bin` are set perform a `make clean` from the top level directory and in libsrc.
3) Removed `make clean` when building libs. Maybe this is there due to incorrect make dependencies?
4) Exporting `TARGETS` from the `-p` option too soon causes problems when building SDCC since it uses `TARGETS ?= all` in its Makefile. `TARGETS` are now only passed to the libsrc Makefiles.
5) Ensure testsuite Makefile is multi-job safe. The existing Makefile used the same file names for different targets which causes problems when -j is used to invoke make.
6) Building the zxn target in _DEVELOPMENT requires files derived from the zx target build and the zx target build depends on the z80 target build.

That's it. ;)


suborb:
@iratahack - sorry I'm afk at the moment, I'll take a look at this soon though

suborb:
That looks good, thanks for pushing forward with the multi-build safety. We're getting there!

PR #1803, opened 2021-06-05
Prefix all programs + expose the perl utils in snap

iratahack:
1) Added z88dk- prefix to:
    sccz80
    zsdcc
    zsdcpp
    asmpp.pl
    asmstyle.pl
2) Updated snapcraft.yaml to include zsdcc, zsdcpp, asmpp.pl and asmstyle.pl
3) Added new command-line option to z88dk-appmake, '--bank-order' used to specify
the order the memory banks are stored in the tap file for ZX target, rather than the default '01234567'.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -96,8 +96,6 @@ fi
 
 if [ $do_clean = 1 -o $do_clean_bin = 1 ]; then              # Dont remove bin, as zsdcc and szdcpp must be built by hand in win32
   $MAKE clean
-  $MAKE -C libsrc clean
-  $MAKE -C libsrc/_DEVELOPMENT clean
 fi
 
 
--- a/build.sh
+++ b/build.sh
@@ -97,6 +97,7 @@ fi
 if [ $do_clean = 1 -o $do_clean_bin = 1 ]; then              # Dont remove bin, as zsdcc and szdcpp must be built by hand in win32
   $MAKE clean
   $MAKE -C libsrc clean
+  $MAKE -C libsrc/_DEVELOPMENT clean
 fi
 
 
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -104,11 +104,5 @@ apps:
   z88dk-asmstyle.pl:
     command: bin/z88dk-asmstyle.pl
     plugs: [home, removable-media]
-  z88dk-zsdcc:
-    command: bin/z88dk-zsdcc
-    plugs: [home, removable-media]
-  z88dk-zsdcpp:
-    command: bin/z88dk-zsdcpp
-    plugs: [home, removable-media]
 
 
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -98,5 +98,17 @@ apps:
   z88dk-font2pv1000:
     command: bin/z88dk-font2pv1000
     plugs: [home, removable-media]
+  z88dk-asmpp.pl:
+    command: bin/z88dk-asmpp.pl
+    plugs: [home, removable-media]
+  z88dk-asmstyle.pl:
+    command: bin/z88dk-asmstyle.pl
+    plugs: [home, removable-media]
+  z88dk-zsdcc:
+    command: bin/z88dk-zsdcc
+    plugs: [home, removable-media]
+  z88dk-zsdcpp:
+    command: bin/z88dk-zsdcpp
+    plugs: [home, removable-media]
 
 
--- a/src/appmake/zx.c
+++ b/src/appmake/zx.c
@@ -80,7 +80,8 @@ static struct zx_tape zxt = {
     0,          // noloader
     0,          // noheader
     0,          // parity
-    0           // khz22
+    0,          // khz22
+    NULL	// bank order
 };
 
 static struct zx_sna zxs = {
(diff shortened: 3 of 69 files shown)


src/appmake/zx.c
```diff
@@ -155,6 +156,7 @@ option_t zx_options[] = {
     { 0 , "merge",     "Merge a custom loader from external TAP file", OPT_STR, &zxt.merge },
     { 0 , "blockname", "Name of the code block in tap file", OPT_STR, &zxt.blockname },
     { 0,  "clearaddr", "Address to CLEAR at",       OPT_INT,   &zxt.clear_address },
+    { 0,  "bank-order","Order banks this way in the tap file (default '01234567')", OPT_STR, &zxt.bank_order },
```
suborb, line 159: Unfortunately this change is incompatible with the built in loader here: https://github.com/z88dk/z88dk/blob/master/lib/target/zx/classic/zx_tapeloader.asm

snap/snapcraft.yaml
```diff
@@ -98,5 +98,17 @@ apps:
   z88dk-font2pv1000:
     command: bin/z88dk-font2pv1000
     plugs: [home, removable-media]
+  z88dk-asmpp.pl:
+    command: bin/z88dk-asmpp.pl
+    plugs: [home, removable-media]
+  z88dk-asmstyle.pl:
+    command: bin/z88dk-asmstyle.pl
+    plugs: [home, removable-media]
+  z88dk-zsdcc:
```
suborb, line 107: There's no need to expose zsdcc and zsdcpp outside of the snap.

build.sh
```diff
@@ -97,6 +97,7 @@ fi
 if [ $do_clean = 1 -o $do_clean_bin = 1 ]; then              # Dont remove bin, as zsdcc and szdcpp must be built by hand in win32
   $MAKE clean
   $MAKE -C libsrc clean
+  $MAKE -C libsrc/_DEVELOPMENT clean
```
suborb, line 100: libsrc/clean should clean libsrc/_DEVELOPMENT as well

snap/snapcraft.yaml
iratahack: Got it. Backed out those 2 files.

build.sh
iratahack: Confirmed. I see the top level Makefile does a clean in libsrc which does _DEVELOPMENT in the new-clean target. Backed out these 2 lines.

src/appmake/zx.c
iratahack: Although this is incompatible with the existing loader I do think it is a useful feature to change the order the banks are written to the tap file. i.e. loading bank 5 first which could contains the screen memory plus user code negates the need to use the --screen option. How would you feel about an new loader where the user passes a pointer to a table with the banks to be loaded? This way the user can specify which order the banks are loaded? No worries if this doesn't work for you, I can back out this change.

iratahack: Hi @suborb , I've been looking at https://github.com/z88dk/z88dk/blob/master/lib/target/zx/classic/zx_tapeloader.asm and I see a few things I don't understand.
For each bank, load_block is called followed by a 'ret c' presumably to stop loading other banks if there is an error. However, load_block does 'and a' before it returns so the carry flag is always reset.
Also, the BASIC variables used (current memory page and ERR-SP) and are in bank 5, so calling load_block after bank 5 has been loaded would clobber these two memory locations in bank 5. Maybe some refactoring is needed here?

iratahack: I found the same effect can be achieved by creating a tap file for each bank and concatenating them together. I have reverted the changes.

suborb: @iratahack yes, it looks like I wasn't expecting people to load anything into banks 5,2,0.

The system variable thing is tricky - we're using a ROM routine to do the loading - I suspect we don't need to save the bank 23388, but 23613 I think does need to be saved.

I'll sort out the carry flag issue - I was obviously in two minds about what to do there!

suborb:
Thanks, I'll merge this in at the same time as I rebuild the binaries for sdcc.

src/appmake/zx.c
iratahack: @suborb, I took a look at the disassembly for ld-bytes (call 0x556) I couldn't see where it uses err-sp (23613) so maybe its not needed. Take a look https://skoolkid.github.io/rom/asm/5C3D.html and https://skoolkid.github.io/rom/asm/0556.html.

suborb: I thought it was for the break key detection?

It looks like there's a return via SA_LD_RET: https://skoolkid.github.io/rom/asm/053F.html which calls rst $08 with an error.

iratahack: Ah, yes. I got it now. I guess the only way to be 100% sure code isn't clobbered is to copy the ROM loader and remove the break detection and therefore the dependency on  err-sp. Or just ignore err-sp since there is no way to recover from break being pressed.

iratahack: @suborb , I pulled the ROM ld-bytes routine for my own project and removed the break checking, verify code and block header type checking and changed the border colors. It's less than 150 bytes when assembled. If you want to snag it instead of calling the ROM, feel free. It's here https://github.com/iratahack/CastleEscape/blob/develop/src/crt/ld_bytes.asm

suborb:
Fingers crossed....

suborb:
Just a note for information purposes, you can't export snap commands with "." in them.

PR #1811, opened 2021-06-09
zsdcc - newlib - fix peephole 12

### LINKED ISSUE #1809: zsdcc peephole regression
Trying to resolve a regression in zsdcc. This memory dump function works when `main.c` is compiled with r9958 (and also some more recent versions), but doesn't work when compiled with r12419. This blocks #1799 for me.

[`main.c`](https://github.com/z88dk/z88dk/files/6614421/main.c.txt)

```sh
# zcc +yaz180 -subtype=rom -SO3 -v -m --list --max-allocs-per-node2000000 main.c -o main -c
```

I'll raise an [issue #3248 on the sdcc bug tracker](https://sourceforge.net/p/sdcc/bugs/3248/).

__EDIT__ Bisected down to r12070 working.

feilipu:
sdcc peephole 12a and 12b were setting the location pointed to by `hl`, without first setting `hl`.
This lead to writing into a wild location.

I think deleting the offending `ld (hl),a` instruction is the right way to fix this.

__Background__

Recently SDCC has been improved to be able to use `hl` to restore the stack `sp` when returning from functions with a large number of parameters, and/or large parameters.

These means that it has become quite possible or even common for `hl` to contain the same value as `sp` following a function return, and also potentially for that value to be pointing at the function return address on the stack.

```asm
860   007C  CD 00 00    	call	_fprintf
861   007F              ;	adjustStack by 8
862   007F  21 08 00    	ld	hl,8
863   0082  39          	add	hl, sp
864   0083  F9          	ld	sp, hl    ; <<< hl and sp are the same here
865   0084              ;	genPlus
866   0084              ;setupPair	HL
867   0084              ;fetchLitPair	hl
868   0084              ;setupPair	HL
869   0084              ;fetchLitPair	hl
870   0084  3A 04 00    	ld	a,(_ya_md_origin_65536_419)
871   0087  77          	ld	(hl),a    ; <<< this has overwritten the function return address
872   0088              ; peephole z88dk-bugfix-12a
```
Closes #1809

### DIFF
--- a/libsrc/_DEVELOPMENT/sdcc_peeph.2
+++ b/libsrc/_DEVELOPMENT/sdcc_peeph.2
@@ -969,7 +969,6 @@ replace restart
 	ld	(hl),a
 } by {
 	ld	a,(#%1)
-	ld	(hl),a
 	ld	hl,#%1
 	; peephole z88dk-bugfix-12a
 }
@@ -981,7 +980,6 @@ replace restart
 	ld	(hl),a
 } by {
 	ld	a,(#%1+%2)
-	ld	(hl),a
 	ld	hl,#%1 + %2
 	; peephole z88dk-bugfix-12b
 }
--- a/libsrc/_DEVELOPMENT/sdcc_peeph.2
+++ b/libsrc/_DEVELOPMENT/sdcc_peeph.2
@@ -962,27 +962,7 @@ replace restart {
 	; peephole z88dk-bugfix-11b
 }
 
-replace restart
-{
-	ld	hl,#%1
-	ld	a,(hl)
-	ld	(hl),a
-} by {
-	ld	a,(#%1)
-	ld	hl,#%1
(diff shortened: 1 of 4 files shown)


suborb:
What was there is definitely wrong!

I don't know the sdcc output well enough to understand why it's doing a load and store to the same address: maybe memory mapped port or volatile?

The replacement is also longer and in some cases slower, so I'm not sure what the point of the rule is.

It's probably safer to delete the rule completely?

feilipu:
> It's probably safer to delete the rule completely?

I was going with the conservative approach, of just deleting what was definitely broken. But killing both the rules 12a and 12b might actually be the best thing.

Actually, reviewing all the rules targetting r9750 sdcc for current relevance would be a good job to do too.
But not sure I want to take it on. 🤷

suborb:
> But, killing both the rules 12a and 12b might actually be the best thing. Your call.

I think it probably is.

PR #1812, opened 2021-06-09
added support different origin addresses for application on pps target

witchcraft2001:
In Sprinter DSS, an application can be launched from different addresses, not just $8100 as specified in the original pps_crt0.asm file.


suborb:
Thanks for the contribution.

I didn't know that, though docs on the Sprinter were never very common of course

PR #1832, opened 2021-08-09
build on OpenBSD

jg1uaa:
I tried to build z88dk on OpenBSD, we have to fix these files.

Makefile: OCC variable should be set, regardless using ccache.
build.sh: pwd cannot take arguments. (what this argument means?)
src/ticks/hook_io.c: OpenBSD needs to include <sys/stat.h> explicitly.

After modified these files, use build.sh with some parameters like this:

`CC=egcc CXX=eg++ CPPFLAGS=-I/usr/local/include ./build.sh`
or
`CC=cc CXX=c++ CPPFLAGS=-I/usr/local/include ./build.sh` (cc as clang)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,8 +27,8 @@ CC ?= gcc
 # Prefix for executables (eg z88dk-, hence z88dk-z80asm, z88dk-copt etc)
 CROSS ?= 0
 
+OCC := $(CC)
 ifneq (, $(shell which ccache))
-   OCC := $(CC)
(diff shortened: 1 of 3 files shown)


suborb:
Ah, thank you, it's been a while since I tried on OpenBSD - it's nice to know it still (mostly) works!

> build.sh: pwd cannot take arguments. (what this argument means?)

I've no idea why I did that, it should be `dirname $0` - I'll fix it up after merge.

Issue #1823, opened 2021-07-23
classic - 8085 undocumented opcodes


feilipu:
I'm building a [8085 based CPU board for RC2014](https://feilipu.me/2021/08/15/8085-cpu-on-the-z80-bus/), and I'm planning to use the CA80C85B device at the same clock as standard in the RC2014. This is a device that formally adds the [undocumented opcodes](https://en.wikipedia.org/wiki/Intel_8085#Undocumented_instructions) to the 8085 instruction set.

[CA80C85B](https://github.com/z88dk/z88dk/files/6866198/CA80C85B-TUNDRA.pdf)

[Link to opcodes](https://pastraiser.com/cpu/i8085/i8085_opcodes.html) denoted by  __*__.

I don't think z80asm assembles these instructions when using the `-m8085` option.

Is that correct?
Would it be possible to add them to z80asm, please, if they're not already there?

suborb:
It does understand the Z80 style equivalent opcodes and I’m pretty certain that intel is supported too.

feilipu:
That's good. I would prefer Zilog style.

Is there any source or doc description of what they are in Zilog codes?

pauloscustodio:
Please look up the z80asm test file, https://github.com/z88dk/z88dk/blob/69f7a4d0185827c016ad5a18c2ea0b920af35f06/src/z80asm/dev/cpu/cpu_test_8085_ok.asm

feilipu:
Thanks.

feilipu:
Looking a bit further at this, there seems to be some opportunities in the 8085 support libraries to use the undocumented instructions more extensively.

I've attached some reference documents, an original [Dehnhardt/Sorensen](https://github.com/z88dk/z88dk/files/6875306/UnDoc85.pdf), and an excerpt from [the 8085B datasheet](https://github.com/z88dk/z88dk/files/6875310/8085_Extended_Opcodes.pdf).
[Ken Shriff's blog](https://www.righto.com/2013/02/looking-at-silicon-to-understanding.html) also adds to the story.

Some things to think about (and then do).
- [ ] use the Underflow Indicator (K) flag with 16 bit decrement and `JUI` `JNUI` instructions to manage loops, like `LDIR` emulation, more cleanly. 16 bit decrement flag is set on overflow to -1, not to 0, note to self.
[...]

PR #1836, opened 2021-08-17
8085 specific additions

feilipu:
Following up on #1823 

- [x] split 8085_opt library to support divergence from 8080
- [x] separate current 8085 specifics into separate files

Just feeling my way around the makefile.

__EDIT__  This has turned into a lot of whitespace changes, which makes it hard to review.
The only real outcome should be the creation of separate `8080_opt.lib` and `8085_opt.lib` files.

I'll leave this here to merge, and open a new PR with actual changes.


feilipu:
I can't find any references using the `8085_opt.lib` that would be affected, so I'll just merge this as it has no other effect.

Issue #1390, opened 2020-02-03
Debugging symbols support


suborb:
Breaking out from #1387 here are the tasks that will be needed:

- [ ] Get zsdcc to generate .adb when in c1mode
- [ ] zcc to treat .adb as an output file and concatenate them just before linking
- [ ] Add .adb file generation to sccz80
- [ ] Transform sdcc inline annotations to C_LINE pragmas
- [ ] Make copt skip over C_LINE pragmas
- [ ] z80asm to append link information to .adb file
- [ ] Update ticks to parse .adb file and allow source level debugging

spth:
I guess in the long term, ELF/DWARF is the future. The hc08, s08 and stm8 backends of SDCC already support ELF output with DWARF debug information.

suborb:
Prompted by @desertkun I moved this along a bit in the `feature/sccz80_cdb` branch, I've run out of the steam at the moment.

I've been focussing on sccz80 since previously that wasn't generating anything.

- `--debug` turns on adb information generation for sccz80 and sdcc
- zcc copies the .cdb file generated by sdcc
- For sccz80 adb information is written into symbols so ends up in the .map file with a `__CDBINFO__`. There's URL encoding of the CDB line happening here, with `_` as the replacement for `%`
- Ticks has been updated to read C_LINE information and allows a breakpoint to be set on a C line, there's also some status stuff to print where we are in a file in the prompt.
- Ticks can read source code and show it in the debugger
- I also added watchpoints to ticks because they're useful

The CDB information is extracted from the .map, decoded and then I've just run out of steam for the moment. What needs to be done (I think):

- [ ] C_LINE needs to be extended with scope/level information
- [ ] ticks needs to parse the cdb information and do something useful with it (tempted to lift the sdcc/debugger code TBH)
- [ ] Some scripts to convert between various formats (for emulicious taking the .map and generating an sdcc style cdb file is appropriate), other emulators read .map files, so adding the sdcc cdb into the .map makes sense there

Although adding support to ticks is fairly pointless in itself, it's hoped that it will provide some copy/paste/modify code for other emulators to pick up.

However, as I said I've run out of steam at the moment so I'm going to park this for a bit.

desertkun:
@suborb thank you for your efforts.

desertkun:
Some progress.

[...]

PR #1837, opened 2021-08-17
Ignore duplicate __CDBINFO__, __C_LINE_ / __ASM_LINE_ fixes

desertkun:
While exploring debugging symbols support, I found several issues with debugging symbols generation.
1. When several object files are linked together, debug info about objects that they all include, collide together. See [minimal example](https://github.com/desertkun/z88dk-debug-error-minimum-example). The fix: ignore symbol redefinitions on `__CDBINFO__xxx` symbols.
2. `__C_LINE_N` and `__ASM_LINE_N` instruction do not take into account includes as well. So when an object `main.c` includes `dependency.h` and `dependency.h` happens to introduce `C_LINE` label at line 5, corresponding `C_LINE` label at `main.c:5` could no longer be defined, because such symbol already exists. Solution: add a suffix with url-encoded filename past line number, aka `__C_LINE_10_main.c` except `main.c` is url-encoded, with `%` replaced with `_`. Ticks, surprisingly, ignores anything past number on the symbol name and uses filename comment as a reference anyway.


desertkun:
#1390

suborb:
I’m away at the moment but the general principle is fine by me. 

@pauloscustodio can you review/merge etc?

pauloscustodio:
I'm also away at the moment, and can have a look next week.

desertkun:
@pauloscustodio I would appreciate if you could look at it. I have major ticks rework at hand that needs this to be merged first.

pauloscustodio:
@desertkun , thanks for contributing. Merged.

desertkun:
@pauloscustodio Thanks a lot for merging. I've got a question: I have reworked ticks in [my fork's branch](https://github.com/desertkun/z88dk/commits/ticks_gdbclient) to separate tick's debugger from tick's emulator to allow ticks to have several emulator backends, one of which would still remain tick's internal emulator, and another one would become ticks debugger on top of "gdb client". 

The question is: would you want to look at it in two stages:
1. Refactor of ticks, separate debugger from emulator, no new functionality, introduce concept of an emulator backend, of which there will be only one at that commit
2. Additional emulator backend as gdb client, which could connect to Fuse (to which I have [implementation](https://github.com/desertkun/fuse/commits/gdbserver) but would need to work with Fuse's maintainers to get that in.

Or you're fine with looking at it all at once?

Thing is in development yet, but I want to set a framework I could rely on, since I've already got a conflict. Thanks.

pauloscustodio:
I like the two-step approach, but @suborb , this is your baby... what do you think?

PR #1839, opened 2021-08-22
SFCB structure offset fixed

yottatsa:
As per http://www.seasip.info/Cpm/bdos.html#102, the stamps end up at
`FCB+18Ah (24)`, which is `1 + 8 + 3 + 1 + 11`.
See http://www.seasip.info/Cpm/bdos.html#104 for BCD.

### DIFF
--- a/include/cpm.h
+++ b/include/cpm.h
@@ -76,17 +76,17 @@ struct fcb {
 };
 
 struct sfcb {
-    uint8_t    drive;          /* drive code */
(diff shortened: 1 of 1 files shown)


yottatsa:
[Working code example](https://github.com/yottatsa/cpmutils/blob/eaba7ec924bdfb6bb2447b84ef13403e029ec76d/types.h#L10-L25) for `CP/M Plus` on `+3`.

PR #1845, opened 2021-08-29
RC2014 - 8085 CPU support

feilipu:
Building support for the 8085 CPU using the RC2014 target.
Two subtypes `acia85` and `basic85` are being built.

In progress...


libsrc/_DEVELOPMENT/string/z80/asm__memlwr.asm
```diff
@@ -28,22 +28,33 @@ asm__memlwr:
...
    ld (hl),a
-   
-IF __CPU_GBZ80__
+
+IF __CPU_INTEL__ || __CPU_GBZ80__
    inc hl
    dec bc
+
+IF __CPU_8085__
+   jp NK,loop
```
suborb, line 50: The K and V flags aren't supported by ticks.

We should probably add support in (though working in that bit of ticks is bit like wading through treacle) if we're going to start using them in the library.

feilipu: Yes, I just wanted to get some simple things like setting interrupt status and memory set / copy working so I could use the same start up code for both machines. But then the strings library has some functions with no `__CPU_INTEL__` support, so it would use the Z80 code paths, so I decided to fix that....

And I'm going to need to stop (unravelling a ball of string) somewhere,.. before the terminal support code at least.

So, I can revert the `__CPU_8085__` things in the interim, and just fix `__CPU_INTEL__` issues, if ticks is too much to do right now?

suborb: I think I should be able to do something. http://www.righto.com/2013/02/looking-at-silicon-to-understanding.html seems to defined the behaviour.

feilipu: Thanks.
Fell down a hole of string functions.
Finally, I think they're all done (except the really hard ones).

feilipu:
Walked into this project the wrong way.
I'm closing this PR and will try again, hopefully without needing to touch every file in newlib.

PR #1848, opened 2021-08-31
z88dk-gdb: restore command (upload binary to the machine)

desertkun:
This adds command to deploy a binary onto debugged machine. That way, a debugging session gone wrong, can be just "restored". The second parameter is an optional restore address, when omitted, the `__head` symbol is used from the debugging information.

Also, fixed a bug: `set <register> xxxx` did not work as it sent twice the data.

### DIFF
--- a/src/ticks/debugger.c
+++ b/src/ticks/debugger.c
@@ -118,6 +118,7 @@ static int cmd_out(int argc, char **argv);
 static int cmd_trace(int argc, char **argv);
 static int cmd_hotspot(int argc, char **argv);
 static int cmd_list(int argc, char **argv);
+static int cmd_restore(int argc, char **argv);
 static int cmd_help(int argc, char **argv);
 static int cmd_quit(int argc, char **argv);
 static void print_hotspots();
@@ -144,6 +145,7 @@ static command commands[] = {
     { "hotspot",cmd_hotspot,       "<on/off>", "Track address counts and write to hotspots file"},
     { "list",   cmd_list,          "[<address>]",   "List the source code at location given or pc"},
     { "help",   cmd_help,          "",   "Display this help text" },
+    { "restore",cmd_restore,       "<path> [<address>]",   "Upload binary into machine memory"},
     { "quit",   cmd_quit,          "",   "Quit ticks"},
     { NULL, NULL, NULL }
 };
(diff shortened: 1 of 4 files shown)


src/ticks/debugger.c
```diff
@@ -1038,7 +1040,26 @@ static int cmd_quit(int argc, char **argv)
...
+        address = parse_address(argv[2]);
+    } else {
+        address = symbol_resolve("__head");
+        if (address == -1) {
+            printf("Warning: could not resolve starting address and no address is provided.\n");
+            return 0;
+        }
+    }
 
+    if (bk.restore(argv[1], address))
```
suborb, line 1056: Is it worth taking an argument on the command line that is the binary and restore defaults to using that?

desertkun: I thought about that but z88dk-gdb doesn't take binary as an argument. We might add one just so this command couild use it. Ticks doesn't have this command implemented at all.

suborb: Okay, I'll pick this up when I add the ticks command.

src/ticks/debugger_gdb.c
```diff
@@ -539,6 +544,47 @@ void debugger_step()
...
+
+        addr += read_;
+    }
+
+    fclose(f);
+
+    mem_requested_amount = 0;
+
+    // zero out all registers except for pc
+    struct debugger_regs_t regs;
```
suborb, line 581: As I merged I thought about this bit.

Many of the CRTs expect a sensible stack value when entering - this sets the sp to top of memory which on the +zx isn't a great place - by default it comes in somewhere around the CLEAR value.

desertkun: Could you please fix that by doing `regs.sp = 0xFFFF;`?

desertkun: Oh now I see. Maybe we can find a symbol in debugging info to restore it from?

suborb: I've just pushed a save/restore of sp over this operation. It's not perfect, but it does mean that sp stays roughly in the area it should be.

A user can always set sp manually should it have drifted to far as a result of multiple restores.

suborb: > Oh now I see. Maybe we can find a symbol in debugging info to restore it from?

Normally sp on entry is inherited from BASIC etc - some targets set sp explicitly on entry, for those there's no problems.

PR #1854, opened 2021-09-07
Fix #1852: accept C_LINE inside DEFVARS and DEFGROUP

### LINKED ISSUE #1852: (z80asm) C_LINE directive not supported with DEFVARS block
Slightly odd one this! 

In stdio.h classic has a DEFVARS within #asm block.

When compiling the library with -debug the output is liberally splattered with C_LINE directives, so that we effectively end up with something like this:

```
DEFVARS 0 {
        C_LINE  126,"stdio.h"
    fp_desc         ds.w    1

}
```
Which reports syntax errors.

pauloscustodio:


desertkun:
Do you want maybe squash those two commits down to one, and them force-push this branch?

pauloscustodio:
> Do you want maybe squash those two commits down to one, and then force-push this branch?
Will do tonight.

PR #1858, opened 2021-09-15
Fix #1856 - store strings in object file as word-counted strings

### LINKED ISSUE #1856: (z80asm) Can't generate labels with > 255 characters
The debugging work that's going uses defc constants to encode the CDB debug information.

Unfortunately it looks like when we start to instrument structures, the length could exceed 255 fairly easily.

When this happens we end up with the fatal message: "too long for byte counted string".

@pauloscustodio is there any chance we could shift to word counted strings?

pauloscustodio:


src/z80asm/t/issue_341.t
```diff
@@ -40,18 +40,14 @@ my $map = join("\n", grep {/test.c:|test1.asm:/} split('\n', slurp("test.map")))
 check_text($map, <<'END', "map file contents");
 __C_LINE_0_test_2ec             = $0000 ; addr, local, , test_c, , test.c:0
 __C_LINE_1_test_2ec             = $0000 ; addr, local, , test_c, , test.c:1
-__C_LINE_2_test_2ec             = $0000 ; addr, local, , test_c, , test.c:2
```
desertkun, line 43: Are you sure removing these was an okay behavior? Why these lines no longer appear? A they empty?

pauloscustodio:
> Are you sure removing these was an okay behavior? 

The lines are generated by zcc by inserting C_LINE directives in the asm source files. This is an active area of development at the moment and I expect the results to vary.

PR #1863, opened 2021-09-22
rc2014-8085 - minimal library

feilipu:
A minimal set of changes to have an RC2014-8085 library and subtypes.


lib/config/rc2014.cfg
```diff
@@ -11,17 +11,23 @@ CRT0     DESTDIR/lib/rc2014_crt0
 OPTIONS  -O2 -SO2 -iquote. -D__Z88DK -D__RC2014 -D__Z80 -clib=sdcc_iy -subtype=basic
 
 CLIB     default -lrc2014_clib -lndos -Cc-standard-escape-chars
+
+CLIB     rc2014-8085 -Cc-standard-escape-chars -D__SCCZ80 -D__8085 -D__8085__ -l8085_opt -lndos -lrc2014-8085_clib
```
suborb, line 15: I think this needs to have -startuplib=8080_crt0 ?

feilipu: I'll add it. I nearly threw all my toys out of the cot this afternoon, but perhaps some progress now.

suborb:
As a heads up, there's some faults in the 8085 rules file - an aes256 example generates incorrect results.

feilipu:
Ok. I'm keeping the PR synced with master in hope to avoid issues.

Still struggling with the best crt0 code. But it's painful to match newlib style to classic library functions. Will try classic style today, and see whether that works better.

suborb:
> Still struggling with the best crt0 code. But it's painful to match newlib style to classic library functions. Will try classic style today, and see whether that works better.

That sounds curious. What functions are missing?

feilipu:
>> Will try classic style today, and see whether that works better.

> That sounds curious. What functions are missing?

The last thing to fix was missing the definitions for [`fputc_cons`](https://github.com/z88dk/z88dk/blob/rc2014-8085/lib/crt/classic/crt_runtime_selection.asm#L513) if the classic runtime selection code is not used.

I'm now at a place where all the code pieces are compiled / linked and ending up in plausible locations.
I have a `helloworld.c` program that just returns though rather than printing appropriately. So something's still amiss.

I'm not sure about the heap creation and the strings / stdio setup, and whether I've mangled something bridging between newlib and classic. If you've a (long) moment, I'd appreciate any feedback on the [PR](https://github.com/z88dk/z88dk/tree/rc2014-8085) please.

suborb:
Of course I'll take a look.

 I'll also update ticks to support running generated files: I think most of the bits are there, there's a simple emulated acia on 0x80 and 0x81 which is hopefully enough.

feilipu:
I've just done Basic85 for time being. Using RST interfaces and $9000 origin.

suborb:
I did have one of your MSBASIC builds running on ticks once: I think when we were trying to sort out some FP issue.

My laptop has ceased to exist this week (I think the thermal paste has failed and cooked it)  so I've probably lost that work, however I suspect it won't needed.

libsrc/_DEVELOPMENT/stdlib/z80/asm_atoi.asm
```diff
@@ -32,12 +32,12 @@ asm_atoi:
 asm0_atoi:
```
suborb, line 32: The classic 8080 uses the atoi that's in libsrc/stdlib/8080/asm_atoi

If the newlib one has now been adapted we can switch over.

libsrc/_DEVELOPMENT/target/clib_variables.inc
```diff
@@ -38,27 +38,55 @@ IF __clib_malloc_heap_size = -1
       
    ld hl,-1                    ; ensure sp == 0 is accepted
    add hl,sp                   ; hl = top of the stack
-   
+
+IF __CPU_INTEL__ || __CPU_GBZ80__
```
suborb, line 42: We could make this cleaner by always using bc as __BSS_END_tail the the CPU variation is just on the subtraction

libsrc/_DEVELOPMENT/target/clib_variables.inc
```diff
@@ -38,27 +38,55 @@ IF __clib_malloc_heap_size = -1
...
    sbc hl,de                   ; hl = space between top of stack and end of bss - 1
+ENDIF
+
+   jp C, __Exit                ; if stack overlaps bss, this is a fatal error
 
-   jp c, __Exit                ; if stack overlaps bss, this is a fatal error
-   
    ld bc,__crt_stack_size + 14 ; amount of space reserved by stack + minimum heap size - 1
+
+IF __CPU_INTEL__ || __CPU_GBZ80__
```
suborb, line 63: To avoid the nesting maybe:

IF __CPU_8085__

ELIF __CPU_GBZ80__ | __CPU_8080__

ELSE
  ;; z80 path
ENDIf

libsrc/_DEVELOPMENT/target/crt_program_exit.inc
```diff
@@ -37,10 +37,17 @@
             
                ; program jumps to caller
 
+              IF __CPU_INTEL__
```
suborb, line 40: Is this not emulated by z80asm?

libsrc/_DEVELOPMENT/target/clib_variables.inc
feilipu: Thanks.. didn't know ELIF was accepted. I didn't notice it being used.
There's quite a bit of nesting needed otherwise.

libsrc/_DEVELOPMENT/target/crt_program_exit.inc
feilipu: I don't know. Not one I asked for. I'll check.

feilipu (commented):
Working on amalloc at the moment. Needs some adaption too.

libsrc/_DEVELOPMENT/target/clib_variables.inc
feilipu: I am not sure that the contents of `bc` aren't otherwise needed.
But there's definitely a better cleaner way to do it, which I didn't do (trying to not affect other code paths).

libsrc/_DEVELOPMENT/target/crt_program_exit.inc
feilipu: As far as I can see `ld de,(**)` is not emulated for INTEL or GBZ80.
It throws a warning, and I can't otherwise find it.

@pauloscustodio Could it be added to z80asm, please?

suborb:
I'd already done the work on ticks. Name the binary so it has rc2014 in it and then:

```
z88dk-ticks -m8085 -pc 0x9000 -x a.map -debug rc2014.bin
```

It patches the rsts directly for console in/out.

feilipu:
I'll leave the `-subtype=acia85` work for later, and merge this now.
It is working well enough to move forward, I think.

PR #1889, opened 2021-10-21
Added bank support for CPC

iratahack:
Added memory bank support for the CPC6128. Similar to the ZX +3 target, bank binary images are named &lt;filename&gt;_BANK_n.bin and &lt;filename&gt;_CODE.bin (the main program) as generated by zcc. Although, crt0 doesn't currently support memory banks for the CPC target zcc can be used to link assembly code which does.

.wav file support still only supports applications which contain a single binary (the main program).


iratahack:
@suborb , By the way, I also have changes to appmake to add the same bank support to .wav file generation. I can push them now or hold off until you've had chance to digest the current changes to appmake.

PR #1895, opened 2021-11-05
Handle nested include files.

iratahack:
Without this change, only the first level of include files are flattened
and the file name for the nested INCLUDE directives are converted to hex
which causes the assembler to barf.


pauloscustodio:
Thanks for contributing, @iratahack

PR #1896, opened 2021-11-10
rc2014 - 8085 newlib device and driver files

feilipu:
- Provide 8085 compatible device and driver files.
- Provide compatible C linkage for sccz80.


suborb:
Random question: If I understand properly, these aren't just purely for rc2014, they could be run on any target that has the hardware.

So, is it worth moving these out of the rc2014 into drivers or similar?

feilipu:
I'm concentrating on drivers 3 pieces of hardware which are available for the rc2014, to build 8085 versions and then link them to sccz80 C wrappers.
- 82C55 PPI (as an IDE interface)
- 6850 ACIA Serial
- SIO/2 Serial

The drivers are fairly generic, and use port constants where possible to make them more generic. But, I don't know how useful they'd be to general systems. For example, the yaz180 uses 16-bit port addressing so the IDE driver for that won't work for rc2014. But, it isn't an 8085 machine anyway, so mute point.

I sort of think that there'd be limited usefulness at this stage, given there's only one 8085 target...
Unless you have another plan?

suborb:
Nothing planned at the moment, but I do know that: https://z88dk.org/forum/viewtopic.php?f=12&t=11600 is hopefully incoming which uses another serial chip of course!

It was just a sense that the serial chip code in particular is probably re-usable across many targets though of course I can't find anything at the moment and as you say, it may not actually be worth it since they're pretty thin.

feilipu:
It looks like (from a quick read), that there's a different serial chip, and possibly it can use the same [SD interface code written for the scz180](https://github.com/feilipu/z88dk-libraries/tree/master/diskio_sd) (compiled for 8085 of course).

Related, once I've got these drivers adapted, I'll need to find a way to discriminate for adding 8085 `z88dk-lib` files. I'm ok to copy them over manually currently though.

feilipu:
Pretty happy to integrate 8085 and z80 device driver code, with a better outcome for z80 too.

Looking back at my old code it was easy to see that I had no idea what I was doing.
I think it is slightly improved now, but still learning.

PR #1901, opened 2021-12-09
fix for `asm_zx_saddrcright` and zx screen address optimisations

ped7g:
I did verify functionality of code only externally with sjasmplus in my custom tests, but I didn't try to actually build it with z88dk z80asm (as I don't use z88dk personally).

It *should* work, but I have no idea if you have these under automated testing or not, if not, maybe doing sanity check before merge by someone using z88dk would be good idea, just in case.

### DIFF
--- a/libsrc/_DEVELOPMENT/arch/ts2068/display/z80/asm_tshc_pxy2aaddr.asm
+++ b/libsrc/_DEVELOPMENT/arch/ts2068/display/z80/asm_tshc_pxy2aaddr.asm
@@ -19,13 +19,6 @@ EXTERN asm0_zx_pxy2saddr
 
 asm_tshc_pxy2aaddr:
 
-   ld a,h
-   and $07
-
-IF __USE_SPECTRUM_128_SECOND_DFILE
-   or $e0
-ELSE
-   or $60
-ENDIF
+   scf
 
    jp asm0_zx_pxy2saddr
--- a/libsrc/_DEVELOPMENT/arch/zx/display/z80/asm_zx_pxy2aaddr.asm
+++ b/libsrc/_DEVELOPMENT/arch/zx/display/z80/asm_zx_pxy2aaddr.asm
@@ -30,22 +30,19 @@ asm_zx_pxy2aaddr:
    srl l
 
    ld a,h
-   rlca
-   rlca
-   ld h,a
-   
+
+IF __USE_SPECTRUM_128_SECOND_DFILE
+   ld h,$d8/4
+ELSE
+   ld h,$58/4
+ENDIF
+
+   rla
+   rl h
+   rla
+   rl h
+
    and $e0
    or l
    ld l,a
-   
-   ld a,h
-   and $03
-
-IF __USE_SPECTRUM_128_SECOND_DFILE
-   or $d8
-ELSE
-   or $58
-ENDIF
-
-   ld h,a   
    ret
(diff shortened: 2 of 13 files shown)


suborb:
These functions aren't under auto-test unfortunately - if you know of a test suite I can adapt let me know though.

I'm understandably nervous about changes to these functions since they're core and have been problematic in the past.

ped7g:
well. I just ran them against original versions in assembly, going through all possible inputs to make sure the result is identical.

I have no idea how you use these from C (especially the carry flag return) and how you build your tests to be automated and frankly I'm not interested to learn it, as I personally do asm-only stuff and already happened to maintain one assembler project.

I can write some pseudo-C variant of each function to provide test values, but somebody familiar with z88dk would have to turn them into working test and call the asm variants properly.

For example the main one `asm_zx_pxy2saddr` is like:
```
// C function to calculate screen address from pixel [x,y] coordinates
uint16_t c_zx_pxy2saddr(uint8_t x, uint8_t y) {
    uint16_t adr = __USE_SPECTRUM_128_SECOND_DFILE ? 0xc000 : 0x4000;
    adr += (y << 5) & 0x1800;
    adr += (y << 8) & 0x0700;
    adr += (y << 2) & 0x00e0;
    adr += (x >> 3);
    return adr;
}
// TODO test loop ... something like
/*
  // doing full 256x192 test which may take some while if running at only 3.5MHz Z80
  for (uint8_t y = 0; y < 192; ++y) {
    uint8_t x = 0;
    do {
      uint16_t expected = c_zx_pxy2saddr(x,y);
      uint16_t asm_adr = ???;
      if (expected != asm_adr) error...
    } while (++x);  // end when 255->0
  }
  ok...
*/
```
etc...

ped7g:
@suborb 
> I've no idea what the tshc functions is meant to do, looks totally
wrong to me.

The Timex "US ZX" (model 2068?) has "hicolor" and "hires" modes, where the classic ZX bitmap-vram-region at 0x4000..0x57FF is used together with second region 0x6000..0x77FF. The second region is used for attributes in "hicolor" mode and for another 8 pixels in "hires" mode (interleaving per 8 pixels), with the same memory layout as original 0x4000 region.
(and the extra 0x6000..0x7aff region can be used also for double-buffering of classic ZX 256x192 mode with 8x8 attributes, this is different double-buffering from ZX128, having both screens in Bank 5)

So for example in "hicolor" mode (attributes are then 8x1, not 8x8) for position [8,10] the pixel byte is read from 0x4221 and attribute byte is read from 0x6221 (the difference is just in top three bits %010 vs %011).

The routine name using "aaddr" suggest this is probably meant for "hicolor" mode, but technically can be used to calculate address in the second region for any of those Timex modes, layout is identical. (again, I'm missing the full context how this is used from C, so I can provide only the low-level asm style info and how the HW works, no insight into the C interface level).

suborb:
> The Timex "US ZX" (model 2068?) has "hicolor" and "hires" modes, where the classic ZX bitmap-vram-region at 0x4000..0x57FF is used together with second region 0x6000..0x77FF. The second region is used for attributes in "hicolor" mode and for another 8 pixels in "hires" mode (interleaving per 8 pixels), with the same memory layout as original 0x4000 region.

Yes, I remembered that in the end (reversing the tshc acronym helped for me!). 

It's an unfortunate downside of working on so many targets that you tend to forget the fine details after a while: eg I remember that +kc has an unusual screen layout, but I couldn't tell you what it is without looking at code/docs.

Anyway, I can confirm that these routines do match on a bc, de, hl, carry flag basis so thanks for the contribution.

Issue #1683, opened 2021-01-25
asmpp.pl does not correctly interpret escaped characters in quotes


iratahack:
asmpp.pl converts strings in quotes to comma separated values.

Escaped characters such as "\n" or "\r" are converted to two comma separated values, i.e. "\n" becomes db  92,110 instead of db  10

PR #1903, opened 2021-12-11
Addresses issue #1683

iratahack:
Basic escaped characters are converted to ASCII before converting to decimal.


pauloscustodio:
Thanks

PR #1905, opened 2021-12-13
Included \b and \e in escape character processing

iratahack:


pauloscustodio:
@iratahack , thanks for contributing

PR #1907, opened 2021-12-16
A more CMake friendly asmpp.pl

iratahack:
Command line processing:
 - Ignore -c which is passed in by CMake
 - Remove the space between -o and the filename passed in by CMake
 - Added -v to enable display of @cmd
 - Added --ucase to command line usage text

Include file processing:
 - #include can now be preceded by white space


pauloscustodio:
@iratahack, thanks for contributing

Issue #1911, opened 2021-12-22
intel / gbz80 - optimise copy / set loops


feilipu:
For the Intel / GBZ80 newlib/classic implementations the implementation of copy / set loops could be improved.

eg. [asm_memcpy](https://github.com/z88dk/z88dk/blob/master/libsrc/_DEVELOPMENT/string/z80/asm_memcpy.asm#L86), and [asm_memset](https://github.com/z88dk/z88dk/blob/master/libsrc/_DEVELOPMENT/string/z80/asm_memset.asm#L82).
There are many other examples.

In z80asm eg. [_z80asm_ldir](https://github.com/z88dk/z88dk/blob/master/src/z80asm/dev/z80asm_lib/ldir.asm#L10).

The copy / set loop currently looks like:
```asm
.asm_ldir
    ;
.loop
    ; stuff to copy / set etc
    dec bc        ; 6
    ld a,b        ; 4
    or c          ; 4
    jr NZ,loop    ; 10
```
making loop cost 24 cycles per byte or total __6144__ cycles per page.

An alternative implementation looks like:
```asm
.asm_ldir
    ld a,c         ; 4
    dec bc         ; 6
    inc b          ; 4
    ld c,a         ; 4
.loop
    ; stuff to copy / set etc
    dec c          ; 4
    jr NZ,loop     ; 10
    dec b          ; 4
    jr NZ,loop     ; 10
```
making loop cost 14 cycles per byte or __3616__ cycles per page (or less).

@suborb If you're ok with the change, I can do a PR across the library functions.

But @pauloscustodio, would you be able to do the same for z80asm please?
Not sure I can get all the build tests etc right.

ped7g:
the prologue can be:
```
  dec bc  ; 6
  inc b  ; 4
  inc c  ; 4
```
-4T and doesn't clobber A

suborb:
@feilipu go for it

pauloscustodio:
This is a space/time tradeoff. The proposed solution is faster but costs 5 more bytes.

Maybe we can keep two versions of the library functions (e.g. asm_ldir and asm_ldir_fast), call the shorter routine by default and call the faster routine if the z80asm option -opt-speed is passed (this option today only replaces JR by JP).

To build the whole system with the option without changing the build scripts, the Z80ASM environment variable can be used, e.g.
```
Z80ASM=-opt-speed make
```

feilipu:
[...]

PR #1912, opened 2021-12-23
intel / gbz80 - optimise copy / set loops

feilipu:
For #1911


feilipu:
Not sure how deeply the `asm_mem*` functions are exercised by the test suite?

PR #1917, opened 2021-12-28
Implement macros

pauloscustodio:
z80asm is now a macro assembler!

Implemented MACRO/ENDM, REPT/ENDR, REPTC/ENDR, REPTI/ENDR, EXITM, LOCAL, DEFL, #define, #undef

Documentation at https://github.com/z88dk/z88dk/wiki/Tool---z80asm---preprocessor


suborb:
That's one way to poke me to upgrade the nightly environment! 

@pauloscustodio, could you update this page: https://github.com/z88dk/z88dk/wiki/installation with the compiler requirements?

pauloscustodio:
Done

suborb:
With luck we should get a nightly tonight. It's being built on an completely new machine, so I've no idea if the windows build will actually work.

suborb:
@pauloscustodio would it be possible to check in the generated files need to build the z80asm binary? 

I've sorted out the snap build, but docker is based on alpine and pulling in perl and all the deps there seems like a pain.

pauloscustodio:
The generated files are checked in. What errors are you getting?

pauloscustodio:
@suborb, I will have a look

pauloscustodio:
Opened a new issue #1922

suborb:
Sorry I fell asleep!

 I was asking whether it was possible to check in the output of the template commands - that way for a build only (no tests) Perl isn’t needed

pauloscustodio:
The problem is solved for now with #1922. When I get to work on removing the .err files, this issue will disappear.

PR #1933, opened 2022-01-03
Ensure nextreg immediates are processed in the correct order

iratahack:
Variable arguments to BUF_PRINTF seem to be processed in reverse order causing the 2 immediates for nextreg n,n' to be displayed in reverse order. This change forces the 2 immediates to be displayed in the same order as z80n memory.


suborb:
Apologies for missing this one - this issue is bit of a pitfall when developing on a Mac 

https://z88dk.org/forum/viewtopic.php?p=19942#p19942

PR #1935, opened 2022-01-04
allow redefinition of CRT_ORG_CODE

nippur72:
(I'm not sure this is the best way to do it) this allows redefinition of `CRT_ORG_CODE` in order to change the starting address of the program via compiler switch `-pragma-define`. I tried `-zorg` but it didn't work.

### DIFF
--- a/lib/target/sol20/classic/sol20_crt0.asm
+++ b/lib/target/sol20/classic/sol20_crt0.asm
@@ -41,7 +41,7 @@
 	defc	CONSOLE_COLUMNS = 64
         INCLUDE "crt/classic/crt_rules.inc"
 
-     IFNDEF CRT_ORG_CODE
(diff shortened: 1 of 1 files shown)


lib/target/sol20/classic/sol20_crt0.asm
```diff
@@ -41,7 +41,9 @@
 	defc	CONSOLE_COLUMNS = 64
         INCLUDE "crt/classic/crt_rules.inc"
 
+     IFNDEF CRT_ORG_CODE
```
suborb, line 44: It looks like we normally do this with:

```
IF !DEFINED_CRT_ORG_CODE
```

Not that it really matters either way...

nippur72: ok I've changed it to `IF !DEFINED_CRT_ORG_CODE`

PR #1939, opened 2022-01-08
ticks: complex compressions evaluation with “print” (or “p”)

desertkun:
This change adds implementation to the `print` command of ticks, which was unimplemented prior to this.
It uses flex/bison to parse expression according to `expressions.l`/`expressions.y` and with help of `exp_engine.c` becomes able to do C stuff like referencing, de-referencing, pointer math, type cast, etc.

**How to test it**
- Pull this example project: https://github.com/desertkun/z88dk-debug-error-minimum-example
- compile it with make
- run `z88dk-ticks -x example.map -d -l 0x8000 -pc 0x8000 example__.bin`
- but a breakpoint over method "check" `b check`
- continue `c`
- `info locals`
- `p kek`
- `p test`
- `p (uint8_t)1`
- `p *kek + 10`
- `p (uint8_t*)test`
- etc

**Limitations**
- Structs are not yet supported, only primitive types
- No pointers of pointers
- ~~Only local variables can be printed out as of now, but that's not a limitation of evaluation but rather inability to obtain global variables at all~~

### DIFF
--- a/src/ticks/exp_engine.c
+++ b/src/ticks/exp_engine.c
@@ -359,6 +359,9 @@ int expression_result_value_to_string(struct expression_result_t* result, char* 
         case TYPE_UNKNOWN: {
             return snprintf(buffer, buffer_len, "<unknown>");
         }
+        case TYPE_FUNCTION: {
+            return snprintf(buffer, buffer_len, "<function>");
+        }
         case TYPE_ARRAY: {
             int maxlen = Max(10,Min(10, result->type.size));
             int offs = snprintf(buffer, buffer_len, "%#04x [%d] = { ", result->as_pointer.ptr, result->type.size);
@@ -366,7 +369,7 @@ int expression_result_value_to_string(struct expression_result_t* result, char* 
             for ( int i = 0; i < maxlen; i++ ) {
                 offs += snprintf(buffer + offs, buffer_len - offs, "%s[%d] = ", i != 0 ? ", " : "", i);
                 struct expression_result_t elr = {};
-                debug_resolve_expression_element(&result->type, RESOLVE_BY_POINTER, ptr, &elr);
+                debug_resolve_expression_element(&result->type, result->type.first, RESOLVE_BY_POINTER, ptr, &elr);
                 if (is_expression_result_error(&elr)) {
                     offs += snprintf(buffer + offs, buffer_len - offs, "<error:%s>", elr.as_error);
                     break;
@@ -439,7 +442,7 @@ int expression_result_value_to_string(struct expression_result_t* result, char* 
             }
         }
         default: {
-            return 0;
+            return snprintf(buffer, buffer_len, "<unknown:%d>", result->type.first->type_);
         }
     }
     return 0;
--- a/src/ticks/exp_engine.c
+++ b/src/ticks/exp_engine.c
@@ -39,56 +39,42 @@ uint8_t is_expression_result_error(struct expression_result_t* result) {
 void set_expression_result_error(struct expression_result_t* result, const char* error) {
     result->flags |= EXPRESSION_ERROR;
     strcpy(result->as_error, error);
+    expression_result_free(result);
 }
 
 void reset_expression_result(struct expression_result_t* result) {
     result->flags = EXPRESSION_UNKNOWN;
+    expression_result_free(result);
 }
 
-void expression_result_free_typechain(struct expression_result_t* result)
+void expression_result_free(struct expression_result_t* result)
 {
-    type_chain* next = result->type.next;
-    while (next) {
-        type_chain* next_next = next->next;
-        free(next);
-        next = next_next;
+    if (result->type.first) {
+        free_type(result->type.first);
+        result->type.first = NULL;
     }
 }
 
-static uint8_t is_expression_a_pointer(struct expression_result_t *exp) {
-    return exp->type.type_ == TYPE_GENERIC_POINTER ||  exp->type.type_ == TYPE_CODE_POINTER;
-}
-
-uint8_t is_expression_signed(struct expression_result_t *exp) {
-    return exp->flags & EXPRESSION_TYPE_SIGNED;
-}
-
-void set_expression_signed(struct expression_result_t *exp, uint8_t is_signed) {
-    if (is_signed) {
-        exp->flags |= EXPRESSION_TYPE_SIGNED;
-    } else {
-        exp->flags &= ~EXPRESSION_TYPE_SIGNED;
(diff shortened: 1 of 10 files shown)


src/ticks/expressions.l
```diff
@@ -0,0 +1,78 @@
...
+#include "exp_engine.h"
+#include "expressions.tab.h"
+
+int lookup_word();
+int lookup_primitive_type();
+
+%}
+
+word [a-zA-Z_][a-zA-Z0-9_]*
+primitive_type "char"|"unsigned char"|"int8_t"|"uint8_t"|"int16_t"|"uint16_t"|"int32_t"|"int"|"uint32_t"|"float"
```
suborb, line 17: +long, long long etc?

suborb:
I'm a bit bogged down in sorting some graphics problems, but I really want to give this a try - give me a couple of days and then I'll be able to focus.

desertkun:
Thanks, one thing I noticed is apparently global symbols have zeros as addresses, even though they're in the E memory space. I tried to also add code to print global variables, but their location is uknown at this time.

src/ticks/expressions.l
desertkun: Resolved by not having a separate token for types at all. Any word is first tried to be a type now (see `lookup_word`)

desertkun:
A was able to make it work to ticks to display global variables, as well.



The `info` section was refactored as well, to accept `breakpoints` (for gdb compatibility) and now `locals`/`variables`, which can be shortened during calls, e.g. `i v`

desertkun:
Kind of rewrote the thing, now instead of having my own home brew type system, am reusing `type_chain`. One step from being able to decode structs.
```
$8282    (_check+0)> i l
  <char*>test = 0x8bf2("hello!")
  <struct test_t*>kek = 0xffb9
  <int16_t>value = 50

$8282    (_check+0)> bt
 * function check+0 (<char*>test = 0x8bf2("hello!"), <struct test_t*>kek = 0xffb9, <int16_t>value = 50)
       at main.c:12
   function main+32 (<int16_t>h = -29710, <struct test_t>v = {0xffb9})
       at main.c:22
```

src/ticks/exp_engine.c
```diff
@@ -0,0 +1,517 @@
...
+                case TYPE_LONG:
+                {
+                    struct expression_result_t local = {};
+                    expression_dereference_pointer(result, &local);
+                    char buff[128];
+                    expression_result_value_to_string(&local, buff);
+                    sprintf(buffer, "%#04x(%s)", result->as_pointer.ptr, buff);
+                    break;
+                }
+                case TYPE_CHAR: {
```
suborb, line 405: This one needs to isprint() the character or we will end up trashing the terminal

src/ticks/exp_engine.c
```diff
@@ -0,0 +1,517 @@
...
+        case TYPE_LONG: {
+            return 1;
+        }
+        default: {
+            return 0;
+        }
+    }
+}
+
+void expression_result_value_to_string(struct expression_result_t* result, char* buffer) {
```
suborb, line 368: debug.c::debug_print_element has got very similar printing logic - can we have just one copy of the code?

src/ticks/exp_engine.c
```diff
@@ -0,0 +1,517 @@
...
+            expression_result_type_to_string(&from->type, is_expression_signed(from), tp);
+            sprintf(to->as_error, "Cannot dereference type (not implemented): %s", tp);
+            break;
+        }
+    }
+}
+
+type_chain expression_string_get_type(const char* str, uint8_t* is_signed) {
+    type_chain result = {};
+    if (strcmp(str, "void") == 0) {
```
suborb, line 135: I think this might be better/more easily maintained as an array of structs:

struct {
   char *typename;
   char.  assigned;
   enum type_record_type;
} types[] = {
  { "void", 0, TYPE_VOID },
  ...
};

And just iterating through - will make it easier to add the 64 bit/float types

suborb:
> Thanks, one thing I noticed is apparently global symbols have zeros as addresses, even though they're in the E memory space. I tried to also add code to print global variables, but their location is uknown at this time.

Sorry I'm catching up....for globals you can just get the address of label. Function statics may be slightly more complicated of course.

desertkun:
Thanks for comments, I'll address those. As for globals, I'll resolved that.

src/ticks/exp_engine.c
desertkun: Sadly, no, as expression_result_t is a separate entity. I can convert whatever in a `debug_print_element` by calling `debug_resolve_expression_element` into an `expression_result_t` first and then call `expression_result_value_to_string`, what do you think?

src/ticks/exp_engine.c
desertkun: Thanks.

src/ticks/exp_engine.c
desertkun: Ugh, `debug_print_element` has this offset thing going on that really complicates things...

suborb: Ah yeah. Makes me think we should use UT_string? It’s just trying to build up a buffer and not overflow ot.

desertkun: Let me combine the two somehow

desertkun: Ok, figured how make it to work. Wasn't able to test array thoroughly, because CDB generates a line which parser doesn't like (character `x` in particular)
```
Warning cannot parse type record.
Warning: could not add debug info on symbol.
Decoded cdb: <S:Fmain.c$haha55$0_0$0({4}DA2x,SI:S),E,0,0>
```
I think I should rewrite the parser anyway using yacc/bison, and fix arrayys there

desertkun:
I not sure I like the approach I did with `EXPRESSION_TYPE_SIGNED` flag there, but I had my reasons, mainly `struct type_record_s` has `type_chain* first` as a pointer, which means even the first type has to be allocated, which I don't like doing, as I'd prefer `type_chain` to be a part of value structure. Maybe I can change that to `type_chain first` instead, and follow it in "we guarantee we have first, others not so much" kind of patter?

desertkun:
OK I ditched `EXPRESSION_TYPE_SIGNED` flag and instead relying on `type_record` now, except for some freeing I had to do. But I've put a catch around leaks and it seems I have caught them all.

suborb:
Such a great feature - let me sort out the build mess I created earlier this evening, and I'll merge this weekend.

Your work on this is really appreciated

Issue #1944, opened 2022-01-10
am9511 - rework for 8085 and sccz80


feilipu:
I'm planning to rework the am9511 math library to achieve following.

- [ ] make sccz80 first class (remove parameter stack swapping overhead, etc) - wasn't much of this going on actually.
- [ ] remove the abstraction layer for file linking (confusing) - ain't broke - won't fix.
- [x] support Intel 8085 (8080 easy addition as needed).
- [x] reduce z80 support to "traditional" 8-bit I/O
- [x] reduce to `long` and `float` handling (remove `int`)
- [x] assume that "old" sdcc stack parameter passing will be retained

feilipu:
Initial testing with RC2014-8085 using [classic Whetsone](https://github.com/z88dk/z88dk/tree/master/libsrc/_DEVELOPMENT/EXAMPLES/benchmarks/whetstone/z88dk-classic) benchmark.
 - 8085+MBF32 -> 78.2 Seconds.
 - 8085+AM9511 ->30.4 Seconds.
```
$ zcc +cpm -clib=8085 -vn -O2 -DSTATIC -DPRINTOUT whetstone.c -o whetstone --math-am9511_8085 -lndos -create-app
$ zcc +cpm -clib=8085 -vn -O2 -DSTATIC -DPRINTOUT whetstone.c -o whetstone --math-mbf32_8085 -lndos -create-app
```

PR #1945, opened 2022-01-11
am9511 - 8080 / 8085 rework

feilipu:
Related to #1944.


libsrc/_DEVELOPMENT/math/float/am9511/c/8085/am9511_asinh.asm
```diff
@@ -185,41 +185,37 @@
...
+	call	l_f32_add
 	call	_sqrt
-	call	dpush
-	ld	hl,12	;const
+	push	de
+	push	hl
+	ld	hl,8	;const
 	add	hl,sp
-	call	dload
+	call	l_glong
```
suborb, line 211: Shouldn't this be `call fabs` since `fabs()` is a sccz80 library function?

feilipu: TBH, I'm playing with the [makefile](https://github.com/z88dk/z88dk/blob/8827c31cb7847d249f71d274748766590664330f/libsrc/_DEVELOPMENT/math/float/am9511/c/Makefile#L14) to try to get something that looks sensible for use with sccz80/classic/8085. This was the first cut.

At the moment it is compiling against newlib. I know that's incorrect, but I haven't got the right magic to pull in the classic header files and make sure it is using sccz80 to compile. Hints / suggestions on how to get something sensible are welcome. `make` is not my friend.

suborb: Ah right, yeah, `zcc +test -clib=8085` should do it for you.

feilipu:
Perplexed by these build errors in the DAI math library. :confused: 
Haven't touched that to my knowledge.

suborb:
> Perplexed by these build errors in the DAI math library. 😕
> Haven't touched that to my knowledge.

My thoughts:

1. daimath uses `math/cimpl/pow_valid.c` to implement `pow()
2. math.h defines pow() to be `pow_callee()`
3. So pow_valid.c ends up exporting` pow_callee()` not `pow()`

feilipu:
> Probably a `#undef pow` needed in `pow_valid.c`

I've matched the classic math headers calling convention for am9511 and math32 to the newlib am9511 and math32 headers.
Does DAI math use either of these two classic math header files?

That's the only place I can guess that could cross over...?

But, otherwise please let me know if I should change `pow_valid.c` (since I don't want to touch something that I don't really understand).

suborb:
Ah, you're right actually - it should be using math_dai32.h, but it's not.

Take a look at the Makefile for daimath32 and you'll see why. It needs to be `--D__MATH_DAI32` not `-D__MATH_AM9511`

feilipu:
Ah. Ok.
Will fix in next push the am.

feilipu:
Chasing down an issue with `%f` format conversion (something to do with the `ftoa.c` helpers).
Pretty sure there's a calling convention mismatch somewhere that is causing a stack problem.

feilipu:
Now an off by 1 in the `%f` format conversion. Or it might be a trashed register.
Could be the last thing.

feilipu:
Fixed the `%d` issue, where the math library `l_long_div_u` doesn't provide the modulus in `a`.
But the `%f` issue remains...

feilipu:
Adding the specific  `l_long_div_m` function, which returns the modulus LSB in `a`, fixes the issue with `%d` and `__print_number`. But it creates an issue where an actual `(unsigned long)x / (unsigned long)y` calculation is required.

There ends up being a double definition of `l_long_div_u`, one from the `l_` source, and one from the `am9511` math source.

Any suggestion on the cleanest way to fix this duplication, while enabling use of the math library option if available?

feilipu:
I'm at a loss now as to why the`%f` format conversion isn't working. Confirmed that `floor()` is working right.
The binary results are right. Just not the print.

Last thing is the calling convention for ftoa (but I fixed that already, I thought :confused: ).
```
floor(2.1666665077) should be 2.0000000000 and we got 2.0000000000
floor(0x3fc00000) should be 0x3f800000 and we got 0x3f800000

floor(-3.8870081901) should be -4.2500000000 and we got -4.2500000000
floor(0xc0670a3d) should be 0xc0800000 and we got 0xc0800000

floor(1000.0110473632) should be 1000.0009765625 and we got 1000.0009765625
floor(0x447a00a5) should be 0x447a0000 and we got 0x447a0000

floor(-1000.0110473632) should be -1001.0009765625 and we got -1001.0009765625
floor(0xc47a00a5) should be 0xc47a4000 and we got 0xc47a4000
```

PR #1946, opened 2022-01-11
Arm64 fixups

suborb:
- Various ARM64 casting fixups - still working on x86_64


suborb:
Working on:

* x86_64
* i386
* arm64

I think that's enough, arm32 always had issues with 64 bit numbers.

PR #1948, opened 2022-01-16
ticks: breakpoint work

desertkun:
Small quality of life improvements:
1. `delete <id>` or `d <id>` to be more compatible with regular gdb
2. `delete` or `d` will ask to delete all breakpoints
3. When doing `quit` with active breakpoints, ask if user wants to remove them, otherwise the emulator will get stuck when those dangling breakpoints are hit.
4. Breakpoint freeing.

### DIFF
--- a/src/ticks/debug.c
+++ b/src/ticks/debug.c
@@ -1057,6 +1057,43 @@ static uint16_t get_current_framepointer(struct debugger_regs_t *regs, size_t* i
     return wrap_reg(regs->xh, regs->xl);
 }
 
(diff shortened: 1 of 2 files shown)


desertkun:
I've added a commit to help recover backtrace when the symbol is known, but function behind it isn't. Sometimes compiler would produce essentially garbage symbols like `ASMDISP_RECV_CALLEE  = $0007` so `$0038` would match to it, but there is no function behind it. This change recovers from hat as well.

suborb:
That sounds like the Spectranet library is using an old convention for implementing callee functions.

Regardless, I can see this condition occurring a lot when down in low memory.

PR #1954, opened 2022-01-25
classic - differentiate signed vs unsigned multiply

### LINKED ISSUE #1894: sccz80 - unsigned multiply
Unlike divide and modulus, I [don't think that sccz80](https://github.com/z88dk/z88dk/blob/master/lib/z80_crt0.hdr#L99) issues an unsigned multiply for `uint16_t` or `uint32_t`.  Anyway, if it did then the support files for differentiating between signed and unsigned multiplies are now available.

Currently in both sccz80 and sdcc `uint32_t` is being handled as a `long` signed number.

Ideally, the code paths for these two types (`unsigned int` and `unsigned long`) could be separated from the signed version, removing the sign analysis code for the unsigned code path.
 - `l_mult_u`
 - `l_long_mult_u`

feilipu:
When the result of a multiply is the same size as the multiplicands there's no reason to differentiate on signed or unsigned, except where the algorithm is trying to optimise on zeros, or to avoid sign handling overhead.

The newlib integer and long fast math libraries do this for z80, by [making the multiplicands positive](https://github.com/z88dk/z88dk/blob/master/libsrc/_DEVELOPMENT/math/integer/fast/l_fast_muls_16_16x16.asm#L47) before multiplication, so it makes sense to allow sccz80 to use the correct unsigned multiply where appropriate (eg. for pointers that could look negative although they're just large) to avoid the overhead of double `l_neg` handling.

Aside from z80 case, the sccz80 signed `l_mult` and unsigned `l_mult_u` (etc) implementations are identical.
Or at least they are today.

### DIFF
--- a/lib/z80_crt0.hdr
+++ b/lib/z80_crt0.hdr
@@ -97,6 +97,7 @@
 	EXTERN    l_bcneg
 	EXTERN    l_case	;Integer case
 	EXTERN    l_mult	;Integer hl = hl * de
+	EXTERN    l_mult_u	;Integer unsigned *
 	EXTERN    l_div		;Integer signed / hl=de/hl, de=de%hl
 	EXTERN    l_div_u	;Integer unsigned /
 	EXTERN    l_mod		;Integer % (for gbz80) hl=de%hl
@@ -148,6 +149,7 @@
 	EXTERN	l_long_mod
 	EXTERN	l_long_mod_u
 	EXTERN	l_long_mult
+	EXTERN	l_long_mult_u
 	EXTERN	l_long_asr_u	;Long unsigned >>
 	EXTERN	l_long_asr_uo	;Long unsigned >> shift in c
 	EXTERN	l_long_case
@@ -307,6 +309,7 @@
 	EXTERN	l_i64_add
 	EXTERN	l_i64_sub
 	EXTERN	l_i64_mult
+	EXTERN	l_i64_mult_u
 	EXTERN	l_i64_mod
 	EXTERN	l_i64_mod_u
 	EXTERN	l_i64_div
--- a/libsrc/_DEVELOPMENT/l/sccz80/4-r2k/l_mult.asm
+++ b/libsrc/_DEVELOPMENT/l/sccz80/4-r2k/l_mult.asm
@@ -3,6 +3,7 @@ SECTION code_clib
 SECTION code_l_sccz80
 
 PUBLIC  l_mult
+PUBLIC  l_mult_u
 PUBLIC  l_mult_0
 
 ; Entry: hl = value1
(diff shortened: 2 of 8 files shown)


ped7g:
> by making the multiplicands positive before multiplication
you don't need to do even that, unless the test+negation makes it faster than doing 16x16 of original values. Ie. 16=16x16 case is sign type "indifferent", always producing the correct bit pattern.
I tried to quickly check the classic Z80 code what it does, but there's way too many indirections, etc.. not very suitable to follow in web. So I'm not sure if those two negations are faster or slower, but they should produce the same bit pattern in the truncated result.

But just for inspiration, for the Z80N with HW 8x8 MUL I'm slowly building collection of supposedly most optimal mul routines, but it's still way incomplete.

But there's also other type of signed mul optimisation, avoiding the input negation and post-result negation for signed muls by using unsigned mul always, and patching result by subtracting the extra value added by unsigned mul, I tried to explain how it works in the comment (with general rules, how to deal with signed muls):
https://github.com/ped7g/ZXSpectrumNextMisc/blob/master/snippets/mul.i.asm#L221

feilipu:
>> by making the multiplicands positive before multiplication

> you don't need to do even that, unless the test+negation makes it faster than doing 16x16 of original values.

The newlib fast multiplication libraries do a lot of demotion to smaller (and smaller) multiplies where the leading bytes are zero, so making numbers positive, and double negation (and hence leading bytes zero) makes a lot of sense for signed values. The small multiplication libraries just hammer through so there's no differentiation needed.

I've been thinking about using the `inc d` `dec c` instruction on leading zero detection to trigger the negation after the first `inc d` instruction. But not fully formed ideas yet, hence the cryptic comment on the PR.

ped7g:
but you will get zero in whole upper byte only for values -1 to -255 (0xFFFF .. 0xFF01), so then for me the more meaningful test than `bit 7,high_byte` would be `inc high_byte : call z,special_partial_negate : dec high_byte` .. with the special partial negate knowing there will be extra dec high_byte following, so doing something like:
```
  ; ... pre-mul test
  inc d  ; check sign of DE
  call z,special_partial_negate_de
  dec d
  ;...

;...
special_partial_negate_de:
  ld a,e
  cpl
  ld e,a
  inc de
  inc d   ; make the following `dec d` reach 0 in D
  ret
```

feilipu:
> with the special partial negate knowing there will be extra dec high_byte following, so doing something like...

Yes, exactly. But, I'm a bit slower thinking this through than you, clearly... :smile:

feilipu:
Have added demotion where either parameter MSW is zero. Easy add.

Have tested checking sign first,before multiplying absolute values with MSW check, and then setting sign of result, but the resulting code is slower than just testing for zero MSW (and ignoring sign).

PR #1961, opened 2022-01-31
Sega Master System: Fixed issues related to interrupts and accessing VDP I/O ports

iratahack:
The following are addressed in this pull request:

1) The *timer* variable was being incorrectly updated in the *int_RASTER* routine. This variable is not accessible from C and not accessed by any other asm code. In theory it could be removed altogether, but I left it where it is, corrected the code and added _ (underscore) to it's name so it can be accessible from C.

2) Sega Master System programs compiled with z88dk do not run correctly on the RetroCopy_Win_0.960 emulator. Specifically the screen is messed up. I have tracked this down to bit 1 in VDP register 0 needing to be set. I have confirmed on 3 other emulators that setting this bit has no adverse effects. Sonic The Hedgehog sets it, so it must be good ;)

3) The routine asm_z80_pop_ei used to use the zero flag to determine if interrupts should be enabled or not at the end of the routine. I believe this should be the carry flag based on the companion code in asm_z80_push_di.

4) Routines which access VDP registers are now protected from interrupts by the following routines:
* l_tms9918_disable_interrupts
* l_tms9918_enable_interrupts


iratahack:
I'm starting to rethink the current approach of accessing the VDP hardware.

Currently, interrupts are disabled for the minimum amount of time, typically around the 2 writes to the VDP command port.

I'm thinking it would be better if these routines didn't touch the interrupts at all. This would allow these functions to be called by an ISR, which is very handy. If the coder wants to call them from outside an ISR they need to disable interrupts around the call to protect them from the ISR.

Alternatively, additional wrapper functions could be created which disable and enable interrupts around each function. For example:

_set_sprite_isr:
    ; Do something with VDP regs
    ret

_set_sprite:
    di
    call _set_sprite_isr
    ei
    ret

Any thoughts?
IH.

suborb:
Thanks for spotting and updating these - I keep forgetting about the extra SMS bits of the library.

Regarding the interrupts, the tms9918 use l_push_di/l_pop_ei to keep track of the interrupt state correctly. For consistency it makes sense to use those.

iratahack:
Thanks, @suborb . Let me take a look at those routines and see how to fit them in.

iratahack:
@suborb all changes are in and tested. Also touched-up a couple of the sample apps.
IH.

libsrc/target/sms/set_vdp_reg.asm
```diff
@@ -1,23 +0,0 @@
-	SECTION	code_clib
```
iratahack, line 1: This routine is not used. Instead the one from .../video/tms9918/gen_set_vdp_reg.asm is used.

libsrc/_DEVELOPMENT/z80/z80/asm_z80_pop_ei.asm
```diff
@@ -72,7 +72,7 @@ ELSE
 
    IF __Z80 & __Z80_NMOS
    
-      jr NZ, di_state
```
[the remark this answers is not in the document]
feilipu: I broke it inadvertently.
Thanks for fixing.

lib/target/sms/classic/sms_crt0.asm
```diff
@@ -94,14 +94,9 @@ int_RASTER:
 
 ;int_VBL: 
     push    hl 
-    ld      hl, timer
```
iratahack, line 97: I can see what this code is trying to do bit incrementing 'A' doesn't set the carry when it wraps to 0 and 'adc 1' will add 1 or 2 if carry is set. So, keeping it simple, just load a 16-bit value, increment it and save it.

lib/target/sms/classic/sms_crt0.asm
```diff
@@ -220,13 +216,14 @@ gg_palette:
...
     defw 0x0fff             ;ff ff ff
+ENDIF
 
     DEFC SpriteSet          = 0       ; 0 for sprites to use tiles 0-255, 1 for 256+
     DEFC NameTableAddress   = $3800   ; must be a multiple of $800; usually $3800; fills $700 bytes (unstretched)
     DEFC SpriteTableAddress = $3f00   ; must be a multiple of $100; usually $3f00; fills $100 bytes
 
 _Data: 
-    defb @00000100,$80 
+    defb @00000110,$80
```
iratahack, line 226: The RetroCopy_Win_0.960 emulator doesn't display the screen correctly unless bit 1 is set. Seems to have no adverse effects on the 3 other emulators I tried.

libsrc/target/sms/sms.hdr
```diff
@@ -2,3 +2,13 @@
 	DEFC	SpriteSet          = 0       ; 0 for sprites to use tiles 0-255, 1 for 256+
 	DEFC	NameTableAddress   = $3800   ; must be a multiple of $800; usually $3800; fills $700 bytes (unstretched)
 	DEFC	SpriteTableAddress = $3f00   ; must be a multiple of $100; usually $3f00; fills $100 bytes
+
+
+	DEFC	__IO_VDP_DATA      = 0xbe
```
iratahack, line 7: Defines for accessing the VDP registers. Maybe these should be pulled from the video library? They seem to be defined in multiple places and I wasn't sure which ones to use so here they are again.

examples/sms/apktest.c
```diff
@@ -12,7 +12,7 @@ extern char testtxt_apk[];
 void main() {
 	char buf[100];
 
-	aplib_depack(testtxt_apk, buf);
+	aplib_depack(buf, testtxt_apk);
```
iratahack, line 15: The parameters seemed to be switched here.

examples/sms/chicken.c
```diff
@@ -94,12 +94,15 @@ void main() {
 		while (y < 176) {
 			while (get_vcount() < y) {
 			}
+#if 0
+			// load_palette() is too slow for this to work
```
iratahack, line 98: Due to adding calls to disable/save and restore interrupts the timing of load_palette and scroll_bkg has changed. They not take longer. This appears to be too long for this example to work. By removing the palette changes at least the cars move left and right correctly.

examples/sms/Makefile
```diff
@@ -3,10 +3,10 @@ COMPILER ?= sccz80
 CFLAGS = -vn
 
 ifeq ($(COMPILER),sdcc)
-   CFLAGS += -compiler=sdcc  -SO3 --max-allocs-per-node200000 --fsigned-char 
-all: test.sms libctest.sms  apktest.sms testsnd.sms
+   CFLAGS += -compiler=sdcc -SO3 --max-allocs-per-node200000 --fsigned-char
+all: test.sms libctest.sms pongmstr.sms pngmstr2.sms chicken.sms testsnd.sms libpsgtest.sms
```
iratahack, line 7: Included more working examples for SDCC. apktest does not work with SDCC, it fails to link because the apk routine is missing.

examples/sms/Makefile
```diff
@@ -3,10 +3,10 @@ COMPILER ?= sccz80
...
 else
-all: test.sms libctest.sms pongmstr.sms pngmstr2.sms apktest.sms chicken.sms testsnd.sms
+all: test.sms libctest.sms pongmstr.sms pngmstr2.sms apktest.sms chicken.sms testsnd.sms libpsgtest.sms
```
iratahack, line 9: Added a new example to showcase PSGlib.

examples/sms/chicken.c
```diff
@@ -34,17 +36,16 @@ int rand_speed() {
 }
 
 void main() { 
-    int x = 0; 
-    int y = 0; 
-    int  i, j; 
+    unsigned char x = 0;
```
iratahack, line 39: Optimizations so that the chicken example app runs after adding the interrupt protection to the routines which access the VDP hardware.

examples/sms/libpsgtest.c
```diff
@@ -0,0 +1,66 @@
+#include <sms.h>
```
iratahack, line 1: New example to play a psg file and scroll flashing text across the screen.

libsrc/_DEVELOPMENT/l/sccz80/9-common/l_pop_ei.asm
```diff
@@ -2,7 +2,10 @@ SECTION code_clib
 SECTION code_l_sccz80
 
 PUBLIC l_pop_ei
+PUBLIC l_pop_ei_jp
```
iratahack, line 5: Export this entry point so routines can jump to it and exit through it rather than calling it and then returning.

src/z80asm/asmstyle.pl
```diff
@@ -80,6 +80,15 @@ sub parse_line {
                 else { die; }
 	    }
         }
+        # #define
```
iratahack, line 83: Added support to format #defines's in asm code.

examples/sms/Makefile
suborb: I've just exported the app symbols for classic/sdcc so I'll add apktest once I've merged this pr.

libsrc/target/sms/sms.hdr
suborb: ...so one of my plans for a reorganisation is to put all the defines needed for a target library into a single place in `libsrc/target/XXX/?????.inc` - for example the ports for the PSG, 1 bit sound, graphics resolution/flags etc

This would remove the duplicated definitions. I did start moving stuff around in preparation but as usual with these things I ended up with a somewhat larger snowball than I intended to.

suborb:
Thank you for this cleanup, I checked 3dcity as well and it's still working!

PR #1962, opened 2022-01-31
Ticks breaking

desertkun:
I think I have figured out a way to do real hardware debugging, e.g. a gdbserver run on spectrum itself (via spectranet).

That would require slight spectranet change, though, while I am going to try and make the change, I need these.

1. Remote is "NonBreakable", means you can not break execution by pressing CTRL+C, you'd have to press the NMI button on the hardware.
2. Memory is limited and thus packet size is reduced, we have to respect that.
3. That debugger is going to support only sp, pc, hl, de, bc and af for now, and thus ix is optional for debugging (since we have __debug_framepointer anyway)
4. Restore did not respect sp (whoops)

### DIFF
--- a/src/ticks/backend.h
+++ b/src/ticks/backend.h
@@ -42,6 +42,7 @@ typedef struct {
     debugger_write_memory_cb debugger_write_memory;
     debugger_read_memory_cb debugger_read_memory;
     void_cb invalidate;
+    uint8_t breakable;
     void_cb break_;
     void_cb resume;
     void_cb next;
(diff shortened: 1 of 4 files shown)


desertkun:
@suborb Hope you find the time to get this merged. I am trying to put things together, even made a website with all the tools needed to debug code. https://speccytools.org/

suborb:
Apologies for the delay on getting this merged: not enough hours in the day at the moment.

Looking forward to seeing the stub and seeing if it can be generalised.

desertkun:
Thanks. That depends on https://github.com/spectrumero/spectranet/pull/16 getting merged, hopefully it does anytime soon

PR #1979, opened 2022-02-20
fix: add .gitignore entries for zsdcc_*_src.tar.gz and include/alloc/malloc.h

jorgegv:
fix: add .gitignore entries for zsdcc_*_src.tar.gz and include/alloc/malloc.h


suborb:
Thanks - I've stopped malloc.h from being copied and added zsdcc to the .gitignore now.

PR #1983, opened 2022-02-21
feat: add common scancodes to input.h

jorgegv:
Add common scancodes to input.h so that key translation table is not brought in for static key mappings known at compile time


suborb:
Thanks again, I reworked this to put the scan codes in arch/zx/input.h since they're not valid across all targets.

PR #1986, opened 2022-02-26
sp1 sdcc + classic fixes

jorgegv:
Comiple and link works for this test program:

```c
// sdcc   + classic -> OK	zcc +zx -vn -compiler=sdcc -lsp1-zx sp1.c -o sp1.bin
// sccz80 + classic -> OK	zcc +zx -vn -lsp1 sp1.c -o sp1.bin
// sdcc   + newlib  -> OK	zcc +zx -vn -clib=sdcc_iy sp1.c -o sp1.bin
// sccz80 + newlib  -> OK	zcc +zx -vn -clib=new sp1.c -o sp1.bin

// for classic, use this include
#include <arch/zx/sprites/sp1.h>

// for newlib, use this include
//#include <games/sp1.h>

void main(void) {
    sp1_UpdateNow();
}
```


libsrc/_DEVELOPMENT/temp/sp1/zx/sprites/asm_sp1_AddColSpr.asm
```diff
@@ -33,9 +33,19 @@ csalloc:
    push bc
    ld hl,24                   ; sizeof(struct sp1_cs)
    push hl
+
+; classic malloc exits with carry set if success, reset on failure
+; newlib asm_malloc exits with carry reset if success, set on failure
```
suborb, line 38: I don't think this is correct: u_malloc() is a C function so "can't" return with a carry flag condition.

jorgegv:
memory allocation in `asm_sp1_AddColSpr.asm`, `asm_sp1_CreateSpr.asm` and `asm_sp1_DeleteSpr.asm` is done by directly calling `asm_malloc` and `asm_free`. They need to be changed to calls to `_u_malloc` and `_u_free` in Classic.

Fixed in 5d890083688cee73e1673f2e50f42e0c93cf1a2d

libsrc/_DEVELOPMENT/temp/sp1/zx/sprites/asm_sp1_AddColSpr.asm
jorgegv: I understand that, but indeed the asm code in the SP1 functions check for this C or NC flag, and they are inverted in the asm heap alloc routines... The asm code is _using_ the Carry flag to check for errors in both places...

Anyway, I just put inside the IF-ENDIF the code that was different between the Classic and the Newlib code, so I guess this is harmless re: the PR content.

I think I have an explanation: maybe this is a bug on the original Classic-SP1 sources: supposedly you could use your own memory allocation routines by setting `u_malloc` and `u_free` pointers, but indeed the SP1 code was using the Carry check for errors, and not the return value. And because everyone just set `u_malloc = malloc` and `u_free = free`, and everything worked, nobody noticed.

So you can set your own routines...as long as you end up quickly calling the internal `malloc`, `HeapAlloc`, etc. ones...which set/reset the Carry flag for error reporting (which is what SP1 expects).

suborb: Yes, that would be that it worked through pure luck - the examples implement u_malloc which just calls malloc directly so carry would be preserved. 

The luck seems to be holding....

jorgegv: ...and may be this is the reason that Newlib-SP1 sources just call `asm_malloc` instead of using the alternate pointers.

jorgegv:
A new important change: added new SP1 header from Newlib into temporary location `arch/zx/sprite-new/sp1.h` copied from Newlib SP1 include file for SCCZ80 (matching the ASM SP1 files adapted from Newlib SCCZ80). In that file, fixed function declarations with more than 1 parameter to have `__smallc` calling convention and match ASM code.

jorgegv:
Updated all SP1 examples to work. Indeed making all examples working helped to discover a couple of obscure fixes needed.

All examples are working except `ex5e.c` which has some strange glitches apparently not related to SP1 (under investigation)

Meanwhile, @suborb , you can check the changes made to the glue functions. I have applied the fixes unconditionally, but maybe they should be conditional depending on compiler/lib.

jorgegv:
Don't waste time trying to Fix the conflict, I'll do It on a new fixed branch and a new PR

PR #1988, opened 2022-03-06
(intel/gb) memswap optimisation for size >= 2

ped7g:
for size == 1 it is +4T (to produce bc=0 upon exit)

Question: the bc=0 on exit (`asm_memswap`) seems to be unused by the library code, are those ABI described in comments mandatory and locked, or may it be worth to break it and say BC is just modified? If it would be possible to drop bc=0 on exit, then the `ld b,a` before `ret` can be removed too and the routine will have same performance for size==1 as current one.

I'm also planning to add few more changes optimising some GB paths, so draft PR at this moment, but I did want to ask about that bc=0, as it's a bit annoying there's +4T for size==1.

### DIFF
--- a/libsrc/_DEVELOPMENT/string/z80/asm_memswap.asm
+++ b/libsrc/_DEVELOPMENT/string/z80/asm_memswap.asm
@@ -41,23 +41,26 @@ IF __CPU_INTEL__ || __CPU_GBZ80__
    inc b
    inc c
 
+   ld a,b
+loop_a:
+   push af
+
 loop:
-   push bc
-
    ld a,(de)
-   ld c,a
+   ld b,(hl)
 
-   ld a,(hl)
+   ld (hl+),a
+   ld a,b
    ld (de+),a
 
-   ld a,c
-   ld (hl+),a
-
-   pop bc
    dec c
    jr NZ,loop
-   dec b
-   jr NZ,loop
+
+   pop af
+   dec a
+   jr NZ,loop_a
+
+   ld b,a
 
 ELSE
 


feilipu:
> Question: the bc=0 on exit (`asm_memswap`) seems to be unused by the library code, are those ABI described in comments mandatory and locked, or may it be worth to break it and say BC is just modified?

Can't comment on the validity of the ABI itself but, with respect to creating "unnecessary surprise" for the user, I feel it would be unwise for INTEL / GBZ80 ABI to differ from the Z80 implementation, whatever that may be.

ped7g:
Question2: the `ld a,(hl+)` on `__CPU_INTEL__` becomes `ld a,(hl) : inc hl`? Because there are several argument-loading prologues which can be faster on GB by using the dedicated `ld a,(hl+)` but would become slower on 8080 if it is macro-expanded like I guess it is. Is it worth to write two different code paths for INTEL vs GB then (I mean if I do the change, are you willing to adopt it and maintain :) ... as I'm not regular contributor to z88dk (or user of it), just cherry picking pieces of asm to optimise for the sake of having fun with Z80-like CPUs)

feilipu (approved):
~~Could save 4 ticks by using `loop_b:` `push bc`and `loop_c:` rather than moving big loop counter into `a` register.

Avoids having to adjust `bc` to zero too.~~

And for interest I wrote same thing elsewhere recently using `ex de,hl` rather than moving the byte across registers. It is no faster, but imho code looks nicer to swap the pointers (knowing it's done with just a single flip flop internally).

~~The same big loop, small loop trick could be done with the `asm_strrev` function too.~~

ped7g:
@feilipu 
> Could save 4 ticks by using loop_b: push bcand loop_c: rather than moving big loop counter into a register.

not so easy, because it needs two temporary registers for "swap", and you can't do `push/pop bc` in outer loop, because that would preserve original C, while the loop mechanics need to do original C + C=0 aferwards, so using the A for outer counter is a trick how to decouple B from C and use `push/pop af` to preserve the outer counter.

`ex de,hl` is not available on GameBoy, but is available on 8080 and 8085, interesting, haven't been aware of that. So further branching may be helpful in some cases, hm.

feilipu:
Hmm. Better I don't try to code on my 📱.

ped7g:
hm... I tried a bit more to figure out what exactly happens in code like https://github.com/z88dk/z88dk/blob/master/libsrc/_DEVELOPMENT/string/c/sccz80/_memstrcpy_.asm#L12

but I can't find any docs about z80asm fake instructions and how it expands stuff like `ld  hl,sp+2` on 8080/8085.. I can guess and assume, but that's error prone.

and I'm puzzled why the 8080/8085 code branch doesn't do pop/push like regular z80, as from the instruction set tables I have found for 8080 it seems the `pop+push` is 21T while 2x `ld r8,(hl) \ inc hl` is 24T, etc.. I'm very familiar with ZX and Z80 and I think I have decent grasp about GB, but I may be missing some important details about Intel stuff.

So I think this one change in GB stuff will be enough for now, back to tampering with some code in quicksort (where this whole activity originally started, this GB optimisation was just distraction :) ). Turning this into non-draft.

suborb:
The 808x and gbz80 don’t preserve all bits in f so pop/push can’t be used.

PR #1990, opened 2022-03-11
tests - stdlib - adding qsort tests for classic and new lib

ped7g:
Adding some tests before tampering with newlib sort implementation a bit more.

The modification of labels in shellsort asm is because they were clashing with local labels emitted in the test C file itself, which is including the asm files (technique copied from test/suites/zx, otherwise I have no idea how to better test newlib stuff, as it has no +test target, plus the config ahead of building library is getting into way of testing various configurations, so including the sources directly seems to be pretty effective, until they contain generic `i_#` labels which are used also by the compiler itself...).

### DIFF
--- a/libsrc/_DEVELOPMENT/EXAMPLES/benchmarks/sorting/z88dk-classic/Makefile
+++ b/libsrc/_DEVELOPMENT/EXAMPLES/benchmarks/sorting/z88dk-classic/Makefile
@@ -1,0 +1,56 @@
+
+.PHONY: all clean
+
+SOURCES += $(wildcard *.c)
+
+all:	benchmark.txt
+
+STYLES_VAL=0 1 2 3
+STYLES_NAME=ran ord rev equ
+
+CFLAGS += -D__Z88DK
+
+# EXESUFFIX is passed when cross-compiling Win32 on Linux
+ifeq ($(OS),Windows_NT)
+  EXESUFFIX 		:= .exe
+else
+  EXESUFFIX 		?=
+endif
+
+MACHINE = z88dk-ticks$(EXESUFFIX) -counter 999999999999
+
+define compile_classic_sccz80	# arg1 = binary name, arg2 = extra defines, arg3 = extra linking options
+	zcc +test -vn $(2) $(CFLAGS) -O2 $< -o $(1) -lndos $(3) -m
+endef
+
+define compile_classic_zsdcc	# arg1 = binary name, arg2 = extra defines, arg3 = extra linking options
+	zcc +test -vn $(2) $(CFLAGS) -compiler=sdcc -SO3 --max-allocs-per-node200000 $< -o $(1) -lndos $(3) -m
+endef
+
+define benchmark_case	# arg1 = compiler, arg2 = binary name, arg3 = extra defines, arg4 = extra linking options
+	$(call compile_classic_$(1),$(2),-DTIMER $(3),$(4))
+	@echo -n "$(2:%.bin=%) " >> $@
+	@# run sort + verification + no timing for exit status (FAIL), if OK, run again for timing
+	$(MACHINE) $(2) > /dev/null \
+	&& $(MACHINE) -x $(2:%.bin=%.map) -start TIMER_START -end TIMER_STOP $(2) >> $@ \
+	|| echo " FAIL" >> $@
+
+endef
+
+define benchmark_styles # arg1 = compiler, arg2 = numbers-count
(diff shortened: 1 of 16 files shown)


feilipu:
Just some quick notes.
- You've checked in an empty `config_private.inc` file, that should not be added.
- Perhaps use `__INTEL__` rather than __8080__` as the check as there's some 8085 machines that might sneak through the cracks otherwise.

ped7g:
> You've checked in an empty config_private.inc file, that should not be added.
but how should I do it then? the compilation fails without it:
```
../../../libsrc/_DEVELOPMENT/stdlib/z80/sort/asm_quicksort.asm:50: error: file open: config_private.inc
  ^---- INCLUDE "config_private.inc"
make: *** [Makefile:16: test.bin] Error 1
```
adding empty file is another practice copied from test/suites/zx test.

will check the intel thing, whether it should apply only to added sorts, or also to the stuff excluded previously by 8080

feilipu:
Hmm. It is odd an empty file is needed to make it work. I don't know why. 🤷‍♂️

ped7g:
because the asm_quicksort newlib implementation is including it.

when newlib is being built as library, there is global config_private.inc, but as I'm hijacking the source into classic-lib target, there's no file. Providing the empty one in project folder looks as sensible kludge to me. (also the real configuration is in the test's C source right ahead of including the newlib asm, so it's all more self-contained and easier to dig into later)

I'm too fresh to z88dk project to have strong opinions about how the tests and build infrastructure should be changed, so I'm currently trying just to mimick other similar cases I managed to find, not planning to become core contributor, just looking into specific asm fragments to improve their performance and try out some extra ideas, but leaving the core maintenance on you guys, sorry :) - although obviously I don't want to intentionally increase the burden, but I may accidentally).

ped7g:
@feilipu also just now I realised.. there are two sort implementations in classic-lib:
`qsort_sccz80_callee` vs `qsort_sdcc_callee`
it looks like both asm files contain identical implementation with different header dealing with different calling convention.

which means the test I did does exercise only one of the two? I don't even know if it builds sccz80 or sdcc variant. On my own projects I would resolve this by running test suites for both compilers, but I'm not sure if here the omission is intentional to keep CI build times lower, or just nobody did add the rules to run both variants of tests.

ped7g:
> * Perhaps use `__INTEL__` rather than `__8080__` as the check as there's some 8085 machines that might sneak through the cracks otherwise.

tried that, but then the `$(compile_8080)` fails, as `__8080__` is defined, but `__INTEL__` is not. I did try to search for `__INTEL__` through whole git repo and haven't find any mention of it.

So I don't know how to apply this suggestion correctly.

feilipu:
Sorry half asleep... `__CPU_INTEL__` is what I mean.

ped7g:
> Sorry half asleep... `__CPU_INTEL__` is what I mean.

Doesn't compile either, I can see it being defined in z80asm (`options.c`), so I guess in asm blocks it would work with z80asm `IF`, but I can't find where the define would be done for .c files with `#ifdef/#ifndef`.

feilipu:
Yes, sorry. Looking into it properly now.
For c the compiler needs either or both of `-D__8080__` and `-D__8085__`.
So, you're probably OK just using `__8080__` as you've done.
I'll be quiet now.

ped7g:
so now sorts are test-covered for default configs and have full bechmarks.

I'm now planning to modify the newlib asm implementation a bit to make it a bit faster, maybe including my hand-written qsort example in the benchmark for direct comparison too.

ped7g:
And I think I'm done with this.

My own curiosity was satisfied (comparing z88dk sort vs my hand written asm), the newlib quick sort should be slightly faster than it was (testing my idea how to get rid of the `divu` in middle-pivot case).

To the point when IMHO now the random-pivot case probably can't compete and could be just removed to simplify the source and newlib build config (and I don't have any idea how to avoid the two `divu` in random pivot case).

And there's Makefile to automate testing all variants, so it should be easier to replicate the results in future for anyone else.

feilipu:
Looks like worthwhile speed up across all sort cases and some cases especially improved.

ped7g:
ok, so I did take a quick look also at newlib shellsort and insertion sort, just minor optimisations on the machine code level (not touching algorithms or general flow too much).

And now I'm definitely done with this PR. Sorry for bumping this so many times... :)

suborb:
Great stuff, thank you!

PR #1996, opened 2022-03-31
Fixed bugs in section name handling

iratahack:
Two fixes:

1) Ignore case when searching for section names
2) When adding a new section, subtract 2 from the name length to account for removing the first 2 characters of the string

### DIFF
--- a/src/ticks/syms.c
+++ b/src/ticks/syms.c
@@ -5,12 +5,6 @@
 #include "ticks.h"
 #include "debug.h"
 
-#if defined(_WIN32) || defined(WIN32)
-#ifndef strncasecmp
-#define strncasecmp(a,b,c) strnicmp(a,b,c)
-#endif
-#endif
-
 static symbol  *symbols[65536] = {0};
 static symbol  *symbols_byname = NULL;
 
--- a/src/ticks/syms.c
+++ b/src/ticks/syms.c
@@ -5,6 +5,12 @@
 #include "ticks.h"
 #include "debug.h"
 
+#if defined(_WIN32) || defined(WIN32)
+#ifndef strncasecmp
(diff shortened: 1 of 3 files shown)


src/ticks/syms.c
```diff
@@ -352,7 +352,7 @@ int address_is_code(int addr)
 
     while ( sect != NULL ) {
         if ( addr >= sect->start && addr < sect->end) {
-            return (strncmp(sect->name, "code_",5) == 0 || strncmp(sect->name, "smc_",4) == 0 );
+            return (strncasecmp(sect->name, "code_",5) == 0 || strncasecmp(sect->name, "smc_",4) == 0 );
```
iratahack, line 355: Ignore case when searching for section names.

suborb: Afk. But I think there’s a #define that deals with windows portability.

iratahack: I found the code below in ticks.c
```
#if defined(_WIN32) || defined(WIN32)
#ifndef strcasecmp
#define strcasecmp(a,b) stricmp(a,b)
#endif
#endif
```
I can add something similar to syms.c to handle `strncasecmp()`

suborb: Please bump it up to ticks.h which is included by both I believe.

suborb:
Thanks for this - just out of interest what's the need fo case insensitivity?

iratahack:
@suborb , I noticed my code wasn't being disassembled. I use capital CODE, BSS, DATA and RODATA when naming my sections. I noticed the code in syms.c which creates a table of these standard sections (read_symbol_file) is case insensitive but the code which does the search (address_is_code) to see if an address is in the code section is not. This caused any standard section name not all lower case to be treated as data.

PR #1997, opened 2022-04-04
Allow banked code with symbol addresses >= 0x10000 to be disassembled.

iratahack:


src/ticks/ticks.h
```diff
@@ -13,6 +13,8 @@
 #include "uthash.h"
 #include "utlist.h"
 
+#define MAX_ADDRESS 0x100000
```
iratahack, line 16: Up to 1MB of banked address space.

src/ticks/disassembler_alg.c
```diff
@@ -687,7 +687,7 @@ int disassemble2(int pc, char *bufstart, size_t buflen, int compact)
         buf[offs++] = ' ';
         buf[offs] = 0;
     }
-    offs += snprintf(buf + offs, buflen - offs, ";[%04x] ", start_pc);
+    offs += snprintf(buf + offs, buflen - offs, ";[%04x] ", start_pc & 0xffff);
```
iratahack, line 690: Only display bits 15:0 of the address.

src/ticks/disassembler_alg.c
```diff
@@ -27,7 +27,7 @@ typedef struct {
 
 
 #define READ_BYTE(state,val) do { \
-    val = bk.get_memory(state->pc++); \
+    val = bk.get_memory((state->pc++ & 0xffff)); \
```
iratahack, line 30: Limit memory accesses to the 64KB buffer size.

PR #1998, opened 2022-04-05
z88dk-dis: Handle banked code with symbol addresses > 65535

iratahack:


src/ticks/disassembler_main.c
```diff
@@ -49,13 +49,13 @@ int main(int argc, char **argv)
     char  *program = argv[0];
     char  *filename;
     char  *endp;
-    uint16_t    org = 0;
-    int         start = -1;
-    uint16_t    end = 65535;
+    unsigned int    org = 0;
+    int    start = -1;
+    unsigned int    end = -1;   // Max value (all f's)
```
suborb, line 54: This seems a bit fishy and playing around with limits and type conversion.

If it were INT_MAX then would we just disassemble the memory block 2^16 times?

src/ticks/ticks.h
```diff
@@ -13,6 +13,9 @@
 #include "uthash.h"
 #include "utlist.h"
 
+#define SYM_TAB_SIZE    0x10000
+#define BUFF_SIZE       0x10000
```
suborb, line 17: BUFF_SIZE might as well live in disassembler_alg.

src/ticks/disassembler_main.c
iratahack: Probably UINT_MAX since *end* is unsigned?

src/ticks/ticks.h
iratahack: Although it is also used in disassembler_main.c to allocate the buffer.

src/ticks/disassembler_main.c
suborb: The function that does the disassembly takes a signed int I think?

iratahack: It does. I was following the original code which defined *end* as unsigned. I'll change it to signed and test it.

iratahack: Updated and tested.

suborb:
Apologies, I'm being really slow with PRs at the moment and something was bugging me about this one that I couldn't quite describe.

I think I know what you're trying to do - get labels printed when you disassemble a binary corresponding to a bank - but I'm worried that just adding labels regardless of banks might cause the disassembly of the "main binary" to be printed with erroneous labels - I can't remember which way the linked list goes.

If I'm correct, I'm wondering if we need to preserve the bank information in syms.c and pass in a bank to match when mapping an address to a label. For the disassembly the bank index could be a command line argument.

iratahack:
@suborb , I reviewed the code, when the symbol table is created, the symbol address (which can be > 65535) is saved in the symbol table. So we already have the bank number in the symbol table.

However, I did find a problem in `find_symbol()` where it returns the first entry in the linked list at that address mod 0x10000. This can cause an incorrect symbol to be displayed depending on the order of the linked list. To address this, when searching for a symbol, I suggest adding a check for the address passed into find_symbol() with the address in the symbol table. Like below.

This means only the symbol which matches the address, including the bank number, will be displayed. 

```
    while ( sym != NULL ) {
        if (sym->address == addr)
        {
            if (preferred_type == SYM_ANY)
            {
                return sym->name;
            }
            if (preferred_type == sym->symtype)
            {
                return sym->name;
            }
        }
        sym = sym->next;
    }
```
Let me know what you think.
IH.

suborb:
Perfect - a neat way of doing it.

PR #2002, opened 2022-04-10
TVC time() fixed for ticks of system (instead of ticks * 65536)

dikdom:
fix for TVC time()


feilipu:
Doesn't `ld de,0` work?

dikdom:
> Doesn't `ld de,0` work?

Bad habbit. I'll fix it soon. 
(ld de,0 is 10T, while ld d,0 ; ld e,d is 11T altogether. For some reason I thought this is faster...)

feilipu:
Done? Or more to add?

dikdom:
> Done? Or more to add?

Done.

PR #2003, opened 2022-04-15
TVC graphic routines 

dikdom:
The graphic routines are backed up by the ROM routines. The TVC ROM coordinate system is (0,0) - (1023,959) in all the three TVC resolutions - the OS will do the transformation to the real pixels. 
(0,0) is the lower left corner, so some transformation needs to be done in the z88dk lib also if we want to match with the graphics.h philosophy (0.0 should be top left).
Remaining methods (getx/y, point, multipoint, drawb, fillb, clg, clga, uncircle, xor methods) are still missing, I would like to see if the approach is ok. _callee usage is a bit foggy to me. 
I made this to be a separate gfx library because originally I thought that a fast approach also could be implemented, leaving out the ROM functions. This would be a tremendous work as I see right now. The coordinate transformation could be spared, the circle drawing could be implemented with a much faster, direct pixel access, and the filled box drawing can be somehow sped up.
The tvc_romgfx lib advantage is that it doesn't require anything to be placed in a special bank (outside of the vROM bank), because the vRAM access is anyway in ROM. (hm, point() will be tricky, I'll see).


dikdom:
drawb, circle, fillb, vram mapping implemented.

suborb:
Thanks - now the primitives are working is this good to go in?

dikdom:
> Thanks - now the primitives are working is this good to go in?

Yes, it is. Hope it is not a problem that not _all_ methods are implemented from `graphics.h` ..

suborb:
Not a problem at all - something is better than nothing!

PR #2006, opened 2022-04-25
sp1 sdcc + classic fixes (new)

jorgegv:
This PR has the goal of using same SP1 sources from newlib, for compilation with Classic + SDCC. It fixes the mess from PR https://github.com/z88dk/z88dk/pull/1986, which has been withdrawn.

Changes to be done:

- [x] Create new library targets and Makefile
- [x] Ensure all SP1 C stubs save/restore IX when needed
- [x] Ported all examples as a testbed for migration
- [x] Remove old sources from build
  - [x] Replace sp1.h with newlib one
  - [x] Replace references to `spectrum` sources with `zx` ones, then remove `spectrum` directory (old SP1 classic sources)
    - [x] Migrate `zx81hr` sources
    - [x] Migrate `ts2068hr` sources
- [ ] Debug issue with `ex5e.c`


include/arch/zx/sprites/sp1.h
```diff
@@ -217,6 +217,7 @@ extern void               __LIB__   sp1_MoveSprAbs(struct sp1_ss *s, struct sp1_
 extern void               __LIB__   sp1_MoveSprRel(struct sp1_ss *s, struct sp1_Rect *clip, void *frame, char rel_row, char rel_col, char rel_vrot, char rel_hrot);
 extern void               __LIB__   sp1_MoveSprPix(struct sp1_ss *s, struct sp1_Rect *clip, void *frame, uint16_t x, uint16_t y);
 
+// hook1 and hook2 must have prototype: void userfunc( uint8_t count, struct sp1_cs *c ) __smallc;
```
suborb, line 220: If this the case then prototype properly - if there's parsing issues then a `typedef void (*sp1_hook_func)( uint8_t count, struct sp1_cs *c ) __smallc` should get around them.

suborb:
Thanks for re-doing.

Please feel free to delete the classic version as part of this PR and update the `arch/zx/sprites/sp1.h` header.

include/arch/zx/sprites/sp1.h
jorgegv: The function can have __z88dk_fastcall or not, and the prototype would be different for each. The functions using the hooks can use both types, so in order to be compatible, a `void *` pointer needs to be passed.

jorgegv:
Classic SP1 sources disabled from the build, but the `zx81hr` and `ts2068hr` sources reference some files under `spectrum` dir, so I'm not removing it until I also fix those sources.

suborb:
@jorgegv is there more to come or should I merge this through?

jorgegv:
> @jorgegv is there more to come or should I merge this through?

It depends on what we need/want. I just ported the Spectrum sources, but when I started porting the other ones, it turned out they are a bit different than the Spectrum ones. There is even one struct that has 1 byte less for TS and ZX81 (for the lack of attributes), so they need a good review.

For now, the Spectrum port is working, but the others are still pending, and we can't remove the old sources until TS and ZX81 have been migrated also to the new sources.

Also, there is one issue with example `ex5e.c`. It does not work exactly as when compiled with SCCZ80, and I've still to know why, but I think I have the issue pretty cornered in function `sp1_IterateUpdateRect`, which is called by `swap_dfile`. Still working on it though.

So, your choice. Do you have any deadline in mind?

P.S. I'm updating the task list to reflect this issues.

jorgegv:
Maybe we can just move the TS and ZX81 code to a `deprecated` subdir and keep SP1 building for them with the old code... I think I'll do this.

jorgegv:
> Maybe we can just move the TS and ZX81 code to a `deprecated` subdir and keep SP1 building for them with the old code... I think I'll do this.

@suborb Done, you can merge this PR if you wish.

If you do, I'll open an issue for tracking the problem with `ex5e.c`, and you can assign it to me (I'll assign it myself if I am allowed)

suborb:
This is still in a branch in z88dk and not merged to master yet.

PR #2010, opened 2022-05-07
Ticks misc improvements

desertkun:
Commit messages say it all. The most notable is addition of `stack` command:

```
> stack
Examining stack (8 values) starting from $614a:
Values: 6cd8 0000 61c2 0000 0000 0000 0000 0000
Detail:
  $614a (offset 0): 6cd8
    $6c9d _main (+59)
  $614c (offset 2): 0000
    ?
  $614e (offset 4): 61c2
    $61c2 cleanup (+0)
  $6150 (offset 6): 0000
    ?
  $6152 (offset 8): 0000
    ?
  $6154 (offset 10): 0000
    ?
  $6156 (offset 12): 0000
    ?
  $6158 (offset 14): 0000
    ?
```

Also, fixed symbol parsing, due to an error a lot of const symbols were attributed as an addr symbol and thus show up in stack examinations especially near the $0000 address.

### DIFF
--- a/src/ticks/backend.h
+++ b/src/ticks/backend.h
@@ -16,6 +16,7 @@ typedef void (*debugger_write_memory_cb)(int addr, uint8_t val);
 typedef void (*debugger_read_memory_cb)(int addr);
 typedef void (*get_regs_cb)(struct debugger_regs_t* regs);
 typedef void (*void_cb)();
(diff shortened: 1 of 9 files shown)


suborb:
Great stuff, thank you once more.

PR #2011, opened 2022-05-08
Primo

dikdom:
Primo target reworked. Native console is used for printing characters on screen. .ptp and .pri files can be generated by appmake, these formats are used by the emulators and multicart also.


lib/target/primo/classic/primo_crt0.asm
```diff
@@ -57,15 +57,7 @@ program:
...
-	ld	hl,$6800
-	ld	a,(__sp+1)
-	cp	$80
-	jr	c,set_screen
-	ld	hl,$a800
-	cp	$c0
-	jr	c,set_screen
-	ld	hl,$e800
-set_screen:
+	ld	hl,($0013)
```
suborb, line 60: Haha. That’s the easy way! :)

suborb:
I don’t suppose there’s a good keyboard reading routine in the ROM that can be used as well?

dikdom:
> I don’t suppose there’s a good keyboard reading routine in the ROM that can be used as well?

Yes, there is. 
http://primo.homeserver.hu/doc/konyvek/primofuzetek-szoftver.pdf - Hungarian is not that hard... :)
(at the end there is the - part of - ROM code, with comments in Hungarian...)

I planned to replace that as well and swap out completely the generic console management. But for the latter one I got scared with the ioctl routines of the generic console... :)
BTW, I replaced the fputc with the native routine but I still cannot remove the FONT definition, something still uses that... Strange.

suborb:
> > I don’t suppose there’s a good keyboard reading routine in the ROM that can be used as well?
> 
> Yes, there is. http://primo.homeserver.hu/doc/konyvek/primofuzetek-szoftver.pdf - Hungarian is not that hard... :) (at the end there is the - part of - ROM code, with comments in Hungarian...)

Somehow I missed that...I think I ended up doing the entire port from the hardware definition rather than anything else. At least it's not Finnish!

Generally there are 2 (/3) ways of doing the console:

- generic console: this tends to drive the hardware directly and provides a layer of portability across machines - the intention is that if you use it you can just recompile for a different machine with. There's a capabilities function so everything can be decided at runtime.
- ansi terminal: works on the same principle as the generic console
- native console: this usually just calls the ROM printer and exposes the OS print codes (though we usually implement 12d for clear screen) - this means things are totally non-portable but provides a familiar environment for developers used to that machine and also makes for a minimal binary size.

With that in mind, emulating all the operations of the gencon with the ROM printer may not be a worthwhile thing to do.

suborb:
> BTW, I replaced the fputc with the native routine but I still cannot remove the FONT definition, something still uses that... Strange.

I'll have a look at that this evening - is this showing up with hello world?

dikdom:
> With that in mind, emulating all the operations of the gencon with the ROM printer may not be a worthwhile thing to do.
You're suggesting to go back to the generic console routines? So everytime a printf() is called then a complete font is included with the runtime code?

> I'll have a look at that this evening - is this showing up with hello world?
I don't understand your question. When I tried to compile the primo lib without the FONT variable defined, the library build failed.
On the other with this PR it is possible to build a runtime having a printf() which results calling the ROM print routines. (zcc +primo testprint.c -Cz"-t ptp"  ; and you can import that into Ultimo emulator by ALT-F then Autoload)

suborb:
> > I don’t suppose there’s a good keyboard reading routine in the ROM that can be used as well?
> 
> Check page F-68, KYBHND (jumped from $001D), that is kind of 'getkey', and F-71, KKBHND (jumped from $0025), scans keyboard and waits until a key is pressed.

Thank you!

> You're suggesting to go back to the generic console routines? Or ... ? In the PR the native console is implemented - if my understanding is correct. Coincidentally (?) 0x0c is indeed CLEAR-SCREEN (see page F-59).

Oh no, not at all! Having a choice is good, I was just mentioning that we've tended to keep the code for the native console as small as possible so there's not much normalisation.

> I don't understand your question. When I tried to compile the primo lib without the FONT variable defined, the library build failed. On the other with this PR it is possible to build a runtime having a printf() which results calling the ROM print routines. (zcc +primo testprint.c -Cz"-t ptp" ; and you can import that into Ultimo emulator by ALT-F then Autoload)

The magical incantation before this PR switched the defaults was `-pragma-redirect:fputc_cons=fputc_cons_native`. Totally understand the concern about the font - on several targets we point it to that characters in the ROM since it's in the right format.

lib/config/primo.cfg
```diff
@@ -15,7 +15,7 @@ OPTIONS		 -O2 -SO2 -iquote. -Ca-IXIY -Cl-IXIY -DZ80 -DPRIMO -D__PRIMO__ -M -clib
 CLIB		default -lprimo_clib -lndos -LDESTDIR/lib/clibs/ixiy
 
 SUBTYPE     none 
-SUBTYPE     default -Cz+mamepp
+SUBTYPE     default -Cz+primo
```
suborb, line 18: Can you add back:

`SUBTYPE mamepp -Cz+mamepp` to still allow creation of the snapshot files.

suborb: I can fix it up, but you've created two subtypes named "default"!

suborb:
Great, thank you for the contribution!

PR #2015, opened 2022-05-12
ticks: setting a breakpoint on incomplete filename

desertkun:
Some projects, like mine, generate `.map` with very long filenames. That makes setting a breakpoint on those cumbersome, as currently exact filename is required.

This change allows partial matching, if nothing matches verbatim.

```
b client.c:71
Adding breakpoint at '/Users/desertkun/Documents/Work/zx-sandbox/client/src/client.c' $6a03 (_client_message_sync_obj+268)
i b
1:	PC = $6a03 (_client_message_sync_obj+268) 
```

Also, added support for anonymous types

```
struct map_object_t
{
    uint16_t object_id;
    uint16_t x;
    uint16_t y;
    map_object_type_t type;
    union
    {
        uint8_t payload;
        uint8_t client_id;
    };
};
```
is processed as such
```
$2 = <struct map_object_t> {object_id=3, x=94, y=17, type=1, anon_3={payload=2, client_id=2}}
```

### DIFF
--- a/src/ticks/debug.c
+++ b/src/ticks/debug.c
@@ -748,7 +748,7 @@ int debug_find_source_location(int address, const char **filename, int *lineno)
     return 0;
 }
 
-int debug_resolve_source(char *name)
+int debug_resolve_source(char *name, const char** corrected_name)
 {
     char *ptr;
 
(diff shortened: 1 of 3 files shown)


desertkun:
@suborb mind taking a look into this?

suborb:
@desertkun I keep planning to, but end up running out of time in the evening to do anything fun :(

desertkun:
@suborb C-c-can I join the collaborators so I wont be a burden? I won't touch anything besides ticks :)

desertkun:
Thanks, I'll merge this, then 😀

PR #2019, opened 2022-05-16
Update primo_crt0.asm

EnduroEntrance:
Without this check, the zcc -zorg option gives an error.
With this change, the -zorg option can define the value of CRT_ORG_CODE.
I tested, it works correctly with this modification.

### DIFF
--- a/lib/target/primo/classic/primo_crt0.asm
+++ b/lib/target/primo/classic/primo_crt0.asm
@@ -37,8 +37,9 @@
         defc    CONSOLE_ROWS = 24
         INCLUDE "crt/classic/crt_rules.inc"
 
+IF !DEFINED_CRT_ORG_CODE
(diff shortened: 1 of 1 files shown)


dikdom:
Just out of curiosity: what's the point changing the org value? Do you have an external .ptp/.wav generator? How do you load the code (into emulator or machine)? (using other value than 0x4400 causes that the generated .pri/.ptp file would not work)

EnduroEntrance:
Currently i am writeing a Memory Test program for the Primo A64 computer. In this machine the first memory bank is damadged, but the LOAD command works properly. I load the memory test program from the D000 memory address.
I created the program with z88dk. It generated a .pp file. I converted the .pp file to .ptp, then the .ptp file to .wav with external programs.
Then I load the .wav into the Priom A64. It works. For test i load the .ptp into Ultimo Primo emulator. It is work correctly too.
The original zcc command not enabled the -zorg options. (It gives an error.) If i load my test program from then 4400 address, it does not work.
My patch only add a new feature. If not use -zorg options, the program load from 4400.

suborb:
Nice to know the use case - thanks!

PR #2021, opened 2022-05-17
Shadow registers are used instead of reading port $00 (which gives wrong result)

dikdom:


dikdom:
Oh... 
- this PR contains an other complete pull request's content...
- The first commit was a complete mess, IN was used instead of LD
- The second commit is OK and also contains some source formatting..

PR #2025, opened 2022-06-03
IDE debug (mi/2)

desertkun:
This PR is a home for ongoing development for `mi/2` protocol implementation for z88dk-gdb, aka visual debugging with IDEs

Standard document: https://sourceware.org/gdb/onlinedocs/gdb/GDB_002fMI.html

This is being only tested with CLion, but will eventually test with Visual Studio Code.

For VSCode, have this `launch.json`:
```
{
    "version": "0.2.0",
    "configurations": [
        {
            "type": "gdb",
            "request": "attach",
            "name": "Attach to gdbserver",
            "executable": "${workspaceRoot}/build/client.map", // or appropriate .map of your project (-m -debug needed!)
            "target": "127.0.0.1:1337",
            "remote": true,
            "cwd": "${workspaceRoot}", 
            "gdbpath": "/usr/local/bin/z88dk-gdb",
            "autorun": [
            ]
        }
    ]
}
```

CLion configuration:

- [x] Placing, removing breakpoints both before connecting and under run
- [x] Backtrace
- [x] Local variables
- [x] Step Over, Step Into, Finish
- [x] Disassembly view
- [x] Expression evaluation
- [x] Fix global variables not showing up
- ~~Implement recursive view of a variables (only upmost root members are currently shown)~~ Too lazy to do it now
- [x] Test I didn't break command line z88dk-gdb
- [x] Test I didn't break ticks
- [x] Test on Visual Studio Code

Would appreciate any eyes ahead of time.

### DIFF
--- a/src/ticks/debugger.c
+++ b/src/ticks/debugger.c
@@ -1228,40 +1228,35 @@ static int cmd_watch(int argc, char **argv)
         int value = parse_address(argv[2], &corrected_source);
 
         if ( value != -1 ) {
-            elem = calloc(1, sizeof(*elem));
-            elem->number = next_breakpoint_number++;
-            elem->type = breakwrite ? BREAK_WRITE : BREAK_READ;
-            elem->value = value;
-            elem->enabled = 1;
-            LL_APPEND(watchpoints, elem);
-            bk.console("Adding %s watchpoint at '%s' $%04x (%s)\n",breakwrite ? "write" : "read", corrected_source, value,  resolve_to_label(value));
+            elem = add_watchpoint(breakwrite ? BREAK_WRITE : BREAK_READ, value);
+            bk.console("Adding %s watchpoint at '%s' $%04x (%s)\n",
+                breakwrite ? "write" : "read", corrected_source, value,  resolve_to_label(value));
         } else {
             bk.console("Cannot set watchpoint on '%s'\n", corrected_source);
         }
     } else if ( argc == 3 && strcmp(argv[1],"delete") == 0 ) {
-        int num = atoi(argv[2]);
-        delete_breakpoint_by_number_and_log(num);
+        breakpoint *elem = find_watchpoint(atoi(argv[2]));
+        if (elem) {
+            bk.console("Deleting watchpoint %d \n", atoi(argv[2]));
+            delete_watchpoint(elem);
+        } else {
+            bk.console("Unknown watchpoint\n");
+        }
     } else if ( argc == 3 && strcmp(argv[1],"disable") == 0 ) {
-        int num = atoi(argv[2]);
-        breakpoint *elem;
-        LL_FOREACH(watchpoints, elem) {
-            num--;
-            if ( num == 0 ) {
-                bk.console("Disabling watchpoint %d\n",atoi(argv[2]));
-                elem->enabled = 0;
-                break;
-            }
+        breakpoint *elem = find_watchpoint(atoi(argv[2]));
+        if (elem) {
+            bk.console("Disabling watchpoint %d\n", atoi(argv[2]));
+            elem->enabled = 0;
+        } else {
+            bk.console("Unknown watchpoint\n");
         }
     } else if ( argc == 3 && strcmp(argv[1],"enable") == 0 ) {
-        int num = atoi(argv[2]);
-        breakpoint *elem;
-        LL_FOREACH(watchpoints, elem) {
-            num--;
-            if ( num == 0 ) {
-                bk.console("Enabling watchpoint %d\n",atoi(argv[2]));
-                elem->enabled = 1;
-                break;
-            }
+        breakpoint *elem = find_watchpoint(atoi(argv[2]));
+        if (elem) {
+            bk.console("Enabling watchpoint %d\n",atoi(argv[2]));
+            elem->enabled = 1;
+        } else {
+            bk.console("Unknown watchpoint\n");
         }
     } 
     return 0;
--- a/src/ticks/debugger.c
+++ b/src/ticks/debugger.c
@@ -740,7 +740,7 @@ void debug_lookup_symbol(struct lookup_t* lookup, struct expression_result_t* re
         // try static locals
         char sname[128];
         sprintf(sname, "st_%s_%s", fn->function_name, lookup->symbol_name);
-        debug_sym_symbol *s = cdb_find_symbol(sname);
+        debug_sym_symbol *s = cdb_find_symbol(sname, fp->filename);
         if (s != NULL && s->address_space.address_space == 'E') {
             debug_get_symbol_value_expression(s, fp, result);
             debug_stack_frames_free(first_frame_pointer);
@@ -749,7 +749,7 @@ void debug_lookup_symbol(struct lookup_t* lookup, struct expression_result_t* re
     }
 
     // try globals
-    debug_sym_symbol* s = cdb_find_symbol(lookup->symbol_name);
+    debug_sym_symbol* s = cdb_find_symbol(lookup->symbol_name, fp->filename);
     if (s != NULL && s->address_space.address_space == 'E') {
         debug_get_symbol_value_expression(s, fp, result);
         debug_stack_frames_free(first_frame_pointer);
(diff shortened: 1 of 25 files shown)


src/ticks/debugger.c
```diff
@@ -1215,33 +1235,26 @@ static int cmd_watch(int argc, char **argv)
...
-        breakpoint *elem;
-        LL_FOREACH(watchpoints, elem) {
-            num--;
-            if ( num == 0 ) {
-                printf("Deleting watchpoint %d\n",atoi(argv[2]));
-                delete_breakpoint_by_number(elem);
-                break;
-            }
-        }
+        delete_breakpoint_by_number_and_log(num);
```
suborb, line 1250: I think this will end up deleting a breakpoint rather than the watch point?

src/ticks/debugger.c
```diff
@@ -1733,3 +1747,38 @@ static void print_hotspots()
...
+    va_end(args);
+
+    if (len < 0) {
+        /* vsnprintf failed */
+        return;
+    } else {
+        /* Declare a character buffer for the formatted string */
+        char formatted[len + 1];
+
+        /* Initialize a variable argument list */
```
suborb, line 1772: I need to check another project but I’ve got a memory of one platform not liking the reuse of a va_list - a va_copy was needed.

src/ticks/debugger.c
```diff
@@ -1733,3 +1747,38 @@ static void print_hotspots()
...
+
+    /* End using variable argument list */
+    va_end(args);
+
+    if (len < 0) {
+        /* vsnprintf failed */
+        return;
+    } else {
+        /* Declare a character buffer for the formatted string */
+        char formatted[len + 1];
```
suborb, line 1770: We need to check visual studio compatibility.

src/ticks/debugger_mi2.c
```diff
@@ -0,0 +1,1359 @@
...
+static void cmd_stack_list_frames(int argc, char **argv) {
+
+    struct debugger_regs_t regs;
+    bk.get_regs(&regs);
+
+    uint16_t stack = regs.sp;
+    uint16_t initial_stack = stack;
+    uint16_t at = bk.pc();
+
+    char* dump_buffer = malloc(65536);
```
suborb, line 643: Using utstring removes the potential/theoretical buffer overflow

src/ticks/debugger_mi2.c
```diff
@@ -0,0 +1,1359 @@
...
+    mi2_printf_response("done,value=\"%s\"", value);
+}
+
+static void resolve_expression_result_to_child(
+    const char* member, struct expression_result_t* result, char* ptr, int all_values)
+{
+    if (is_expression_result_error(result)) {
+        sprintf(ptr, "child={exp=\"%s\",type=\"<error>\",value=\"%s\"}", member, result->as_error);
+    } else {
+        char type[255] = {};
```
suborb, line 432: I think visual studio falls over this syntax?

src/ticks/debugger.c
desertkun: Haha this was stolen fresh from the internet, if you've got a better alternative feel free to suggest :D 

The goal is to basically have two implementations of printf(fmt, ...), because with mi/2 you can't just printf, there's syntax to follow

src/ticks/debugger_mi2.c
desertkun: Thanks, incredible tool!

src/ticks/debugger.c
suborb: You need something like this:

```
// Copes with some versions of visual studio
#ifndef va_copy
#define va_copy(a,b) (a)=(b)
#endif


....

void stdout_log(const char *fmt, ...)
 {
     va_list args;
     va_list args2;
     int len;

     /* Initialize a variable argument list */
     va_start(args, fmt);
     va_copy(args2,args)

     /* Get length of format including arguments */
     len = vsnprintf(NULL, 0, fmt, args2);

     /* End using variable argument list */
     va_end(arg2);
...
Rest of code as it was
```

src/ticks/debugger.c
desertkun: Thanks, resolved.

src/ticks/debugger.c
desertkun: Ok, I made the change. Works on my MAC. Something like this?

desertkun:
@suborb  I think I figured out why don't global symbols resolve – compiler simply does not omit such info :( 

See https://github.com/desertkun/sym-test (do `make` and notice nothing about `global_kek`'s type in the `a.map`)

desertkun:
Ok that was a simple fix. For some reason, export of static symbols was previously excluded. Do you happen to know why?

suborb:
> Ok that was a simple fix. For some reason, export of static symbols was previously excluded. Do you happen to know why?

I suspect that if you have multiple files with the same symbol in them you might have a conflict at linking time. 

They should end up in .map in anycase - the i_ ones do.

desertkun:
> > Ok that was a simple fix. For some reason, export of static symbols was previously excluded. Do you happen to know why?
> 
> I suspect that if you have multiple files with the same symbol in them you might have a conflict at linking time.
> 
> They should end up in .map in anycase - the i_ ones do.

I see, there's confusion with `LSTATIC` and `STATIK`. Regardless, I think we want to "export" both, and fix any problem with matching those symbols, because I think we should be able to access static variables as well as exported ones. Ugh.

desertkun:
OK got it working, basically I've split both symbol definitions and symbols themselves into two arrays, one for global variables and one for file-local ones. So now when looking up for a symbol, a current filename is taken as a hint to check into file-local array first, then on the global one.

It is context-aware, too, e.g. when you have two statics with the same name on different frames, `info variables` and `print static_var` show values from appropriate frames.

desertkun:
I am on a roll here, got Visual Studio Code working. Any IDE you use I could check?

desertkun:
OK seems like I made everything I wanted at this stage. Since this is a giant change maybe we can ask someone else to glance over? Otherwise it is good to go I think.

src/ticks/syms.c
```diff
@@ -65,7 +65,7 @@ static int symbol_compare(const void *p1, const void *p2)
     return s2->address - s1->address;
 }
 
-static char read_symbol_buf[2048];
+static char read_symbol_buf[8192];
```
desertkun, line 68: One particularly long class simply didn't fit
```
F/Users/desertkun/Documents/Work/zx-sandbox/client/include/client_map.h$map_client_object_t[({0}S:S$object$0_0$0({8}STmap_object_t:S),Z,0,0)({8}S:S$block_bottom_m_2$0_0$0({1}SC:U),Z,0,0)({9}S:S$block_bottom_m_1$0_0$0({1}SC:U),Z,0,0)({10}S:S$block_bottom_p_0$0_0$0({1}SC:U),Z,0,0)({11}S:S$block_bottom_p_1$0_0$0({1}SC:U),Z,0,0)({12}S:S$block_bottom_p_2$0_0$0({1}SC:U),Z,0,0)({13}S:S$block_bottom_p_3$0_0$0({1}SC:U),Z,0,0)({14}S:S$block_right_1$0_0$0({1}SC:U),Z,0,0)({15}S:S$block_right0$0_0$0({1}SC:U),Z,0,0)({16}S:S$block_right1$0_0$0({1}SC:U),Z,0,0)({17}S:S$block_left_1$0_0$0({1}SC:U),Z,0,0)({18}S:S$block_left0$0_0$0({1}SC:U),Z,0,0)({19}S:S$block_left1$0_0$0({1}SC:U),Z,0,0)({20}S:S$target_x$0_0$0({2}SI:U),Z,0,0)({22}S:S$target_y$0_0$0({2}SI:U),Z,0,0)({24}S:S$jump_potential$0_0$0({1}SC:U),Z,0,0)({25}S:S$force$0_0$0({1}SC:U),Z,0,0)({26}S:S$sprite_data$0_0$0({2}DG,SC:U),Z,0,0)({28}S:S$sprites$0_0$0({2}DG,STobject_sprites_t:S),Z,0,0)({30}S:S$flags$0_0$0({1}SC:U),Z,0,0)({31}S:S$dirty$0_0$0({1}SC:U),Z,0,0)({32}S:S$offset_x$0_0$0({1}SC:U),Z,0,0)({33}S:S$offset_y$0_0$0({1}SC:U),Z,0,0)]
```
That's over 1024 and you have to essentially double that because of encoding.

desertkun:
Rebased because I apparently broke master (fixed with 2e07b55927e48fe5415c76e79649606ea001b268)

PR #2027, opened 2022-06-07
add usraddr option to appmake

derekfountain:
As per my [request](https://www.z88dk.org/forum/viewtopic.php?p=20631#p20631) on the forum, this adds an option to appmake for ZX to set the RANDOMISE USR address in the BASIC loader. This is useful if the code start point isn't the same address it loads into.

It's a simple addition, adding usr_address to the zx_tape structure and (optionally) populating it from the command line. If the value is not at the -1 default in the save-loader code then it's picked up and used. Otherwise the loader code just does what it did before.

Tested on a 48K loader, the turbo loader and the +3 loader.

I'm not too familiar with the code, so if I've missed something let me know. :)

### DIFF
--- a/src/appmake/zx-util.c
+++ b/src/appmake/zx-util.c
@@ -280,6 +280,10 @@ int zx_tape(struct zx_common *zxc, struct zx_tape *zxt, struct banked_memory *me
                 exit_log(1,"Could not find parameter ZORG (not z88dk compiled?)\n");
             }
         }
+        if (zxt->usr_address == -1 )
+        {
+          zxt->usr_address = pos;
+        }
 
         if ((fpin = fopen_bin(zxc->binname, zxc->crtfile)) == NULL) {
             exit_log(1, "Can't open input file %s\n", zxc->binname);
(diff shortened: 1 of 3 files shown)


derekfountain:
Test:

Create a file called ea1.c with this in it:

```
#include <stdio.h>

void print_world( void );

void main( void )
{
  printf("Hello, ");
  print_world();
  return;
}
```

Create a file called ea2.c with this in it:

```
#include <stdio.h>

void print_world( void )
{
  printf("world\n");
  return;
}
```

Create a file called ea_sections.asm with this in it:

```
SECTION contended
org 30000

defb 0
```

Build the ea1 code with this:

`zcc +zx -vn -compiler=sdcc -clib=sdcc_iy ea_sections.asm ea1.c -c`

and the ea2 code with this:

`zcc +zx -vn -compiler=sdcc -clib=sdcc_iy ea_sections.asm ea2.c -c --codesegcontended`

Link with:

`zcc +zx -vn -clib=sdcc_iy -m -startup=4 ea_sections.o ea1.o ea2.o -o ea`

Glue it together:
`
z88dk-appmake +glue -b ea --filler 0xDF`

Now you have a program with a section in contended memory, at 30000. The main start point is 32768 as normal. So without the patch appmake will set it to load at 30000, then RANDOMISE USR 30000, which won't work - that'll start the program at the subroutine, which happens to be at the load address. With my patch you can do this:

`z88dk-appmake +zx -b ea__.bin --org 30000 --usraddr 32768 --blockname ea -o ea.tap`

This changes the default behaviour to set the RANDOMISE USR to 32768, which is what it needs to do.

suborb:
Thanks for this, looks good. 

I'll add a pragma at some point later so that a command line option isn't needed and allow startup at a linked address.

PR #2028, opened 2022-06-09
Misc. fixes

iratahack:
This PR contains the following fixes:

1) appmake - Segmentation fault on MSX targets. NULL was being passed for the crtfile.
2) asmstyle.pl - Move conditional statements to first indentation level to be more compatible with simple assemblers. 

IH.


suborb:
asmstyle is just a tool - should have no impact on our tree.

PR #2029, opened 2022-06-10
mi/2: restore program upon connection

desertkun:
Tested on CLion/VSCode and [Fuse fork](https://github.com/speccytools/fuse-for-macosx).

Basically, when z88dk-gdb is fed with a debugging file `folder/<filename>.map`, it checks for neighbouring file `folder/<filename>.bin` or just `folder/<filename>`. If such file exists, upon connections, z88dk-gdb uploads (restores) the binary onto remote, so there is no need to upload such binary to remote in some other way now.

### DIFF
--- a/src/ticks/debugger.c
+++ b/src/ticks/debugger.c
@@ -6,6 +6,8 @@
 #include <signal.h>
 #ifdef WIN32
 #include <io.h>
+#define F_OK 0
+#define access _access
 #else
 #include <unistd.h>                         // For declarations of isatty()
 #include <stdarg.h>
@@ -24,7 +26,7 @@
 #include "srcfile.h"
 #include "exp_engine.h"
 #include "breakpoints.h"
-
+#include "../common/dirname.h"
 
 #define HISTORY_FILE ".ticks_history.txt"
 
@@ -210,6 +212,7 @@ static int hotspots[65536];
 static int hotspots_t[65536];
 size_t current_frame = 0;
 static int last_stacktrace_at = 0;
+UT_string* pending_executable_binary = NULL;
 
 static int interact_with_tty = 0;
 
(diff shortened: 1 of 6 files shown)


src/ticks/debugger.c
```diff
@@ -228,6 +232,64 @@ void debugger_init()
...
 }
 
+uint8_t debugger_read_symbol_file(char* symbol_file) {
+    if (access(symbol_file, F_OK)) {
+        return 1;
+    }
+
+    read_symbol_file(symbol_file);
+
+    char* filename = basename(symbol_file);
```
suborb, line 242: There's zbasename() in common/dirname.c for compatibility purposes - I don't think basename exists for windows.

src/ticks/debugger.c
```diff
@@ -228,6 +232,64 @@ void debugger_init()
...
+    char* ext;
+    if ((ext = strrchr(filename, '.'))) {
+        int filename_w_ext_len = (int)(ext - filename);
+        char* dir = dirname(symbol_file);
```
suborb, line 246: Similarly there's zdirname

src/ticks/debugger.c
desertkun: so `src/common` is a separate library, are you suggesting linking it up, or copying those down? Seems like if I do copy, I only need to copy one file.

suborb: Just add the files to the makefile

desertkun:
Fixed.

desertkun:
@suborb do you think this is free to go?

feilipu:
If it's working as you desire, I can merge it for you. Unless you want further review?

desertkun:
I guess it's fine.

PR #2030, opened 2022-06-11
The initial support for the Corvette

litwr2:
This PR has added the next features:
1) the base support for the Corvette CP/M console applications;
2) support for a non-bootable disk format.

So you can now compile and use applications like "World" or "Adventure".  The latter still has an issue about screen clear handling because the Corvette uses code 31 instead of 12 to clear screen.  :(

Disclaimers:
1) file src/appmake/corvette.c is in a stub stage now, it does nothing.  I don't know how this thing work;
2) I added
```
#include <iomanip>
```
to src/z80asm/src/cpp/utils.h because I could not compile without this line.


suborb:
If the corvette file is unused, can you remove it? - I can see the code doesn't match up with the geometry of the disc.

litwr2:
> If the corvette file is unused, can you remove it? - I can see the code doesn't match up with the geometry of the disc.

Without this file appmake doesn't work.  So it is mandatory.  Sorry, I don't understand that the code doesn't match up with the geometry of the disc.  Would you like to provide some details?

PR #2032, opened 2022-06-15
debugger_gdb.c: OpenBSD support

jg1uaa:
- CLOCK_MONOTONIC_RAW is for Linux, use CLOCK_MONOTONIC instead  if it is not available
- %n format of printf() raises SIGABRT on OpenBSD,  refactor with strlen()


desertkun:
Can you rebase your local branch? It seems it has conflicts. It's weird %n gives a segfault, it's used like 100 times here.

Also congrats on actually using the thing, didn't see any evidence of anyone using it so far.

desertkun:
@suborb haha thanks for reactively fixing the time issue. As per printf, I've been meaning to ditch static arrays in favor of ut_string, let me do that there as well.

jg1uaa:
wow, how do I treat this PR?

jg1uaa:
I do not use gdb.
But CLOCK_MONOTONIC_RAW makes compile error and printf(%n) puts warning.
Former issue is is already fixed, latter will be potential problem so I want to fix now.

jg1uaa:
Ok, this PR is close. Thanks!

PR #2048, opened 2022-07-22
Shared SP1 between classic and newlib

suborb:
Merge after snapshot


zx70:
Can we confirm the merge for this one ?

suborb:
Let's go!

PR #2051, opened 2022-07-23
zcc: +config at any place, support for -xc

desertkun:
This change allows to do the following:

args.txt
```
-xc +zx -debug -fpch-preprocess -v -dD -E -D___CIDR_DEFINITIONS_END
```
test (NO extension)
```
int main() {}
```
then
```
zcc @args.txt test
```

As you can notice, the following changes:

1. `+zx` can be specified as not-first-argument
2. `-xc` is now supported, makes any file type to be treated as C files. Note that there is `-x <filename>` option to build a library, but those are different.
3. `@args.txt` can not only specify files to compile, but also any arguments
4. `-dD` argument tu dump all generated defines.

All this trickery been done to convince CLion that zcc is a real compiler. I am hoping, in consequent PRs, suggest a CMake toolchain file and make zcc to be supported by CLion out-of-the-box.

### DIFF
--- a/include/sys/compiler.h
+++ b/include/sys/compiler.h
@@ -3,6 +3,26 @@
 #ifndef __SYS_COMPILER_H__
 #define __SYS_COMPILER_H__
 
+#if defined(__CLION_IDE__) | defined(__INTELLISENSE__)
+
+#define __LIB__
+#define __SAVEFRAME__
+#define __z88dk_fastcall
+#define __FASTCALL__
+#define __CALLEE__
+#define __SCCZ80
+#define __Z80
+#define __naked
+#define __z88dk_callee
+#define __stdc
+#define __smallc
+#define __preserves_regs
+#define __no_z88dk_declspec
+#define __at(x)
+#define __sfr
+#define __vasmallc
+
+#else
 
 /* Temporary fix to turn off features not supported by sdcc */
 #if __SDCC | __clang__
@@ -25,6 +45,8 @@
 #else
 #define __vasmallc __smallc
 #define __z88dk_deprecated
+#endif
+
 #endif
 
 #ifdef __8080__
--- a/src/zcc/zcc.c
+++ b/src/zcc/zcc.c
@@ -24,6 +24,8 @@
 #include        <inttypes.h>
 #include        <time.h>
 #include        <sys/stat.h>
+#include        "uthash.h"
+#include        "utlist.h"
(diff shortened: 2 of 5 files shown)


include/sys/compiler.h
```diff
@@ -3,6 +3,23 @@
...
+#define __z88dk_fastcall
+#define __FASTCALL__
+#define __CALLEE__
+#define __SCCZ80
+#define __Z80
+#define __naked
+#define __z88dk_callee
+#define __stdc
+#define __preserves_regs
+#define __no_z88dk_declspec
```
suborb, line 19: You probably want to define away  __sfr as well?

src/zcc/zcc.c
```diff
@@ -24,6 +24,8 @@
 #include        <inttypes.h>
 #include        <time.h>
 #include        <sys/stat.h>
+#include        "../../ext/uthash/include/uthash.h"
```
suborb, line 27: Can we just add ext/uthash/include to the header search path?

src/zcc/zcc.c
```diff
@@ -837,25 +879,52 @@ int main(int argc, char **argv)
...
 
     setup_default_configuration();
 
-    gc = find_zcc_config_fileFile(argv[0], argv[gc], gc, config_filename, sizeof(config_filename));
+    char configuration[FILENAME_MAX] = {0};
+
+    for (gargc = 1; gargc < argc; gargc++) {
+        char* aa = argv[gargc];
+        if (aa[0] == '+') {
+            strcpy(configuration, aa);
```
suborb, line 905: It's thankfully long dead, but HP-UX 9 had FILENAME_MAX defined as 14...

suborb:
zcc is a bit of a twisted web at times.

I need to look at the changed file as a whole to see if there's a problem. However I'm thinking about the command line aliases at the moment. They're defined in the config file and so we need to read the config file before we can process the arguments correctly.

desertkun:
Well that's what I am doing. I am looking for `+conf` argument first, in argv or otherwise, and then proceed as usual.

`zcc -Dtest +zx`
and
`zcc +zx -Dtest`
are equal in that respect.

desertkun:
Looking forward for this to be reviewed, prepping a next PR with cmake toolchain introduction!

suborb:
Merge when you like!

desertkun:
Thanks

PR #2056, opened 2022-07-25
Check build deps

jorgegv:
Add a check to build script so that submodule dependencies are met before starting build


derekfountain:
This seems to assume the user has pulled the source tree from git. That's not useful if the user has a problem because they downloaded the auto-generated ZIP archive (which is what I did). In that case they need to be pointed towards the correct archive file to download.

Because Github is not exactly clear on this, and the link to the file the user wants is likely below the fold on the release page, I think the error message should spell out exactly what they should do:

_Dependencies missing! It is likely you downloaded an autogenerated archive of the source, or cloned from Github without the --recursive option.

Either:
  Download the complete archive from: https://github.com/z88dk/z88dk/releases/download/v2.2/z88dk-src-2.2.tgz
or
 Re-clone from Github using: git clone --recursive git@github.com:z88dk/z88dk.git_

derekfountain:
Also, and I need to double check this when I get chance, but I think the Git config files being checked for aren't actually in the TGZ file, so I think this might give a false error even if the correct source archive has been used. Maybe just check whether there are zero files in the subdirectory?

jorgegv:
Boht suggestions implemented, thanks @derekfountain

jorgegv:
Yes, I had already done the last two tests also.

Great.

suborb:
Great, thank you both for this!

PR #2062, opened 2022-08-03
z88dk-dis: Fixed bug in section name handling

iratahack:
Two issues are addressed in this commit.

* A symbol for section __head was being added twice, once for the
actual head and once for __size since the code clobbered argv[0]
when looking-up the symbol for __head.

* Section names can be a simple 1 word name like "code" or they
can be multiple words seperated by '_' (underscore), e.g. "code_2".
This is now handled in address_is_code().


src/ticks/syms.c
```diff
@@ -118,9 +123,7 @@ void read_symbol_file(char *filename)
                         sym->symtype = SYM_ADDRESS;
                     }
                     if (sym->symtype == SYM_ADDRESS) {
-                        if ( sym->address >= 0 && sym->address <= 65535 ) {
-                            LL_APPEND(symbols[sym->address % SYM_TAB_SIZE], sym);
-                        }
+                        LL_APPEND(symbols[sym->address % SYM_TAB_SIZE], sym);
```
iratahack, line 126: No need to check sym->address as it is used as mod(SYM_TAB_SIZE)

desertkun:
To better understand the issue, what's the use case?

iratahack:
@desertkun , please see below...

Example: the assembly code below...

```
	section	CODE
        org     0x2000
start:
        ld      bc, 0x7ffd
        ld      a, 0x30                 ; ROM1 + Lock register access
        out     (c), a

        ld      hl, message
nextChar:
        ld      a, (hl)
        or      a
        jr      z, exit
        rst     0x10
        inc     hl
        jr      nextChar

exit:
        xor     a                       ; Clear carry
        ret

        section RODATA
message:
        db      "Memory banking disabled.", 0x0d, 0
```

Built using `z88dk-z80asm -mz80 -b -m -o48 48.asm` does not disassembly correctly using `z88dk-dis -x 48.map -o __CODE_head 48_CODE.bin`. The output looks like this...

```
                    ld      bc,$7ffd                        ;[0014] 01 fd 7f
                    ld      a,$30                           ;[0017] 3e 30
                    out     (c),a                           ;[0019] ed 79
                    ld      hl,$2014                        ;[001b] 21 14 20
                    ld      a,(hl)                          ;[001e] 7e
                    or      a                               ;[001f] b7
                    jr      z,$0026                         ;[0020] 28 04
                 ...
```
Note the start address is 0x14. This is because 2 symbols have been added for `__CODE_head`. One has the correct address of 0x2000 and the other does not. The symbol with the incorrect address (0x14) is added because argv[0] is clobbered, '_size' is changed to '_head' when creating the section entry. This modified name is then added to the global symbol list. Instead of being called __CODE_size it is called __CODE_head.

When the -o command line option tries to find the user specified symbol it finds the incorrect one with an address of 0x14.

With the above issue resolved, the next issue is in address_is_code(). This function looks for `code_` to determine if an address is in a code section. However, since the leading `__` and trailing `_head`are removed from the symbol, the symbol name is `code` not `code_` which this function looks for. Consequently, the entire file is treated as data and not code.

Checking for `code` and `code_` in address_is_code() addresses both `section code` and `section code_blah`.

snap/snapcraft.yaml
```diff
@@ -30,7 +30,7 @@ parts:
       - BUILD_SDCC: "1"
       - BUILD_SDCC_HTTP: "1"
     override-build: |
-      cpanm -n -L $SNAPCRAFT_PART_INSTALL App::Prove Capture::Tiny CPU::Z80::Assembler Data::Dump Data::HexDump File::Path List::Uniq Modern::Perl Object::Tiny::RW Regexp::Common Test::Harness Text::Diff Text::Table YAML::Tiny
+      cpanm -n -L $SNAPCRAFT_PART_INSTALL Path::Tiny App::Prove Capture::Tiny CPU::Z80::Assembler Data::Dump Data::HexDump File::Path List::Uniq Modern::Perl Object::Tiny::RW Regexp::Common Test::Harness Text::Diff Text::Table YAML::Tiny
```
iratahack, line 33: Path::Tiny is required to run z88dk-asmstyle

suborb:
Sorry it's taken so long to get round to merging..

PR #2074, opened 2022-08-26
LBSearch previously wasn't always returning items that should have...

pjshumphreys:
been found; switched from upper bound address to base+offset algorithm

### DIFF
--- a/libsrc/stdlib/Lbsearch.asm
+++ b/libsrc/stdlib/Lbsearch.asm
@@ -35,9 +35,9 @@ EXTERN l_jpix
 
 .slice
    push hl      ; Get the address at the halfway index + the base address
-   sra h
-   rr l
-   add hl,hl
+   ld a, $fe
+   and l
+   ld l,a
    add hl,de
 
 .compare            ; is key < datum?
--- a/libsrc/stdlib/Lbsearch.asm
+++ b/libsrc/stdlib/Lbsearch.asm
@@ -35,9 +35,7 @@ EXTERN l_jpix
 
 .slice
    push hl      ; Get the address at the halfway index + the base address
-   ld a, $fe
-   and l
-   ld l,a
+   res 0,l      ; Integer division of hl by 2 (to get the midpoint item index) then multiply it 2 because each pointer is 2 bytes.
    add hl,de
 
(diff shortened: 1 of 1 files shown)


pjshumphreys:
I have some sample code that shows how the previous code could fail at https://www.z88dk.org/forum/viewtopic.php?f=12&t=11681

libsrc/stdlib/Lbsearch.asm
```diff
@@ -24,57 +24,42 @@ EXTERN l_jpix
...
-   rr h
-   rr l             ; HL = unrounded address of middle item
-
-   ld a,l           ; shenanigans to ensure HL aligns on item
-   xor e
-   rra
-   jr nc, compare
-   dec hl
+   push hl      ; Get the address at the halfway index + the base address
+   ld a, $fe
```
ped7g, line 38: is this for Z80? If yes, you can do also `res 0,l` to make HL even in 8T and two bytes.

pjshumphreys: It will need to work on an 8080 as well as on a z80 I think

pjshumphreys: Then again, the "rr h" instruction that was here previously isn't available on an 8080 either, so this function has never been 8080 friendly. I'll probably change the code to your suggestion to save a few T states. Every T state will count if this code is to search through large arrays

suborb: https://github.com/z88dk/z88dk/wiki/Classic-8080#stdlibh - bsearch is not available on 808x

pjshumphreys: This is a change to the l_bsearch function in stdlib. z88dk doesn't have bsearch at all it seems. The sample code I posted on the forums used a preprocessor macro to filter out the extra size parameter that bsearch has but that l_bsearch doesn't. bsearch on gcc can operate on arrays of pointers the same as l_bsearch (only) can.

pjshumphreys: Ok @ped7g, I've run my tests and the opcode suggested works fine.

suborb:
Test to follow in next commit

Issue #2071, opened 2022-08-19
Little opt for CP/M fcntl


zx70:
This one started as a small intervention on the CP/M build, but I under-estimated the file bindings.
Sorry for the recent build warnings, I think it is all sorted now.

- Reduced number of warnings in the fcntl library build
- quick and dirty insertion of the CALLEE variant of the bdos() function, together with the "__bdos" centralization it reduces the library size
- __bdos entry usage extended to all the CP/M fcntl lib to centralize and complete the IX preservation, this means that an equivalent label is now declared also for the MSX target
- bios() doubled to biosh(), the latter has different return values, getting HL rather than A

feilipu:
Not sure if you're up for some more optimisation, but...

Some time ago I tried to optimise the `read.c` and `write.c` to use `BDOS` sequential read and write where the request matched the sector alignment and size (128 Bytes). It did work most of the time, but there were corner cases (eg. `lseek`) where it didn't work, so it [had to be backed out](https://github.com/z88dk/z88dk/commit/ccea56eafdbdfab59e59ec1b5357544636e3dab6).

The reason that I thought it could work is that CP/M `PIP` exclusively uses BDOS sequential read and write calls, and `PIP` remains, to my knowledge, the fastest way of copying a file in CP/M.

The PIP PLM source is available [from here](http://www.cpm.z80.de/download/cpm2-plm.zip). I've added it to this issue. [PIP.PLM](https://github.com/z88dk/z88dk/files/9386090/PIP.TXT)

Perhaps you can see how it manages to use the BDOS sequential read and write calls?
(and entirely avoids using BDOS random read and write calls)?

Don't know. I couldn't solve it. :man_shrugging: 

```plm
DISKREAD: PROCEDURE(FCB) BYTE;

    DECLARE FCB ADDRESS;
    RETURN MON2(20,FCB);
    END DISKREAD;

DISKWRITE: PROCEDURE(FCB) BYTE;
    DECLARE FCB ADDRESS;
    RETURN MON2(21,FCB);
    END DISKWRITE;

  FILLSOURCE: PROCEDURE;
    /* FILL THE SOURCE BUFFERS */
    DECLARE (I,J) BYTE;
    NSOURCE = 0;
    CALL SELECT(SDISK);
    CALL SETSUSER; /* SOURCE USER NUMBER SET */
        DO I = 0 TO NBUF;
        /* SET DMA ADDRESS TO NEXT BUFFER POSIITION */
[...]

PR #2080, opened 2022-09-13
bdosh() function for CP/M

zx70:
A BDOS call variant providing a more detailed error code on output (HW error code on MSB as in "Portable CP/M").


zx70:
#2071

Issue #683, opened 2018-04-08
[clg][KC] clg() produces freezes and random glitches on most KC sub-models


Fabrizio-Caruso:
Hi

I have successfully managed to use the sprite engine with KC! 
I have realized that clg() produces random wrong effects and freezes on most KC sub-models at least when used inside CROSS CHASE (tested with jkcemu):
- on KC/4 -> it breaks the input
- on KC/5 -> on the second execution it locks the computer
Similar problems and/or graphics corruptions are also visibile on these two and other sub-models.

zx70:
I can't really explain the impact of clg(), it is a simple ROM call (like the one used all the time by pixeladdress).
In any case I replaced it with a dumb loop clearing the graphics memory (top-left position identified by pixeladdress, then zeroing the next 10240 bytes).
I'm not altering the color attributes to limit the impact, let me know if the tomorrow's build will solve.

zx70:
I think I'm getting closer to the root cause, if it is the way I think (ix register affecting the os stability) the last fix won't help much.

zx70:
so, googling with the right keywords it is easy to confirm..
https://sourceforge.net/p/sdcc/mailman/message/33036952/

suborb:
Aha, lets make kc an ixiy target then  - I presume iy is available?

zx70:
My first experiment with the ixiy workaround and the iy library failed, the resulting stability looks worse than with the existing interrupt wrappers protection.   I'm investigating on possible mistakes,

Fabrizio-Caruso:
@zx70, some initial tests (only KC 85/5) show that the "fix" has worked by probably avoiding by sheer chance the use of a forbidden register.
The root cause is still there, though, if I understand well.

Fabrizio-Caruso:
@zx70 
The new clg() "fixes" all 5 sub-models:
HC 900 *
KC85/2 *
KC85/3 *
KC85/4
KC85/5
The first 3 models (*) have a little glitch: the cursor remains visible on the screen.

zx70:
I'm afraid it is just a matter of probability,  in example dstar.c starts correctly clearing the screen perfectly.
Then the game field is drawn and the ball moves under the keyboard control as expected.
On any next clg() the picture gets garbled, probably after the getk() usage random memory areas are corrupted, or some kind of overflow occurs.   In the "CAOS reference" document it is stated:

Restrictions
[...]

PR #2092, opened 2022-10-01
cleanup clg(), crt_init und crt_exit for KC85/2..

Hojoe42:
This is for issue #683. There were some changes at clg.asm, but they didn't help. The KC 85/2-5 uses the IX register for interrupts, keyboard access and switching graphics memory.
The clg() implementation clears only the pixel memory, but not the color memory. As a result the cursor position was changed to a invalid value. (the cursor uses a different color)

I change clg() back to the CAOS implementation and add some more initialization stuff. For KC85/2-3 set the default window to max size and for KC85/4 switch to the pixel memory. The os expects the pixel memory is active.

After the changes for the IX register the setup code for the interrupts aren't needed any more. I have removed this code.

I tested with dstar and some other programs for KC85/2-5. 

One question: do you have some code styles for tabs and spaces? I am not sure what I have to use.

### DIFF
--- a/lib/target/kc/classic/kc_crt0.asm
+++ b/lib/target/kc/classic/kc_crt0.asm
@@ -4,9 +4,6 @@
 ;       Stefano Bodrato October 2016
 ;		few hints were found on the sdcc lib by Andreas Ziermann and Bert Lange
 ;
-;		OPTIMIZATIONS:
-;		#pragma output nosound   - Save few bytes not preserving IX on tape and sound interrupts
-;
 ; - - - - - - -
 ;
 ;       $Id: kc_crt0.asm,v 1.2 2016-10-10 07:09:14 stefano Exp $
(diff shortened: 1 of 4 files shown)


feilipu:
> One question: do you have some code styles for tabs and spaces? I am not sure what I have to use.

I think the style has changed quite a bit over the past decades and authors.
There's some tabs, some 3 spaces, and some 4 spaces.

The assembly is generated with tabs, because `\t` is easier to parse and generate.

AFAIK the path is slowly moving to 4 spaces - or perhaps that's just me. :stuck_out_tongue_closed_eyes:

Hojoe42:
> AFAIK the path is slowly moving to 4 spaces - or perhaps that's just me. 😝

Good to know. Thanks

suborb:
The interrupt interposers were present for a couple of reasons:

- a couple of library routines unfortunately use both index registers. 
- sdcc will also use both index registers if it needs to

Hojoe42:
> The interrupt interposers were present for a couple of reasons:
> 
>     * a couple of library routines unfortunately use both index registers.

I saw your changes in https://github.com/z88dk/z88dk/commit/1937a9049d71f37856955cac2513e26eeeb57338. I thought this was a single case / mistake. 
What is with https://github.com/z88dk/z88dk/blob/master/libsrc/target/kc/graphics/swapgfxbk.asm or https://github.com/z88dk/z88dk/blob/master/libsrc/target/kc/stdio/generic_console.asm. Both files use the ix register to switch between pixel and color memory. 
How does other targets solve this problem?

>     * sdcc will also use both index registers if it needs to

In the past I tested with sdcc. I used `zcc +kc -compiler sdcc -Cs--reserve-regs-iy ...`. I don't know if this works.

I will restore the interrupt handling.

lib/target/kc/classic/kc_crt0.asm
```diff
@@ -6,14 +6,15 @@
 ;
 ;		OPTIMIZATIONS:
 ;		#pragma output nosound   - Save few bytes not preserving IX on tape and sound interrupts
+;		#pragma output nointerrupsaving   - do not save some interrupt vectors on startup (includes nosound)
```
Hojoe42, line 9: is `output`still right?

feilipu: https://github.com/z88dk/z88dk/wiki/CRT#pragma-overrides

might test in the assembly. Like...
https://github.com/z88dk/z88dk/wiki/Tool---z80asm---directives#ifdef-name--elifdef-name--else--endif

`IFDEF nointerrupsaving`

or 

https://github.com/z88dk/z88dk/wiki/Tool---z80asm---directives#ifndef-name--elifndef-name--else--endif
`IFNDEF nointerrupsaving`

Hojoe42:
I have restored the interrupt interposers and added code to deinstall the code on exit. Should I do something else?

feilipu:
If you have built and tested with both sdcc and sccz80 compilers then you should be ok.

I can't help you test, in sorry.

Hojoe42:
I compile the lib sources with `make clean` and `make`. I suppose that uses the sccz80. Is it possible to compile the libraries with sdcc?

Today I have compiled my test program with sccz80 and sdcc. Both compilers have compiled the program without any warning and the programs have worked in the emulator.

feilipu:
Ok. Then good to go.

PR #2094, opened 2022-10-04
Revert "1bit sound engine to save in zx tape format"

zx70:
Reverts z88dk/z88dk#2093


suborb:
I think you just needed the 8080 etc protection?

PR #2103, opened 2022-10-19
make l_bsearch and l_qsort work for 8080 and gbz80 CPUs

pjshumphreys:


libsrc/_DEVELOPMENT/l/8080/l_jpix_8080.asm
```diff
@@ -0,0 +1,11 @@
+SECTION data_clib
+
+PUBLIC l_jpix_8080
+
+l_jpix_8080:
+   db $c3 ;jp $0000
+
+fake_ix:
```
ped7g, line 8: is the `fake_ix` and split of jp even needed, if the `l_setix_8080` does access `l_jpix_8080` symbol any way? ( `ld (l_jpix_8080+1),a`)

Would make sense to *me* to do `jp some_error_trap_function` with ``l_jpix_8080` label and not even add the `fake_ix` symbol at all, it looks unused.

(I'm not very familiar with z88dk, so maybe I'm missing some technical benefit of this approach)

libsrc/stdlib/Lbsearch.asm
```diff
@@ -29,33 +34,76 @@ EXTERN l_jpix
 .loop
    ld a,h       ; Number of items still to check is 0? if so, return not found
    or l
-   jr nz, slice
+   jp nz, slice
```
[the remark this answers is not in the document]
pjshumphreys: I'll try it. This code would be being called quite a lot if bsearch is used, and I think freeing up even 1 t-state would still worth the trouble.

libsrc/_DEVELOPMENT/l/8080/l_jpix_8080.asm
pjshumphreys: As for "what should the default jump address be?". I don't think there's any z88dk general purpose, low level error handling default function available, and this code should never even be called without the code to set ix/iy it being run first. l_jpix_8080 should not be used by user written programs, only internally by z88dk's clib. I'll tell you what, I'll add a ret instruction to both l_jpix_8080 and l_jpiy_8080 and have the default addresses point to them. If in the future someone modifies the clib to call jpix_8080 without first calling setix_8080 first, it'll just return invalid results back to the calling code rather than jumping to $0000 which would cause a system restart. After that adding unit tests to the clib for new functionality would have to provide any further assurances.

pjshumphreys:
Ok, I hadded some changes based upon this review now

libsrc/_DEVELOPMENT/l/8080/l_jpix_8080.asm
pjshumphreys: Oh, and to clarify the technical benefit of this approach, the existing l_bsearch and l_qsort for the z80 were using the ix and iy registers. The 8080 and gbz80 CPUs don't have these registers :). Plus it would be much harder to do function pointers without something that performs the same function as them. The HL register pair is used by z88dk as a 16 bit scratchpad register, so using "jp (hl)" would be hard to make work as its value doesn't stay unchanged for long.

zx70 (approved):

ped7g (approved):

feilipu:
I'd just point out that the function naming seems a little redundant. The 8080 `l_setiy_8080` file (for example) is already in a newly created `l/8080/` directory. Wouldn't it be better just to use the same label `l_setiy` (for example) and then the main code wouldn't need to have conditionals added?

The correct function included would simply be selected based on which directories are assembled and linked into the library.

Another perhaps better option would be to not create another new directory just for these few functions, but rather to put them in the existing common function directory, with appropriate conditionals added and use the same labels (as above). https://github.com/z88dk/z88dk/tree/master/libsrc/_DEVELOPMENT/l/sccz80/9-common.

It has to be done that way (`l/sccz80/9-common` with conditionals) because unfortunately the [z80 code inherits the 8080 code](https://github.com/z88dk/z88dk/blob/master/libsrc/_DEVELOPMENT/l/sccz80/z80.lst#L79), so you can't just split two same label functions into the two respective directories.

pjshumphreys:
Ok, I've moved the 8080 compatible code into the z80 folder. I think it's kinda a shame that the z80 folder has asm in it that's specifically *not* for the z80 (where to my knowledge it didn't have before). However it would work on a z80, just less efficiently than by using the real ix and iy registers.

feilipu:
I was thinking too that there is no real reason to distinguish the function names with `_8080`. For most machines (i8085/gbz80) that would be wrong anyway.

IMHO, it would be ok to have one file per function, one function label, and just distinguish the between the code assembled using the machine type conditionals in the function.

That would simplify the main search code because it doesn't need conditionals in the main code flow, and is then easier to read and maintain.

Just a thought.

feilipu:
> I think it's kinda a shame that the z80 folder has asm in it that's specifically not for the z80.

Agree.

Ideal case it should go in the `l/sccz80/8080` directory.

But all the code there is included by z80 build, so that won't work easily (unless you exclude the contents of the file for z80 builds). It would take another major directory rework to change this though. 🤷

Perhaps exclusion from z80 within the `l/sccz80/8080` directory is better? Up to you.

pjshumphreys:
the _8080 postfix what just what I originally used when I was solely focussed on making the functions work for the 8080. The gbz80 and 8085 support came about as the clib compilation failed for gbz80. I figured there was not reason to not make it work for gbz80 as well while I'm adding this code.

Personally I prefer the conditionals where the code is called from. I'm concerned the fake ix/iy code might get mixed up with the real ix/iy code at some point in the future. I think it's good to put a figurative 'knot in the handkerchief' to inform future maintainers to not mix things up.

If you insist, I can rename them of course.

pjshumphreys:
> Perhaps exclusion from z80 within the `l/sccz80/8080` directory is better? Up to you.

meh. Just putting it in the z80 folder has kept the changes small. It's not untrue in the sense they will work on a z80. The conditionals keep the situation clear enough I think.

feilipu:
Feeling like a pedant, but there's no other code in the `l/` directories with a machine suffix. The directory location is sufficient to identify which machine it is meant for.

Same with function names, the assumption is that a machine correct function will be linked, and the function name and call (api) will be the same wherever possible.

Again just for reference.

pjshumphreys:
I welcome your being pedantic. It ensures better quality of the project as a whole for everyone. :)

When I tried to not make an l_setix function at all and just put "pop ix" in l_bsearch and l_qsort directly, z80asm complained when compiling for 8080, even though it was protected by a conditional.

Also,  l_setix_8080 and l_setiy_8080 functions must be in writeable code locations, hence why they're put in the data segment rather than the code segment. That means they each consume 4 bytes of data space. the z80 versions can be be located in the code segment without issue, so are directly romable (If that's the proper way to consider it?)

feilipu:
> I welcome your being pedantic. It ensures better quality of the project as a whole for everyone. :)

Well, ok then. 😊

> When I tried to not make an l_setix function at all and just put "pop ix" in l_bsearch and l_qsort directly, z80asm complained when compiling for 8080, even though it was protected by a conditional.

Hmm. That's odd.`__INTEL__` and `__GBZ80__` shouldn't leak. Can't explain that.

> Also, l_setix_8080 and l_setiy_8080 functions must be in writeable code locations, hence why they're put in the data segment rather than the code segment. That means they each consume 4 bytes of data space. the z80 versions can be be located in the code segment without issue, so are directly romable (If that's the proper way to consider it?)

That's not an issue. You can declare the different sections inside the machine exclusion. Either data or code respectively as needed.

Ideally (2nd best) there remains just one file containing `l_jpix` code paths relevant for all machines.

Best from a pedantic view (but too much work) would be to stop z80 inheriting all `l/sccz80/8080` functions, and just inherit `l/sccz80/common` and split the machine types out. There must be other cases where z80 isn't using the 8080 code. But that may be quite a bit of work (and no actual gain). `l/sccz80/gbz80` doesn't inherit 8080, so that's a place to look for interest.

pjshumphreys:
> Hmm. That's odd.`__INTEL__` and `__GBZ80__` shouldn't leak. Can't explain that.

Perhaps z80asm is ensuring "ix" and "iy" don't appear as tokens at the lexing stage before parsing? I haven't dug into z80asm's code to check though.

I did try putting different sections in the machine specific conditionals. but whichever section name came first in the file seemed to be the one used either way from what I recall.

Reworking the locations of whole sets of files seems too much of a risk surface area when working on changes such as the ones here. That's not something myself as an occasional contributor should be doing either I think.

feilipu:
> Best from a pedantic view (but too much work) would be to stop z80 inheriting all l/sccz80/8080 functions, and just inherit l/sccz80/common and split the machine types out. There must be other cases where z80 isn't using the 8080 code.

Actually it is easy to do this. Brain fart.

The z80 machine will pick its `l_jpix` function first. Later ones, including from 8080, will be ignored. See `l_mult` as an example of this.

feilipu:
>> Hmm. That's odd.__INTEL__ and __GBZ80__ shouldn't leak. Can't explain that.

> Perhaps z80asm is ensuring "ix" and "iy" don't appear as tokens at the lexing stage before parsing?

Don't know. Will have a play on the weekend.

> I did try putting different sections in the machine specific conditionals. but whichever section name came first in the file seemed to be the one used either way from what I recall.

That seems like buggy behaviour. Should be flagged it if so. The `IF * ELSE * ENDIF` should win out, I think.

pjshumphreys:
> That seems like buggy behaviour. Should be flagged it if so. The `IF * ELSE * ENDIF` should win out, I think.

Section handling was retro-fitted onto z80asm right?

feilipu:
>> That seems like buggy behaviour. Should be flagged it if so. The IF * ELSE * ENDIF should win out, I think.

> Section handling was retro-fitted onto z80asm right?

🤷 One for @pauloscustodio

pauloscustodio:
The following code assembles for all CPUs, so there is no lexing issue reading ```ix``` in a CPU that does not support it. Could you provide a failing example for analysis?

```
IF __CPU_ZILOG__||__CPU_RABBIT__
	section code
	ld a, (ix+3)
ELIF __CPU_GBZ80__
	section code
	ld hl, ix_
	ld c, (hl)
	inc hl
	ld b, (hl)
	ld hl, 3
	add hl, bc
	ld a, (hl)

	section data
ix_:defw 0
ELSE
	section code
	ld hl, (ix_)
	ld bc, 3
	add hl, bc
	ld a, (hl)

	section data
ix_:defw 0
ENDIF
```

PR #2106, opened 2022-10-20
enable test for l_bsearch and l_qsort to now run for 8080

pjshumphreys:


suborb:
Great, thank you very much - I've updated the wiki page to remove the line indicating that these two were not supported on 8080

PR #2108, opened 2022-10-21
classic - distribute z80 l_ functions between sccz80/common and sccz80/z80

feilipu:
Classic sccz80 `l_` files were split partially in `l/z80/` and partially organised according to machine type in `l/sccz80/`.

This PR has split all of the common files out from the z80 directory and put them in `l/sccz80/9-common`.
The remaining z80 specific files have been moved to `l/sccz80/5-z80`.

__NOTE__ The PR direction is updated to create a `l/util` alternative with similar inheritance structure as `l/sccz80`. Of course much of the new structure is empty currently, but there is now a place to build machine optimised library utility functions.

Based off this reorganisation it is possible to incrementally write or rewrite "common" opcode versions of `l_` functions without disturbing the z80 version, and have the z80, 8085, or gbz80, etc, versions override the common version where a better implementation is available.

I hope this represents the direction planned for integration of the functions. :pray:


feilipu:
Do not understand why this works,...
```asm
EXTERN l_jpiy_addr
l_setiy:
    push af
    ld a,l
    ld (l_jpiy_addr),a
    ld a,h
    ld (l_jpiy_addr+1),a
    pop af
    ret
```
but this doesn't.
```asm
EXTERN l_jpiy_addr
l_setiy:
    ld (l_jpiy_addr),hl
    ret
```
Perplexing.

feilipu:
> but how does this play with the sdcc compiler rather than sccz80?

There is no problem as sdcc doesn't do INTEL.

The other machines provide sdcc specific functions. The sccz80 `l_` functions aren't used.

feilipu:
It is worth noting that `z88dk-z80asm` has implemented many useful synthetic translations from mnemonic to machine specific opcodes.

So `ex de,hl` and `jr nn` is available everywhere, as is `ld (de+),a` or `ld hl,de` for example. If a synthetic instruction can be done with no side effects, then it is usually available. And this makes writing "common" code much cleaner.

pjshumphreys:
> `jr nn` are available everywhere

I wasn't aware of that. In the code I changed previously for l_bsearch and l_qsort I changed the jr instructions to jp as I believed they wouldn't work on the 8080.

feilipu:
>> jr nn are available everywhere


> I wasn't aware of that. In the code I changed previously for l_bsearch and l_qsort I changed the jr instructions to jp.

It is very helpful to write clean assembly. Avoids lots of conditional assembly.

So the author's question becomes do I want a `jp` here because it is faster? Otherwise I can use `jr` without concern and let z80asm decide. 

But it is easy to forget there's no magic and no common synthetic `jr M,nn` instruction, for example, because the gbz80 can't do it. So it does limit choice in some ways. So, if the function performance is important, write a version for each machine, like eg. `l_mult`.

pjshumphreys:
> It is very helpful to write clean assembly. Avoids lots of conditional assembly.

It's irrelevant to this PR and just my own opinion of course, but I consider this a misfeature and potential future tripwire that something as low level as an assembler should generate something other than what the programmer wrote. If the standard library is built on the assumption z80asm will be doing this then that's just the way it is, but I'd want to leave this turned off for asm I write for my own projects. I might be counting bytes of code output or t-states for instance. If it were my decision (which of course it's not :) ) I'd rather this kind of difference be catered for in the library source (using macros perhaps?) rather than internally within z80asm. Rant over :)

feilipu:
> An assembler should generate something other than what the programmer wrote.

The good news is that z80asm will _always_ do what you ask. It only provides synthetics if you're asking the "wrong" thing for the machine type you told it to assemble for.

> I'd want to leave this turned off for asm I write for my own projects. I might be counting bytes of code output or t-states for instance.

It is only "on" if you give it "wrong" mnemonics. So it's totally ok for generating exactly what you intend.

Synthetics came before z80asm could do macro assembly. And since now it does support macros, you can choose your own poison. 😊

pjshumphreys:
Sorry to keep posting off topic here. Shall we continue discussion about synthetics on a new forum thread instead? I'd kinda like it if z80asm could just throw an error if I ask it to generate invalid instructions for a target cpu.

pjshumphreys:
I added a thread on the wiki to continue the synthetic opcodes discussion there. https://z88dk.org/forum/viewtopic.php?t=11697
Sorry for going off topic :S

suborb:
@feilipu aren't some of these routines called from sdcc routines/generic library routines?

feilipu:
I can check again whether any directory are broken across the libraries.

But there's no change in functionality, afaik. 

Its just housekeeping to simplify everything onto the inheritance structure. Anything that was in a library should still be there in the same library as it was.

suborb:
The inheritance is good, but having l_eat_ws (used by atoi/strtou etc) insides an sccz80 directory seems anomalous.

feilipu:
Hmm. My false assumption that all `l_` functions were related to sccz80.

But on the other hand. Having one directory tree for all the code is nice and clean too.

I didn't try to rename/repurpose directories, so if you'd like to suggest alternatives to make it better?

The goal (besides simplifying the structure) was to enable the machine specific implementation more broadly. Like eg. `l_mult` or (now added) `l_setiy`.

suborb:
It's taken me 15 minutes to remember why I numbered the directories (note to future me: globbing)

Given that we've got l/sccz80 l/sdcc then perhaps l/util with a similar structure to l/sccz80 might work?

However, looking through the other directories, I can see that I've just sprinkled `IF __CPU_` fairly liberally so maybe it's overkill for things other than the tightly bound compiler support routines.

feilipu:
When I was pushing the files around, it was without much granularity.
This is a first pass. I didn't want to break anything.

When the file was listed in the 8080 lists, then it went into 9-common.
Though there were actually a few files that shouldn't have been there.

Everything else went into 5-z80.

---
As an outcome, it should be possible to write versions of any compiler function optimised for the machine capability. But, some of the library z80 specific `l_` functions are unlikely to be ported back to 8080 or gbz80, as they're not really needed (or they would have been done already).

I'm happy to keep going with this PR with input on the direction that's preferred.
Or to merge this one and continue with a new PR as needed?

As you say, a `l/util` folder would still need some kind of inheritance structure, as otherwise it becomes difficult to see which files should run on 8080 / gbz80, and which are z80 only.

feilipu:
As a thought, perhaps if the function appears in this [header](https://github.com/z88dk/z88dk/blob/master/lib/z80_crt0.hdr) it is considered compiler support. Otherwise it is library support and goes to `l/util` with the same inheritance structure.

How does that sound?

suborb:
I think that's a good divide - the repeated block instructions aside.

feilipu:
Ok that's done now. Ready to go.

I actually did the divide based on the `SECTION code_l_sccz80` staying in `l/sccz80/` vs. `SECTION code_l` going to `l/util/`, as that fitted much better.

Issue #2099, opened 2022-10-13
lib3d - add 4d and matrices


feilipu:
Looking at the path to integrate the __`lib3d`__ with pieces of a [homogenous coordinate](https://en.wikipedia.org/wiki/Homogeneous_coordinates) [__`3d`__](https://github.com/feilipu/z88dk-libraries/blob/master/3d/source/include/sccz80/3d.h) library.

Quoting from Wikipedia on why to do this: "Homogeneous coordinates are ubiquitous in computer graphics because they allow common vector operations such as [translation](https://en.wikipedia.org/wiki/Translation_(geometry)), [rotation](https://en.wikipedia.org/wiki/Rotation_(mathematics)), [scaling](https://en.wikipedia.org/wiki/Scaling_(geometry)) and [perspective projection](https://en.wikipedia.org/wiki/Perspective_projection) to be represented as a matrix by which the vector is multiplied. By the chain rule, any sequence of such operations can be multiplied out into a single matrix, allowing simple and efficient processing. By contrast, using Cartesian coordinates, translations and perspective projection cannot be expressed as matrix multiplications, though other operations can."

The key things needed are.

1. vectors with an added 4th  ___w___ dimension.
2. matrix operations for 4x4 order. (I updated `3d` library yesterday to use an intermediate matrix, so that the [matrix operations can be chained without costing RAM](https://github.com/feilipu/z88dk-libraries/blob/master/3d/demo/demo_3d.c#L193). e.g. see [`rotx_m()`](https://github.com/feilipu/z88dk-libraries/blob/master/3d/source/rotx_m.c#L63) ).
3. options to choose the existing fixed point functions (multiply, sin, and cosine) or any of the sccz80 supported [floating point library](https://github.com/feilipu/z88dk-libraries/blob/master/3d/source/include/sccz80/3d.h#L58) functions.

@zx70 My `3d` library is not very comprehensive. Really just the matrix operations and the x,y,z,w vector operations. Perhaps I can just add the needed functions and add the dimension to the vector structures of `lib3d`? But, I don't want to mess with other machines' outcomes that I can't test.

- [x] 4d code added
- [x] integrated with existing fixed point build #2111
- [ ] makefiles for floating point functions

zx70:
Feel free to extend, I think the possible impacts are limited, if you are evolving the real 3D stuff.
[...]

PR #2111, opened 2022-10-23
classic - lib3d add homogeneous coordinate matrix operations

feilipu:
Add homogeneous coordinate matrix operations.
For #2099

Still need to update the Makefile to build the library with floating maths point at some stage.

The new vector and matrix operations won't work properly with fixed point maths, but this won't affect the previous functions' performance.

### Vector 4d functions

- unit_v
- dot_v
- mult_v
- scale_v

### Matrix 4d functions

- identity_m
- mult_m
- rotx_m
- roty_m
- rotz_m
- scale_m
- shear_m
- translate_m
- projection_opengl_m
- projection_w3woody_m

### DIFF
--- a/examples/msx/ex1.c
+++ b/examples/msx/ex1.c
@@ -40,13 +40,13 @@ void star_randomize(star_t* st) {
 void star_move(star_t* st) {
 	int x, y, z;
 	z = (st->z -= 2);
-	x = f2i(divfx(st->x, z)) + 128;
-	y = f2i(divfx(st->y, z)) + 96;
+	x = st->x / z + 128;
+	y = st->y / z + 96;
 	if (z < 1 || x < 0 || x >= MODE2_WIDTH || y < 0 || y >= MODE2_HEIGHT) {
 		star_randomize(st);
 		z = st->z;
-		x = f2i(divfx(st->x, z)) + 128;
-		y = f2i(divfx(st->y, z)) + 96;
+		x = st->x / z + 128;
+		y = st->y / z + 96;
 	}
 	st->last = st->addr;
 	st->addr = map_pixel(x, y);
--- a/examples/msx/ex9.c
+++ b/examples/msx/ex9.c
@@ -43,13 +43,13 @@ void star_randomize(star_t* st) {
 void star_move(star_t* st) {
 	int x, y, z;
 	z = (st->z -= 2);
-	x = f2i(divfx(st->x, z)) + 128;
-	y = f2i(divfx(st->y, z)) + 96;
+	x = st->x / z + 128;
+	y = st->y / z + 96;
 	if (z < 1 || x < 0 || x >= MODE2_WIDTH || y < 0 || y >= MODE2_HEIGHT) {
 		star_randomize(st);
 		z = st->z;
-		x = f2i(divfx(st->x, z)) + 128;
-		y = f2i(divfx(st->y, z)) + 96;
+		x = st->x / z + 128;
+		y = st->y / z + 96;
 	}
 	st->sx = x;
 	st->sy = y;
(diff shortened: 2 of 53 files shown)


zx70:
what does the integet->float->integer conversion do?

feilipu:
There is an `i2f()` and `f2i()` that I believe creates a 24.8bit fixed point number (as a long) from a 16bit integer, and the reverse operation. Used in the existing functions, but not in my float based new additions. 

But otherwise I'm not sure what your question relates to?

zx70:
you're totally right,  i2f is not dealing with floats!   sorry it was too long ago

feilipu:
@zx70

But, looking at it again this morning, I am a bit confused about how the fixed point conversions are working.

I see that `f2i()` is actually converting 26.6 (one byte right, two bits left) to `int`.
And `i2f()` does the reverse (one byte left, two bits right with sign extension) `int` to 26.6 fixed point.

So I've changed some of the `#define` because they looked wrong.

 `divfx()` didn't convert back to `int` after the division. Neither did `sqrtfx()`
I've added a `f2i()` to each definition.

But I didn't change the `mulfx()` and `sqrfx()` that look to me like they need to have a left shift of 3 (multiply by 8) to make a correct 26.6 fixed point, which is then converted back to `int`. Or just drop the conversion `f2i()` conversion, as `int` * `int` is still an `int`.

Perhaps I'm not getting their operations correct?
And getting confused about what they should be doing?

```asm
; before
#define mulfx(x,y)	f2i((long)y * (long)x)
#define divfx(x,y)	(i2f(x)/(long)y)
#define sqrfx(x)	(f2i((long)x * (long)x))
#define sqrtfx(x)	(((long)(sqrt(x)))*8)
```

```asm
; now
#define mulfx(x,y)	(f2i((long)(x * y)))
#define divfx(x,y)	(f2i(i2f(x)/y))
#define sqrfx(x)	(f2i((long)(x * x)))
#define sqrtfx(x)	(f2i(((long)(sqrt(x)))*8))
```

zx70:
IIRC this trick is from the Rafael's MSX library,  
For what I could understand it was a way to do a little bit of rounding in the scaling operations and reduce the precision loss due to the integer values.

feilipu:
Hmm. So they seem to only be used within the `lib3d` library. It is possible that they are not totally compliant with their own specification, and what I see as "errors" cancel out when they are used as a pair.

I'll look a bit harder to understand.

zx70:
Yes, for what i could understand it's simply a quick way to multiply the operators by two (thus, sometimes it is sufficient to do it only on one of the parameters).    Later on, a division by two will happen.

feilipu:
So `divfx()` leaves its result as a 26.6 fixed, and in `examples/msx/ex1.c` and `examples/msx/ex9.c` the final `f2i()` conversion is done in the calling function. Which explains why it is missing from the `#define`. Don't know why that was the chosen path.
There's no reason to leave it "wrong", so I'll fix it.

The `mulfx()` should have one conversion to 26.6, so that the binary point is in the correct place. Currently it doesn't have that, so the conversion back to integer moves the binary point 6 places for no reason.<br>Corrected `#define mulfx(x,y)	(f2i(i2f(x)*y))`

Similarly `sqrfx()` has the binary point in the wrong place.<br>Corrected `(f2i(i2f(x)*x))`

But actually the functions that use `mulfx()` or `sqrfx()` would be better just using integer multiplies. There's no accuracy lost in a multiply, and it saves the conversion overhead and reduces the multiply to integer rather than long. Basically, I've put them back to use the commented out code.

And, it is the same for `divfx()` as there's no benefit from using `long` fixed point because the intermediate values are all handled as integers. So I returned returned to the original commented out code. It will be 4x faster doing normal integer divides rather than long fixed point divides.

zx70:
Thank you, it looks better.    At the time I focused on the surface drawing (the 'stencil' stuff) and dumbly converted the original i2f/f2i macros into functions but didn't focus on the real effectiveness of the long<>int conversion job, I just kept them in working order.
There could be a generic variant of those demos, but I woudn't care in this stage.
As you can see they're few examples not very linear and IMO nobody tried to use the library in depth in the past years.
Thus a bit of housekeeping is for the best  :)

feilipu:
I'll merge this and follow up with a make recipe for the floating point functions later.

PR #2129, opened 2022-11-16
Galaxy Gemini

zx70:
New target  (yet another CP/M variant).
Support for low resolution graphics, floppy disk format and various terminal modes.


desertkun:
Just curious, why one commit per file? :D

zx70:
not telling  :P

PR #2140, opened 2022-12-06
Add inp/outp for the 8080

trondd555:
Uses the same technique as AZTEC C which is to dynamically write the subroutine into reserved data memory and call/jump to it.  This should allow the code to remain ROMable.

While here, change outp_callee to use hl instead of af for stack juggling. af was sending the program off into lala land.

### DIFF
--- a/libsrc/stdlib/outp.asm
+++ b/libsrc/stdlib/outp.asm
@@ -8,11 +8,12 @@ EXTERN asm_outp
 .outp
 ._outp
 
-   pop af
+   pop bc
    pop de
    pop bc
    push bc
    push de
-   push af
+   push bc
    
    jp asm_outp
+
--- a/libsrc/stdlib/outp.asm
+++ b/libsrc/stdlib/outp.asm
@@ -8,12 +8,12 @@ EXTERN asm_outp
 .outp
 ._outp
 
-   pop bc
+   pop hl
    pop de
    pop bc
    push bc
    push de
-   push bc
+   push hl
    
    jp asm_outp
 
--- a/libsrc/Make.config
+++ b/libsrc/Make.config
@@ -16,9 +16,9 @@ OUTPUT_DIRECTORY :=  $(dir $(abspath $(lastword $(MAKEFILE_LIST))))
 NEWLIB_DIRECTORY := $(subst //,/,$(OUTPUT_DIRECTORY)/_DEVELOPMENT/)
 
 LIBLINKER = z88dk-z80asm -d -I$(ZCCCFG)/../
-ASSEMBLER = z88dk-z80asm 
+ASSEMBLER = z88dk-z80asm
 CFLAGS = -O2 -vn -c
-AFLAGS = 
+AFLAGS =
 
 Q ?= @
 ZCC := $(Q)zcc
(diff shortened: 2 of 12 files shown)


feilipu:
Maybe fix `wpoke_callee.asm` stack handling while you're at it? Uses `af` too.

trondd555:
Sure.  I'll test and see if it's an issue.  Now that you mentioned it, I see outp.asm does the same thing but I only saw the problem when using af in outp_callee.asm

trondd555:
Oh, outp and wpoke don't get added to the code.  Only *_callee gets added.  That's why I didn't see the issue in there.   I am not sure how these callee functions get chosen so not sure how I could test it.

feilipu:
> Oh, outp and wpoke don't get added to the code. I am not sure how these callee functions get chosen so not sure how I could test it.

The will be called by call by function reference usually, so you could test that way.

But in any case using `af` to hold the return address for an 8080/8085 is an "oops", and it doesn't really need prior testing to warrant fixing it.

trondd555:
Ok, so do you want it fixed just for 8080/8085 or just change it for everything?

feilipu:
> Ok, so do you want it fixed just for 8080/8085 or just change it for everything?

I think there's no problem using `bc` to hold the return addresses for all these `outp.asm`, `wpoke.asm`, and `bpoke.asm` as the general case. So make it so.

We'll get a second opinion before committing.

libsrc/stdlib/outp.asm
```diff
@@ -8,11 +8,12 @@ EXTERN asm_outp
 .outp
 ._outp
 
-   pop af
+   pop bc
```
feilipu, line 11: Need to use `hl` here.
`bc` is used to hold the port.

trondd555: Oops.  Fixed.  Thanks.

feilipu:
Just thinking, it would be better to use `__CPU_INTEL__` rather than `__CPU_8080__`. The i8085 builds usually include 8080 too, but this would make it clearer.

Sorry to make this late suggestion.

suborb:
Agree with the change to `__CPU_INTEL__`

Couple of other minor things:

- Normally we use Zilog opcodes rather than Intel
- `iotmp` could be shared and 3 bytes would be saved

And as a note `__sfr` access does work on the 808x:

```
__sfr __at 0x20 myport;

int main()
{
        myport = 10;
}
```

trondd555:
Pushed last commit before I saw @suborb's comment.

Working on fixing opcodes.

I don't know where to put iotmp such that it's defined/reserved whenever either function is included.

feilipu:
Use a new `iotmp.asm` file and make the space public. Then reference as `EXTERN`.

suborb:
> I don't know where to put iotmp such that it's defined/reserved whenever either function is included.

The easiest way to do it is with a little bit of a refactor:

- Put the 808x implementations in stdlib/8080
- Add an extra file that contains the "iotmp" variable
- Undo the makefile changes

Sorry for making you go round in circles...

feilipu:
> So is the consensuses to move the code to stdlib/8080?

Yes.

Also then the extra `iotmp.asm` file will go there too, and it won't need `IF __CPU_INTEL__` as its use will be obvious from where it is located.

feilipu:
> Moving the code into stdlib/8080 makes gbz80 pick it up and it fails to find the `EXTERN _iotmp`.

just a minor point, but using a leading underscore allows the variable to be found from C as `iotmp`, and since it is 3 bytes it would not fit into any C type and may conflict with someone's variable. Best not to use a leading underscore in the name, or use two of them to hide it from C.

> Should gbz80 filter out these files from stdlib/8080 now or is there some way to make it work that I don't know?

I'd just wrap the functions and reserved space in `IF __CPU_INTEL__` for the time being, and let @suborb change it later if he likes.

trondd555:
Thanks.

Whoops, I see I accidentally reverted some changes. :(  Thanks for catching those.

PR #2141, opened 2022-12-10
switch from FNKBDS to FNKBDZ

Hojoe42:
Currently gets() returns many chars for one keypress. 

The function FNKBDZ acknowledges the the key and prevent duplicate chars.


suborb:
Thank you!

Issue #2150, opened 2023-01-06
sccz80: Expecting constant expression on struct init = { with a cast }


desertkun:
main.c
```C
struct haha_base_t {
    struct haha_base_t *next;
};
struct haha_derived_t {
    struct haha_base_t base;
};
static struct haha_derived_t a = {{0}};
static struct haha_derived_t b = {{(struct haha_base_t*)&a}}; // <--- doesn't like this
int main()
{
    b.base.next = 0;
}
```
Gives
```
main:8:59: error: Expecting constant expression
```
While
```C
struct haha_base_t {
    struct haha_base_t *next;
};
static struct haha_base_t a = {0};
static struct haha_base_t b = {&a};
int main()
{
    b.next = 0;
}
```
does not. Seems like casting a "const expression" `&a` to `(struct haha_base_t*)&a` leads for the expression to lose its constness. Same happens with `&(a.base)` but I am willing to make do with a cast.

The simplest way to reproduce is:
```C
int a = 10;
int *b = (int*)&a;
```

Going to spend weekends debugging how this can be fixed, but also looking for hints.

Seems like `expression` sets con to 0 for cast expressions. Same goes for referencing of a child object of a const expression, but like I said, willing to settle for just a cast. Thanks for any help.

desertkun:
I actually figured a way to make it work with &base.xx kind of notation. See #2151

PR #2151, opened 2023-01-06
sccz80: support to init globals with submembers of other globals

desertkun:
Solves #2150, see test for use case


suborb:
I think there’s similar code when handling sizeof - might worth a check to see if there’s missing cases either way.

desertkun:
Yeah sizeof is focused on size only, won't "port" so to speak.

PR #2152, opened 2023-01-09
Fix GDB TCP connection on Windows 

empathicqubit:
On Windows we need to initialize WinSock and use the correct send/recv functions instead of read/write

### DIFF
--- a/src/ticks/debugger.c
+++ b/src/ticks/debugger.c
@@ -13,6 +13,10 @@
 #include <stdarg.h>
 #endif
 
+#ifdef _MSC_VER
+#include <WinSock2.h>
+#endif
+
 #include "debugger.h"
 #include "utlist.h"
 
@@ -220,6 +224,14 @@ void ctrl_c_handler(int signum) {
 }
 
 void debugger_init() {
+#ifdef _WIN32
+			struct WSAData wsa;
+			int ret = WSAStartup(MAKEWORD(2, 2), &wsa);
+			if (ret) {
+				bk.debug("Socket initialization error: %d\n", ret);
+			}
+#endif
+
     signal(SIGINT, ctrl_c_handler);
     linenoiseSetCompletionCallback(completion, NULL);
     linenoiseHistoryLoad(HISTORY_FILE); /* Load the history at startup */
--- a/src/ticks/debugger.c
+++ b/src/ticks/debugger.c
@@ -5,7 +5,7 @@
 #include <ctype.h>
 #include <signal.h>
 #ifdef WIN32
-#include <WinSock2.h>
+#include <winsock2.h>
 #include <io.h>
 #define F_OK 0
 #define access _access
--- a/src/ticks/debugger.c
+++ b/src/ticks/debugger.c
@@ -5,16 +5,13 @@
 #include <ctype.h>
 #include <signal.h>
 #ifdef WIN32
+#include <WinSock2.h>
 #include <io.h>
 #define F_OK 0
 #define access _access
 #else
 #include <unistd.h>                         // For declarations of isatty()
 #include <stdarg.h>
-#endif
-
-#ifdef _MSC_VER
-#include <WinSock2.h>
 #endif
 
 #include "debugger.h"
@@ -1101,7 +1098,7 @@ static int parse_number(char *str, char **end)
     if ( *str == '$' ) {
         base = 16;
         str++;
-    } 
+    }
     return strtol(str, end, base);
 }
 
@@ -1148,7 +1145,7 @@ static int cmd_stack(int argc, char **argv)
 }
 
 
-/* Parse an address operand. 
+/* Parse an address operand.
  *
  * It may be one of:
  * 1. A number address
@@ -1181,7 +1178,7 @@ const char *resolve_to_label(int addr)
 {
     static char tbuf[1024];
     const char *sym;
-    
+
     if ( (sym = find_symbol(addr, SYM_ADDRESS) ) != NULL ) {
         return sym;
     }
@@ -1248,7 +1245,7 @@ static int cmd_disassemble(int argc, char **argv)
 
 
 
-static int cmd_registers(int argc, char **argv) 
+static int cmd_registers(int argc, char **argv)
 {
     bk.invalidate();
     const unsigned short pc = bk.pc();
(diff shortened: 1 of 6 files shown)


desertkun:
Thank you for your interest in the project.

1. Let's replace read/write to recv/send for all systems alike. 
2. Maybe we can do a little `bk.debug("Socket error: %d\n", WSAGetLastError());` for windows too.
3. Please move `WSAStartup` into `void debugger_init()` since gdb can run in mi/2 mode (a.k.a. in IDEs), and should probably work for windows too. And you can ditch the `WSACleanup`.

Also, what emulator are you using?

empathicqubit:
The emulator I'm using is Wabbit. I added GDB support to it, but it's not merged yet. I noticed in Wireshark it wasn't connecting at all, but it was working before when I was using z88dk-gdb from my Linux VM.

desertkun:
Let me know when it has the support, so I can list it on [speccytools.org](https://speccytools.org/). Also you can refer to [gdbserver](https://github.com/speccytools/fuse-for-macosx/blob/7a2c1e7a69150172bcb7c5967002818a518ee051/debugger/gdbserver.c) support for FuseX fork I made, it is compatible with z88dk-gdb. Also you can take a look at [windows variant](https://github.com/speccytools/fuse/blob/gdbserver/debugger/gdbserver.c) but it lags behind. Any issues you find, I would appreciate PRs into windows fork, too.

empathicqubit:
I'm not sure how to fix the msys build. I guess I need to edit a Makefile but idk

suborb:
debugger_gdb_packets.h only includes Winsock when compiled with Visual Studio: https://github.com/z88dk/z88dk/blob/master/src/ticks/debugger_gdb_packets.h#L8

src/ticks/debugger.c
```diff
@@ -13,6 +13,10 @@
 #include <stdarg.h>
 #endif
 
+#ifdef _MSC_VER
```
empathicqubit, line 16: @suborb

empathicqubit: oh wait. I am dumb

empathicqubit: @suborb I'm actually a little confused. Do I need winsock for mingw or should I change the debugger_init def to be msc only also?

suborb: Winsock is needed for all toolchains on windows.

suborb: I think that means this line: https://github.com/z88dk/z88dk/blob/master/src/ticks/Makefile#L27 needs the now badly named variable `$(GDBLDFLAGS)` adding to the end of it - see the rule for z88dk-gdb

empathicqubit:
Can you help me understand what the i command is doing? It doesn't seem to match the GDB spec. It looks like it is giving the emulator a certain byte offset to run to, but the spec says it should cycle step.

desertkun:
Gdb protocol wise? Yeah I think step over a few bytes

empathicqubit:
@desertkun okay so z88dk does not match the spec? That's fine I just wanted to clarify

desertkun:
It seems you are correct. It should be `i,%d` instead. [Mame](https://github.com/mamedev/mame/blob/master/src/osd/modules/debugger/debuggdbstub.cpp#L1350) does not even support `i`. I think we should just remove that instruction and instead place a temporary breakpoint at specific place since this is what [FuseX gdbserver](https://github.com/speccytools/fuse-for-macosx/blob/master/debugger/gdbserver.c#L397) does. Neither of them do not follow spec, including [spectranet-gdbserver](https://github.com/speccytools/spectranet-gdbserver/blob/master/src/server.c#L134)

What complicates things, it's hard to detect a specific version of [fuse](https://github.com/speccytools/fuse-for-macosx/blob/master/debugger/arch.h#L7) as I wrote dull manifest (and so does [mame](https://github.com/mamedev/mame/blob/master/src/osd/modules/debugger/debuggdbstub.cpp#L695)).

I we can simply bump that version, to `1.1`, and send `i,%` for 1.1 or version missing and `i%` for '1.0'.

But that would require an orchestrating of five things at once. So I'd just enjoy the ride and follow suit 😄

empathicqubit:
@desertkun It's probably best to leave it the way it is, but I think what it actually should be doing is advancing the CPU clock n cycles. So that means that you could get stuck in a state "between" an instruction being executed. It sounds really weird to me though.

desertkun:
the xml manifest (`qXfer:features:read:target.xml:0,3fff`) is the feature of "what is was". Only problem is I did not take advantage from it and copy-pasted it 😄 Had too many things on my plate. I've been spamming the owner of [Es.pectrum](https://www.habisoft.com/espectrum/EN.htm) Emulator for gdbserver support but he has no interest in that and doesn't want to open sources either.

I do feel happy that you are following thru with your emulator. Maybe after you implement it, I can convince the author again ("see, you're being outcompeted 😄" )

empathicqubit:
Can you tell me exactly what I need to do to fix this?

https://github.com/z88dk/z88dk/actions/runs/3878501600/jobs/6614801211

desertkun:
I saw network-related issue, reran for the sake of it

empathicqubit:
> Does z88dk-gdb have any command to set memory other than the file one?

suborb:
> Can you tell me exactly what I need to do to fix this?
> 
> https://github.com/z88dk/z88dk/actions/runs/3878501600/jobs/6614801211

It's "winsock2.h" not "WinSock2.h"

empathicqubit:
Okay. The path I have is C:\Program Files (x86)\Windows Kits\10\Include\10.0.22621.0\um\WinSock2.h

suborb:
> Okay. The path I have is C:\Program Files (x86)\Windows Kits\10\Include\10.0.22621.0\um\WinSock2.h

Gotta love file system case insensitivity..

* mingw: https://github.com/mirror/mingw-w64/blob/master/mingw-w64-headers/include/winsock2.h
* Microsoft docs (example): https://learn.microsoft.com/en-us/windows/win32/api/winsock2/nf-winsock2-wsasetlasterror

empathicqubit:
Yeah. It's funny how it's listed two different ways on the Microsoft page, neither of which is the same as the SDK

desertkun:
Thank you for the effort

empathicqubit:
"No probalo!"

PR #2153, opened 2023-01-09
Fix drive letter colons in GDB line info by parsing them differently

empathicqubit:
On Windows, the colon in an absolute path breaks the line parsing. This is an attempt to fix it. It seems to work correctly, however I did not see any "classic" patterns, so I can't say if they work for sure still, but they should.

### DIFF
--- a/src/ticks/syms.c
+++ b/src/ticks/syms.c
@@ -42,16 +42,35 @@ static int demangle_filename(const char *input, char *filename, char *funcname, 
     *funcname = 0;
 
     // Full packed format
+#ifdef WIN32
+	// Match path with drive letter in it
+    if ( sscanf(input,"%c:%[^:]::%[^:]::%d::%d:%d",filename,&filename[2], funcname, level, scope, lineno) == 6) {
+		filename[1] = ':';
+        return 0;
+    }
+#endif
     if ( sscanf(input,"%[^:]::%[^:]::%d::%d:%d",filename,funcname, level, scope, lineno) == 5 ) {
         return 0;
     }
 
     // Classic in function format: adv_a.c::CHKAWAY:2206
+#ifdef WIN32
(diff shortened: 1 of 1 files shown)


desertkun:
We should probably mangle the colon in the first place when it is generated and replace it with `_3A` (see encode_cdbstring) like similar to `__CDBINFO__S_3aG_24variable_24_30_5f_30_24_30_28_7b_32_7dSI_3aS_29_2cE_2c_30_2c_30` kind of strings in .map files, and then dehex it back. See debug.c:654

empathicqubit:
As far as I can tell this function parses the details after the semicolon, not the variable name before the = . It seemed a bit weird tbh, but it works now

desertkun:
Ok can you instead add an #ifdef'd extra check to `sscanf` that would simply have an explicit match to C: kind of string for filenames?
e.g. `"%c:%[^:]::%[^:]::%d::%d:%d", drivename, filename` and then concatenate them back?

empathicqubit:
I'm wondering if there's another case for colons in Windows paths that I'm forgetting that won't work with a single character.

suborb:
> We should probably mangle the colon in the first place when it is generated and replace it with `_3A` (see encode_cdbstring) like similar to 
> ....
> But then again it might interfere if you decide to hava a file literally called _3A. So we'd have to mangle underscores too.

Oh, I do love double encoding. I think the filename needs to be encoded before being put into the cdbstring. The whole cdbstring will then be encoded (as it currently is) so as long as we decode in ticks it should work out ok even if the filename is _3A

empathicqubit:
I think it's ready to merge now

desertkun:
Impressive trickery 😄 . Did you test it locally works file? I'll merge after it passes

empathicqubit:
I did test it locally with a file with both windows and linux paths in it and it hit all the breakpoints except classic

Issue #2154, opened 2023-01-09
"Full" Windows path with backslashes confuses #include ""


empathicqubit:
```
/c/z88dk/bin/zcc +ti8x -Wall -debug -g -O0 -S -o "build/src/main.i.asm" "Z:\UserProfiles\EmpathicQubit\tyzzy\src\main.c"

Z:\UserProfiles\EmpathicQubit\tyzzy\src\main.c: line 1: file '../3rdparty/z88dk-gdbstub/src/gdb.h' not found
Z:\UserProfiles\EmpathicQubit\tyzzy\src\main.c: line 2: file '../lib/debug-impl.h' not found
make: *** [Makefile:189: build/src/main.i.asm] Error 1
```

But this works:

```
/c/z88dk/bin/zcc +ti8x -Wall -debug -g -O0 -S -o "build/src/main.i.asm" "Z:\UserProfiles\EmpathicQubit\tyzzy\src/main.c"
```

Note the forward slash in the last path part

suborb:
Just from code inspection I spot:

https://github.com/z88dk/z88dk/blob/master/src/ucpp/cpp.c#L679
https://github.com/z88dk/z88dk/blob/master/src/ucpp/cpp.c#L684

I think both the MSDOS and the non-AMIGA case need to be activated.

empathicqubit:
so `if defined(MSDOS) || defined(WIN32)` ?

empathicqubit:
No, actually there would need to be a third option for Windows since both / and \ can work

PR #2155, opened 2023-01-10
MI2 backslashes

empathicqubit:
The MI2 interpreter doesn't escape backslashes in the response which, you guessed it, causes problems with Windows. Now all filenames are escaped. Also fixed another colon scan problem in the symbols.

This PR is based on #2153 because I needed the fixes in there to debug further.

### DIFF
--- a/src/ticks/debugger_mi2.c
+++ b/src/ticks/debugger_mi2.c
@@ -107,6 +107,35 @@ static command mi2_commands[] = {
     { NULL, NULL }
 };
 
+static char* escape(char* input) {
+	int len = strlen(input);
+	int count = 0;
+	char* ptr = input;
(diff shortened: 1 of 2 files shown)


desertkun:
How do I test it from UI perspective? Place & hit a breakpoint? What's the IDE? Ideally I'd appreciate a repo with just the test case.

empathicqubit:
For me I am using the Native Debug extension in vscode

empathicqubit:
It's difficult for me to provide an example at the moment due to #2154

desertkun:
@empathicqubit I have found time to make a round on debugger, does this issue still stand? Could you please provide an example project I can veritfy this with?

PR #2157, opened 2023-01-11
Add concurrency check so outdated jobs get killed

empathicqubit:


empathicqubit:
You can see it working in the logs here: https://github.com/empathicqubit/vscode-cc65-debugger/actions/workflows/check.yml

.github/workflows/build-mingw-on-ubuntu.yml
```diff
@@ -5,6 +5,10 @@ on:
   pull_request:
     branches: [ master ]
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
```
suborb, line 9: This is a great feature, but I think this cancelling would only work on  PRs, not on branches since run_id will be unique per run?

Does this actually need to be:

```
group: ${{ github.workflow }}-${{ github.head_ref || gitHub.ref }}
```

empathicqubit: I think you're right. I guess it depends on what we want. Also I think this would avoid killing manual runs.

suborb: Regardless, I'll merge and see how it actually behaves!

empathicqubit: Thanks!

PR #2159, opened 2023-01-13
BIFROST*2 interface improvement

einar-saukas:
It was too confusing that certain functions must be called adding value BIFROST2STATIC to tile index but not others. Now BIFROST2drawTileH will ignore the BIFROST2STATIC flag so it won't matter anymore if value BIFROST2STATIC was added or not. For further details see:

https://spectrumcomputing.co.uk/forums/viewtopic.php?p=109323#p109323


suborb:
Thanks @einar-saukas I'll merge as soon as the build plays nicely! (infrastructure issue, nothing to do with this PR)

PR #2162, opened 2023-01-15
Add filesystem API and improve console output (add small text)

empathicqubit:


suborb:
I've no idea why the build is wonky :(

empathicqubit:
It's because there's a symlink to the Windows output folder

empathicqubit:
This works on TI 83 Plus now (the file API. Screen stuff is a little sloppy), but I would like it to work on some others if possible before merging.

What are we doing about tests? I have written 5 or so dumb ones. I'm not sure how that would work considering how platform specific it is.

suborb:
I've not come up with a good way of automatically testing machine specific stuff - and going down the route of taking screenshots of an emulator and comparing it to reference pictures is neither sustainable or fun.

What we usually end up doing is creating an example that sits in examples/[machine] that exercises the functionality - they can be hooked into the build so at least "it compiles" and then manually run when/if things change.

suborb:
> @suborb Is there a good way to figure out what causes the program size to be huge? Like some kind of dependency graph or something? My "simple" test program is close to 8K.

Sorry, I totally missed this. Adding -Cl-v is quite useful to figure out which routines are being pulled in.

zx70:
Appmake is finally able to manage the ti83p binary format decently.

zx70:
I'really like to get the code merged  but I'm lost.
Is it necessary to force the library cleanup on any target ?

<<<<<<< ti-improvements
	$(MAKE) -C target/ticalc clean
	$(MAKE) -C target/ticalc
=======
>>>>>>> master


@empathicqubit, plase confirm the changes you are proposing in the Makefiles are still valid, and/or please refresh the two files to resolve the conflict.

libsrc/Makefile
libsrc/target/ticalc/Makefile

zx70:
@empathicqubit please don't give up. ;)
If it's reasonable for you we can open a thread in the forum of in an issue to try together to fill the gap

empathicqubit:
Okie :) I'm deep in the midst of rewriting a 6502 debugger that I wrote, so it might be a little while before I get back to this

PR #2164, opened 2023-01-20
(nabu) Add support for HCCA and retronet

suborb:
Import in retronet by @DJSures for the Nabu.

We use interrupts for both reading and writing to the HCCA interface and support "zero copy" so that we can read/write directly from memory.

In theory the code works from +cpm as well as the +nabu target, however I can't get mame to boot from floppy if the internet adapter is connected.

Ideally these would be wrapped up with in fcntl.

Now the framework is here, we can also add support for an alternate protocol specced out here: https://github.com/hanshuebner/nabu-figforth/blob/main/nabu-comms.md


zx70:
I noticed that in appmake we now have a definition for "naburn_spec", supposedly an HDD for the NABU.
The compiler warning about overflowing the tracks number (set to 16384) looks like a good advice, IMO.

suborb:
I thought I’d bumped tracks up to an int to deal with this?

The images I had locally were working so it just be a missing commit.

zx70:
No, it is not  <blush>

Issue #543, opened 2017-11-30
can't build campus lisp


zx70:
sorry guys,  I must be missing something but I don't seem to be able to build clisp.c and see it run correctly.    after the prompt it becomes unstable,   it can be a simple low memory problem,  can you reproduce the problem ?
eg with sccz80 any instruction but nil creates an error condition.

suborb:
I'll take a look this weekend. This example seems to be a bit of a canary - have we got an example program that generates output? I can add it to the tests to ensure it runs correctly.

zx70:
thankyou for answering  :)
it is sadly the bit I have so far, I'm taking aside sdcc which may require extra tuning basing on the specific page by Alvin (the text looks garbled,  as if the keyboard echo for the literal elements was mapped to a bad font; numbers and direct output look good), I'd focus on sccz80.
I compiled it in many ways,  e.g. with '-DTINYMEM' and -O3 to try to save as much memory as possible. 
I also tried to move zorg.
everytime the any attempt to compute something, eg:
(* 3 4)
.. answers once with:
ERROR.
ILLEGAL FUNCTION CALL: ( . )

...the next attempts will get no answer.

aralbrec:
For sdcc you have to compile clisp.c with `--fsigned-char` because by default it makes `char` unsigned but this program expects `char` to be signed.

I have a second version in the newlib examples that worked with both sccz80 and zsdcc so I will give those a try now.

aralbrec:
Same thing happens in the newlib version - sccz80 reports that illegal function error.  The zsdcc compile crashes.  Looks like a lot of debugging ahead :-(

zx70:
Well done, you're great as always !
...now I can add to clisp.c the optional turtle graphics functions I was keeping aside   :D

zx70:
..is it impacting also zsdcc?  otherwise we can close it

suborb:
I think @aralbrec suggests that it is. I've got: `Build: 3.6.5 #9897 (Mac OS X i386) May  4 2017` and it works ok. So if it's broken there's been a regression.

zx70:
I think the same works also on the late builds.
I only have problems by typing alpha characters (garbled symbols are shown) , which happened also before,  my thought was that it was consequence of a memory overflow/ overlap.

aralbrec:
The sccz80 compile works but an optimized zsdcc compile crashes.

[...]

PR #2165, opened 2023-01-20
Fix #2148 and keep #543 working

### LINKED ISSUE #2148: ~ operator does not work as expected
main.c
```
#include "stdio.h"
#define THE_FLAG (0x10)
int main()
{
        unsigned char c = ~THE_FLAG;
        printf("the flag = %d", c);
}
```
Prints out `the flag = 255` while gcc reports 239.

My program stopped working out of sudden for no reason, and I have no idea what's causing this, but it's something somewhat recent, on the other hand, building sccz80 from a year ago didn't fix the issue either.
```
zcc +test -m -debug -create-app -lndos main.c
z88dk-ticks a.bin
```

suborb:
There's probably several more of these depth charges lying around the place....but lets fix the ones we know about.


desertkun:
Can confirm working on my arm64

PR #2167, opened 2023-01-21
sccz80: l_gcharX, l_gintX, l_gintXsp optimisations

desertkun:
These optimisations shave off 2K of code on my project.

`l_gint_const.opt` showcases the change.

- `inc hl by X`, then call `l_gint` is replaced by `l_gintX`
- `inc hl by X`, then call `l_gchar` is replaced by `l_gcharX`
- `set hl to sp + X`, then call `l_gint` is replaced by `l_gintXsp`

Example of optimisation:

```
	ld	hl,4	;const
	add	hl,sp
	call	l_gint	;
	ld	bc,8
	add	hl,bc
	call	l_gint	;
```
is replaced with
```
	call	l_gint4sp	;
	call	l_gint8	;
```

### DIFF
--- a/testsuite/results/Issue_452_pointers.opt
+++ b/testsuite/results/Issue_452_pointers.opt
@@ -18,6 +18,10 @@
 	pop	bc
 	push	hl
 	push	de
+	ld	hl,0	;const
+	call	l_gintspsp	;
+	call	l_gint4sp	;
+	pop	de
 	ex	de,hl
 	and	a
 	sbc	hl,de
@@ -37,6 +41,10 @@
 	pop	bc
 	push	hl
 	push	de
+	ld	hl,0	;const
+	call	l_gintspsp	;
+	call	l_gint4sp	;
+	pop	de
 	ex	de,hl
 	and	a
 	sbc	hl,de
@@ -58,6 +66,10 @@
 	pop	bc
 	push	hl
 	push	de
+	ld	hl,0	;const
+	call	l_gintspsp	;
+	call	l_gint4sp	;
+	pop	de
 	ex	de,hl
 	and	a
 	sbc	hl,de
@@ -81,6 +93,10 @@
 	pop	bc
 	push	hl
 	push	de
+	ld	hl,0	;const
+	call	l_gintspsp	;
+	call	l_gint4sp	;
+	pop	de
 	ex	de,hl
 	and	a
 	sbc	hl,de
@@ -103,6 +119,10 @@
 	pop	bc
 	push	hl
 	push	de
+	ld	hl,0	;const
+	call	l_gintspsp	;
+	call	l_gint4sp	;
+	pop	de
 	ex	de,hl
 	and	a
 	sbc	hl,de
--- a/lib/z80rules.1
+++ b/lib/z80rules.1
@@ -6837,3 +6837,87 @@ $notcpu 8085
 	ld	hl,%1
 	%3c	(hl)
 	ld	hl,%2
+
+%notcpu 8085
+%notcpu r2ka
+%notcpu r3k
+	inc	hl
+	inc	hl
+	inc	hl
+	call	l_gint	;
+=
+	call	l_gint3	;
+
+%notcpu 8085
+%notcpu r2ka
+%notcpu r3k
+	inc	hl
+	inc	hl
+	call	l_gint	;
+=
+	call	l_gint2	;
+
+%notcpu 8085
+%notcpu r2ka
+%notcpu r3k
+	inc	hl
+	call	l_gint	;
+=
+	call	l_gint1	;
+
+%notcpu 8085
+%notcpu r2ka
+%notcpu r3k
+%check 4 <= %1 <= 8
+	ld	bc,%1
+	add	hl,bc
+	call	l_gint	;
+=
+	call	l_gint%1	;
+
+%notcpu 8085
+%notcpu r2ka
+%notcpu r3k
+%check 1 <= %1 <= 8
+	ld	hl,%1	;const
+	add	hl,sp
+	call	l_gint	;
+=
+	call	l_gint%1sp	;
+
+%notcpu 8085
+%notcpu r2ka
+%notcpu r3k
+	inc	hl
+	inc	hl
+	inc	hl
+	call	l_gchar
+=
+	call	l_gchar3
+
+%notcpu 8085
+%notcpu r2ka
+%notcpu r3k
+	inc	hl
+	inc	hl
+	call	l_gchar
+=
+	call	l_gchar2
+
+%notcpu 8085
+%notcpu r2ka
+%notcpu r3k
+	inc	hl
+	call	l_gchar
+=
+	call	l_gchar1
+
+%notcpu 8085
+%notcpu r2ka
+%notcpu r3k
+%check 4 <= %1 <= 8
+	ld	bc,%1
+	add	hl,bc
+	call	l_gchar
+=
+	call	l_gchar%1
--- a/src/z80asm/t/issue_0341.t
+++ b/src/z80asm/t/issue_0341.t
@@ -39,17 +39,17 @@ spew("${test}1.map", @map);
 check_text_file("${test}1.map", <<END);
 __C_LINE_0_${test_expanded}_2ec = \$0000 ; addr, local, , ${test}_c, , ${test}.c:0
 __C_LINE_1_${test_expanded}_2ec = \$0000 ; addr, local, , ${test}_c, , ${test}.c:1
-__C_LINE_2_${test_expanded}_2ec_3a_3aadd_3a_3a0_3a_3a0 = \$016E ; addr, local, , ${test}_c, code_compiler, ${test}.c::add::0::0:2
-__C_LINE_3_${test_expanded}_2ec_3a_3aadd_3a_3a1_3a_3a1 = \$016E ; addr, local, , ${test}_c, code_compiler, ${test}.c::add::1::1:3
-__C_LINE_6_${test_expanded}_2ec_3a_3aadd_3a_3a0_3a_3a1 = \$017B ; addr, local, , ${test}_c, code_compiler, ${test}.c::add::0::1:6
-__C_LINE_7_${test_expanded}_2ec_3a_3amain_3a_3a0_3a_3a2 = \$017B ; addr, local, , ${test}_c, code_compiler, ${test}.c::main::0::2:7
-__C_LINE_8_${test_expanded}_2ec_3a_3amain_3a_3a1_3a_3a3 = \$017B ; addr, local, , ${test}_c, code_compiler, ${test}.c::main::1::3:8
-__C_LINE_9_${test_expanded}_2ec_3a_3amain_3a_3a1_3a_3a3 = \$017F ; addr, local, , ${test}_c, code_compiler, ${test}.c::main::1::3:9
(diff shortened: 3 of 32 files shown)


feilipu:
These will need to excluded from the 8085, as `l_gint` and friends are replaced by more efficient undocumented instructions.

desertkun:
@feilipu 

```
#notcpu 8080
#notcpu gbz80
```

These would cover the case? Seen them scattered in some places.

feilipu:
```
#notcpu 8085

```

The specific case for 8085 is necessary. It is the only one with `ld hl,(de)` and `ld (de),hl`.

Check where `#cpu 8085` is used to see what is done, and why the `l_gint` needs to stay.

It would be possible to build parallel `l_` functions using `inc de` if your C code examples looked better when compiling for 8085. Do a test to see, please.

feilipu:
Could you check your code to see whether the long runs of `inc hl` appear in a 8085 build, please?

That would inform whether it is worth doing similar 8085 optimisations.

desertkun:
> Could you check your code to see whether the long runs of `inc hl` appear in a 8085 build, please?
> 
> That would inform whether it is worth doing similar 8085 optimisations.

That us very much the case

```
	inc	hl
	inc	hl
	inc	hl
	call	l_gchar
```
and
```
	ld	hl,4	;const
	add	hl,sp
	call	l_gint	;
```
[l_gint_const.c.zip](https://github.com/z88dk/z88dk/files/10472941/l_gint_const.c.zip)
here's how to repro `z88dk-sccz80 -m8080 l_gint_const.c`

I have no idea how to properly optimize for 8080 and I have no "real" project for it that would seriously shrink size. This branch is open for writing, feel free to add commits, I'll try to not mess with them.

feilipu:
It's ok. I'll add the 8085 options in a separate PR later. I don't feel confident that something else won't go wrong.

I've been caught by function tail usage of `de` when trying something like this previously.

For now it is sufficient to preserve the `l_gint` and friends.

desertkun:
I guess I need help figuring out what's going on on windows

```
test_t_issue_0341.c::add::1::1:3: error: undefined symbol: l_gint4sp
  ^---- l_gint4sp
```

Works just fine on my mac and just built on ubuntu, why doesn't it link my new library symbols? Where the list of asm files for windows to compline are listed?

desertkun:
I see, now.
image
Windows builds do not build library, they only get a copy from nightly server.
So the only option I guess, is to commit despite the failure and wait for new nightly?

suborb:
Those builds seem to download nightly to get the libraries. 

So since it’s not merged the libraries aren’t updated.

desertkun:
Right, I'll split PR into two, get the library changes in first, those won't be used for now, and then rabase this one.

desertkun:
I opened #2168 those won't be tested, but won't be used either, just linked into the library, and then this PR will use them. That way we don't break anything.

suborb:
Back on a proper keyboard.

I think  normally the optimisations that drag in more routines are at -O3 level. The 1-3 functions are good where they are since they just prepend the regular `l_` routines.

desertkun:
> Back on a proper keyboard.
> 
> I think normally the optimisations that drag in more routines are at -O3 level. The 1-3 functions are good where they are since they just prepend the regular `l_` routines.

Are you suggesting to move the rules or ?
Also a thing to note is I do not do more calls, I just call different function, that does extra work (and doesn't call anyone on itself)

suborb:
I was wondering if *some* of the rules should move.

I'm guessing that these are called when you access a member of a struct? So chances are it's not a one-off that one of the 4-NN routines is called so we end up with a net memory saving? In which case there's no problem.

The routines in called in -O3 are generally stack related access which probably aren't called that often, thus the net saving is not guaranteed.

desertkun:
> I'm guessing that these are called when you access a member of a struct? So chances are it's not a one-off that one of the 4-NN routines is called so we end up with a net memory saving? In which case there's no problem.

Yeah like I initially said on my project I got combined savings of 2K. A few of optimisations only happened once, so "loss". But others occurred more than 200 times. And it's hard to predict which one is going to be used, because those relate to field and/or stack offset.

As per filename, I struggled a lot to understand how the numbering works. So I placed them around until it worked. You're saying some files only work with -O3? So you're suggesting to move 4+ opts there? (where then?)

feilipu:
Some rules files are only used at -O3.

So the rules should be moved into the -O2 or -O3 rules file. Not the `l_` files.

suborb:
> As per filename, I struggled a lot to understand how the numbering works. So I placed them around until it worked. You're saying some files only work with -O3? So you're suggesting to move 4+ opts there? (where then?)

I was originally suggesting moving them. But I think on average they work out to be a net-gain. So leave them as is.

desertkun:
(I'll rerun win32 builds when nightly comes)

feilipu:
Perhaps you want to kill these rules for rabbit too.
There are some [regressions shown in your test code](https://github.com/z88dk/z88dk/pull/2167/files#diff-e73615d674a9aa8ec3dff8ffb4d93dd1d1377eb247bcf55c1215c2882712b1d3).

Just, I'm not sure what the `%notcpu` code is for rabbit processors. :man_shrugging: 
`%notcpu rcm` or `%notcpu rabbit` is likely, but IDK?

suborb:
> **EDIT** I think [it will be](https://github.com/z88dk/z88dk/blob/master/src/sccz80/main.c#L95)
> 
> ```
> %notcpu r2ka
> %notcpu r3k
> ```

It is now: 7c66a80f32efa5c8e8b2121c2374174b82c49f46

desertkun:
@feilipu something like [this](https://github.com/z88dk/z88dk/pull/2167/files#diff-930d34d36f5b4e35e97fc94331148d39428d2f4a304343bb66b939237ece1aedR6842)?

feilipu:
Yes that seems right.

But I'm wondering about why there are still regression needed in the rabbit optimisation test file. Is it just not updated yet?

There should be no need to do calls to the new functions, when the rabbit has a suitable op code built in.

feilipu:
Also I'm not sure how rabbit code path is generated.

The underlying issue is why the rabbit optimisation tests need to be changed. This needs to be clarified, imho.

desertkun:
> Also I'm not sure how rabbit code path is generated.

> There should be no need to do calls to the new functions, when the rabbit has a suitable op code built in.

There's one testcase for rabbit `testsuite/results/rabbit/rabbit_optimisations.opt`. It didn't work since I forgot to rebase, but since I did, it now expects original code on rabbit. Hope that addresses your concern.

I still modified it since there are a few spaces needed to be fixed.

desertkun:
@feilipu  @suborb  good to go?

testsuite/results/Issue_452_pointers.opt
```diff
@@ -18,6 +18,10 @@
 	pop	bc
 	push	hl
 	push	de
+	ld	hl,0	;const
```
suborb, line 21: Looks like there's a later optimisation that is removed by the new rules you've added

desertkun: How do I make mine run after?

suborb: The only guaranteed way is to put them in a different file.

The rules run in this order (according the -O level):

```
                switch (peepholeopt) {
                case 0:
                    break;
                case 1:
                    rules[num_rules++] = c_coptrules1;
                    break;
                case 2:
                    rules[num_rules++] = c_coptrules2;
                    rules[num_rules++] = c_coptrules1;
                    break;
                default:
                    rules[num_rules++] = c_coptrules2;
                    rules[num_rules++] = c_coptrules1;
                    rules[num_rules++] = c_coptrules3;
                    break;
                }

    { "COPTRULES1", 0, SetStringConfig, &c_coptrules1, NULL, "", "\"DESTDIR/lib/z80rules.1\"" },
    { "COPTRULES2", 0, SetStringConfig, &c_coptrules2, NULL, "", "\"DESTDIR/lib/z80rules.2\"" },
    { "COPTRULES3", 0, SetStringConfig, &c_coptrules3, NULL, "", "\"DESTDIR/lib/z80rules.0\"" },
```

So we'd add an extra one to run after z80rules.1 in all cases.

Before making changes to the zcc driver, pull them into a separate file and then compile with `-custom-copt-rules=filename` and make sure you get the result you expect.

desertkun: Thanks, will look into it.

feilipu: Also, look at the code section without optimisation to see precisely which rule is (not) being triggered.

Perhaps the rule can be tweaked or reordered slightly relative to the new rules, which would be a light touch option to try?

desertkun:
@suborb sadly, with moving out the rules to a rules.3, I noticed a severe degradation on my project

before any rules: 39917
with rules.3: 38973
with rules.1: 38161

With rules.3 they are still applied, but I guess other optimisations get in the way somehow.

I also applied the same to examples, which we could use as benchmark, with this one-liner
```bash
for d in $(find . -name '*.tap') ; do (ls -l $d); done | awk '{print $5}' | paste -sd+ - | bc
```

And didn't get much of an improvement for some reason. So need to investigate both, I guess.

suborb:
Okay, thanks for trying it out, easiest thing to do is to merge them and then I'll test out on a few examples that I've got to see the effect on them.

The examples in the repo are fairly simple so probably not too many offsets that would be hit.

desertkun:
You mean w/o extra file, as it is? That would be more favorable to me. If you're ok, feel free to merge.

suborb:
I've got a plan to protect some of these rules behind extra options if they cause too much bloat in smaller programs.

suborb:
I temporarily moved these rules so they're protected by the "size" tag and took advantage of 534d7a938ba82351244be6b5bacb18e4e88f046c

### MD5

I've got a MD5 program that's I've used a fair bit for tuning in the past:

```
dom@ermintrude md5 % zcc +test md5sum.c -Ch-Osize
md5sum.c:369:28: warning: Converting integer type to pointer witout cast. From unsigned int  to char * fmt*  [-Wincompatible-pointer-types]
md5sum.c:369:31: warning: Assigning 'v', type: unsigned char * v from char * fmt*  [-Wincompatible-pointer-types]
ls -al a.bin%
dom@ermintrude md5 % ls -al a.bin
-rw-r--r--  1 dom  staff  13855 23 Jan 20:29 a.bin
dom@ermintrude md5 % z88dk-ticks -w 30 a.bin -- testfile.bin
Fname: testfile.bin
996d17ca7b233f87a94325b21762dc35
Ticks: 42892990
dom@ermintrude md5 % zcc +test md5sum.c
md5sum.c:369:28: warning: Converting integer type to pointer witout cast. From unsigned int  to char * fmt*  [-Wincompatible-pointer-types]
md5sum.c:369:31: warning: Assigning 'v', type: unsigned char * v from char * fmt*  [-Wincompatible-pointer-types]
dom@ermintrude md5 % ls -al a.bin
-rw-r--r--  1 dom  staff  13968 23 Jan 20:29 a.bin
dom@ermintrude md5 % z88dk-ticks -w 30 a.bin -- testfile.bin
Fname: testfile.bin
996d17ca7b233f87a94325b21762dc35
Ticks: 42826879
```

So a saving of > 100 bytes, at a cost of about 0.1% is good.

PR #2168, opened 2023-01-21
sccz80: library for l_gcharX, l_gintX, l_gintXsp optimisations

desertkun:
Library changes for #2167


desertkun:
@suborb thinking to merge this as soon as it passes, unless you have review objections. This is not regular ticks stuff 😄

suborb:
This one is good to go since it has no effect.

PR #2170, opened 2023-01-22
Issue #2148: More fun around edgecases of 64 bit numbers

suborb:
Seems to fix the issue on arm64 for me.


desertkun:
```
Running test test_rshift16_const.....failed rshift.c:146 (>>1)
Running test test_rshift16_var.....failed rshift.c:14 (>>1)
9 run, 7 passed, 2 failed
```

unless I am doing something wrong? I am basically doing `./build.sh -t`

suborb:
Can you try rebuilding sccz80 manually - I'm wondering if it's not being rebuild for some reason.

desertkun:
I think I rebuilt everything, cleaned up git repo for good measure, still errors out.

suborb:
Hmm. I‘ve no way of debugging until I upgrade. 

It’s really odd that it’s the 16 bit case failing

I know that this does fix a problem on Linux so I’ll merge anyway.

PR #2171, opened 2023-01-22
Fix color escape codes on Windows

empathicqubit:


suborb:
What does this do exactly? I was under the impression that cmd.exe already understood vt100 codes?

empathicqubit:
It does, but you have to turn it on first.

empathicqubit:
clers
terminal
pwsh

empathicqubit:
So this fixes all of them. It also fixes it in VSCode which is the only one I actually care about. There is still a problem where the arrows won't work in Git Bash or VSCode.

suborb:
Ah cool, thank you for the code + explanation!

PR #2172, opened 2023-01-23
sccz80 - 8085 l_gchar l_gint optimisation

feilipu:
Added the optimisations from #2167 for 8085.
Fixed a few other rules that looked wrong for 8085.

### DIFF
--- a/lib/z80_crt0.hdr
+++ b/lib/z80_crt0.hdr
@@ -70,7 +70,7 @@
 	EXTERN	l_gint6sp
 	EXTERN	l_gint7sp
 	EXTERN	l_gint8sp
-	EXTERN	l_gint1	    ;Optimized variants inc hl,x ld hl,(hl)
+	EXTERN	l_gint1		;Optimized variants inc hl,x ld hl,(hl)
 	EXTERN	l_gint2
 	EXTERN	l_gint3
 	EXTERN	l_gint4
--- a/lib/z80rules.1
+++ b/lib/z80rules.1
@@ -445,14 +445,13 @@
 =
 	ld	hl,(_%1)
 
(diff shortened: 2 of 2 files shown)


desertkun:
@feilipu @suborb 
I wrote a simple tool that would take assembly output and hunt for the most used patterns of lines
```bash
z88dk-dis -x build/client.map -o 25000 build/client.bin > dump.txt
z88dk-dis -x build/server.map -o 25000 build/server.bin > dump2.txt
python3 cosize.py dump.txt dump2.txt
```

It would dump, e.g.
```
148 x 4 = 592:
pop     bc
pop     hl
push    hl
push    bc

152 x 3 = 456:
pop     bc
pop     hl
push    hl

148 x 3 = 444:
pop     hl
push    hl
push    bc

125 x 3 = 375:
add     hl,sp
ld      l,(hl)
ld      h,$00

59 x 6 = 354:
ld      h,$00
dec     sp
ld      a,l
pop     hl
ld      l,a
push    hl

65 x 5 = 325:
dec     sp
ld      a,l
pop     hl
ld      l,a
push    hl

60 x 5 = 300:
ld      h,$00
dec     sp
ld      a,l
pop     hl
ld      l,a

66 x 4 = 264:
dec     sp
ld      a,l
pop     hl
ld      l,a

etc
```

Nice little tool to find something to optimize. A lot of these could be pushed into size-over-performance kind of optimisations (which would be triggered how?)

Works if multiple files, too, so you can dump it all files you can grep (e.g. compile examples and get z88dk-dis from them) and see what's is the most of demand.

[cosize.py.zip](https://github.com/z88dk/z88dk/files/10484628/cosize.py.zip)

PR #2176, opened 2023-01-24
8085: Some more low hanging fruit

suborb:
Few more rules for 8085 - cleanup some odd looking generated code + be more aggressive using ld de,sp+NN where we know we're going to reload de.

This saves around about 5% or so on the md5sum testcase and reduces program size quite a lot:

This branch: 13215 bytes, 47555113 ticks
Master: 13905 bytes, 49315247 ticks

### DIFF
--- a/lib/arch/8085/8085_rules.1
+++ b/lib/arch/8085/8085_rules.1
@@ -113,3 +113,24 @@
 	ld	b,d
 	ld	de,sp+%2
 	ld	hl,(de)
+
+%check 0 <= %1 <= 255
+	ld	hl,%1	;const
+	add	hl,sp
+	call	l_glong%2
+=
+	ld	de,sp+%1
+	ex	de,hl
+	call	l_glong%2
+
+
+%check 0 <= %1 <= 255
+	ld	hl,%1	;const
+	add	hl,sp
+	push	hl
+	call	l_glong%2
+=
+	ld	de,sp+%1
+	ex	de,hl
+	push	hl
+	call	l_glong%2
--- a/lib/z80rules.1
+++ b/lib/z80rules.1
@@ -6980,3 +6980,36 @@ $notcpu 8085
 	call	l_gchar
 =
 	call	l_gchar%1
+
+%title Clear up a misoptimisation
+%cpu 8085
+	push	de
+	ld	hl,(de)	;l_gint
+	%2c	hl
+	pop	de
+	ld	(de),hl	;l_pint
+=
+	ld	hl,(de)	;l_gint
+	%2c	hl
+	ld	(de),hl	;l_pint
+
+%cpu 8085
+%check 0 <= %1 <= 255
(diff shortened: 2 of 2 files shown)


feilipu:
As a related question... is there a plan behind having 8085 optimisation in both `z80_rules.` and in `arch/8085/8085_rules.1`?
I guess the question extends to gbz80 at least too.

Also, the `8085_rules.1` file is only run for m100 and test targets, to my knowledge, so currently it is a bit of a waste of effort.
Perhaps this needs to be rationalised, somehow.

suborb:
My thinking is that the `arch/` rules are run as a final step and can do some of the cleanup/cpu specific optimisations after the other files have run.

I suspect that some of the 8085 rules in z80rules.1 are stopping the generic case from running - eg the push/inc/pop case I fixed here is optimised for z80 and I think one of the ld de,sp+NN rules is stopping it from running.

I've probably not been entirely consistent with it though.

Ah, I missed adding it to the cpm and micro85 config files - I'll do that tomorrow.

feilipu:
> Ideally, if we don't want a rule to run for a chip in z80rules.N (because we can optimise better later on), then that should be marked with %notcpu rather than adding %cpu rule which stops the original rule from running.

> I've probably not been entirely consistent with it though.

I've probably messed things up there too, by not knowing about the arch/8085 file. My bad.
Perhaps another clean-up PR at some stage.

> Ah, I missed adding it to the cpm and micro85 config files - I'll do that tomorrow.

Is it possible to add it to run in a subtype?
If so, then while you're at it the +rc2014 basic85 and acia85 should both have it added too.
Thanks 🙏

suborb:
> Is it possible to add it to run in a subtype? If so, then while you're at it the +rc2014 basic85 and acia85 should both have it added too. Thanks 🙏

It is, but I added them to the clib line since that made more sense. Hopefully that's ok

feilipu:
At some point in the next while I'll reorganise to move all the 8085 specific rules into `arch/8085/8085_rules.1`.

I'll check to see that there are no regressions on the md5 code, or on my test code (mainly fatfs). But otherwise it will be just tidying my mess.

desertkun:
> I'll check to see that there are no regressions on the md5 code, or on my test code (mainly fatfs). But otherwise it will be just tidying my mess.

At this point it should be part of CI runs. I can dump memory, can we make it dump only area of it, in hex format? And maybe even compare that to existing hash?

feilipu:
> At some point in the next while I'll reorganise to move all the 8085 specific rules into `arch/8085/8085_rules.1`.

I'm suddenly reminded that I rely on z80_rules.8 to run last (to provide size/speed trading options). 

Specifically I've left `l_sxt` in all 8085 char optimisations.

Adding all 8085 `l_gchar` and `l_gint` stuff to run in a custom file after z80_rules.8 is not going to work.

Is that thought correct?

PR #2177, opened 2023-01-26
sccz80: save a byte on each debug call

desertkun:
Saves a ret and/or pop bc before it


desertkun:
Ugh I opened it prematurely, it crashes my program somehow

Issue #2179, opened 2023-01-27
non-z80 - optimisation rules


feilipu:
I'm looking at this 8085 (and also gbz80) optimisation question.

For reference, I find this current situation.

- `z80_rules.9` is always run first

- -01 - `z80_rules.1` is run

- -02 - `z80_rules.2` then `z80_rules.1`

- -03 - `z80_rules.2` then `z80_rules.1` then `z80_rules.0`

- --opt-code-speed - `z80_rules.8` is always run last (except arch file) irrespective of Ox.

- -m8085 - `8085_rules.1` is always run last.

For gbz80 there are lots of rules using `ex de,hl` which are not excluded for gbz80. Which I guess means that sccz80 will not generate `ex de,hl` for gbz80, otherwise things would be totally broken. Right?

To Do @feilipu: If that is the case, then the few remaining rules `%notcpu gbz80` protection for input `ex de,hl` can be removed (noting there are other opcodes that it also can't do, and obviously excluding those with `ex de,hl` as an output).

For 8085 the `--opt-code-speed` `z80_rules.8` rules remove all of the potential architecture specific `l_gint` `l_gchar` improvements as it is run _before_ the architecture specific file.

To Do @suborb: This should to be changed, so that "speed" optimisation `rules.8` is run after the architecture file, or the `--opt-code-speed` optimisation file is associated per architecture and also run last.

feilipu:
> This should to be changed, so that "speed" optimisation rules.8 is run after the architecture file, or the --opt-code-speed optimisation file is associated per architecture and also run last.

It looks like zcc is already set up to handle an additional CPU related optimisation step, before speed optimisation. I'll have a look at putting the arch files to run there as CPU optimisation, which solves this ordering issue perfectly.

It also leaves the user optimisation step open too.

feilipu:
I think for the gbz80 there is an opportunity for adding `LD HL,SP+int8` instruction optimisation, which at 12 ticks would be faster than `LD HL,int16` `ADD HL,SP` with 20 ticks.

Also `ADD SP,int8` at 16 ticks is still better than `LD HL,int16` `ADD HL,SP` `LD SP,HL` with 28 ticks.

feilipu:
If the gbz80 `add sp+int8` is really an integer, then it would be worth changing the boundaries on the rule checks to allow negative number results, -128 through 127.

This would be useful for establishing the stack frame, if for no other reason.

suborb:
[...]

PR #2180, opened 2023-01-28
sccz80 - 8085 arch optimisation rules

feilipu:
Working on #2179 

Initially a hack into `zcc.c` to set `COPTRULESCPU` to run COPT the 8085 `arch/8085/8085_rules.1` file before the `-opt-code-speed` `z80_rules.8` file. This hack needs to be backed out and replaced with a proper all platform setting to work with all CPU types based on the `-m` command line option.


suborb:
Do we not just need an options that set c_coptrules_cpu and c_coptrules_sccz80 from the command line so that we can deal with multi-cpu targets?

That way single architecture targets can just rely on the `COPTRULES` config options and we can override them on a per subtype/clib basis for the different cpus.

feilipu:
> Do we not just need an options that set c_coptrules_cpu and c_coptrules_sccz80 from the command line so that we can deal with multi-cpu targets?

IMHO, there's no need to yet add another command line option. The `-m` machine type already discerns the correct arch `arch-rules.1` file to be run. It is just a matter of populating the `c_coptrules_cpu` with a string pointing to the correct arch file.

As an extension (totally not suggesting this, but it would be possible), there is also a target specific optimisation file foreseen in zcc, which could be populated by a file somewhere standardised in the target directory, which could cover target custom opcodes, etc.

Then that leaves the user opt file free to be added from the command line.

All of these above are run before the `-opt-code-speed` related rules too.

suborb:
> IMHO, there's no need to yet add another command line option. The `-m` machine type already discerns the correct arch `arch-rules.1` file to be run. It is just a matter of populating the `c_coptrules_cpu` with a string pointing to the correct arch file.

Good point!

> As an extension (totally not suggesting this, but it would be possible), there is also a target specific optimisation file foreseen in zcc, which could be populated by a file somewhere standardised in the target directory, which could cover target custom opcodes, etc.

I think I added this in, in case there was a target where, for example l_gint is implemented by a rst

feilipu:
> It is just a matter of populating the `c_coptrules_cpu` with a string pointing to the correct arch file

The reason that I'm not doing this is that I don't know that I can get the machine type and directory creation string parsing / tokenisation across all win/linux/macos platforms anywhere close to correct. Best to leave this to experts. :smile:

feilipu:
Ok, now the rules are moved over, with no optimisation regression.
`-O0` is actually much better, as the `ld de,sp+%1` opcodes are picked up.

Let's see if things can be tided and rationalised.

feilipu:
Ok. Pending the changes to `zcc.c` this is done.
- removes special opt handling in config files
- removes 8085 specifics out of `z80_rules.1`
- adds 8085 specifics into `arch/8085/8085_rules.1`

suborb:
I'm working on the zcc bit which I'll merge into this branch before merging the whole lot together.

suborb:
I've finished tinkering for the evening now.

The --opt-code-speed options actually seem to slow down the resulting binary now - I get the best results for z80 with just a plain -O2.

feilipu:
> I've finished tinkering for the evening now.

I've finished on this too, so I'll merge it once it has passed CI.

I [rebuilt CP/M-IDE for 8085](https://github.com/RC2014Z80/RC2014/commit/c1361da93ba321c6254cbe05be7d31a72e71319b?diff=split) yesterday, and found that the resulting binaries to be over 400 bytes smaller. This reduction comes from the [FATFS library](https://github.com/feilipu/z88dk-libraries/tree/master/ff/rc2014/lib/newlib/sccz80) mainly, [with 337 bytes](https://github.com/feilipu/z88dk-libraries/commit/144564019b9a7293a95cf6df6cf5a72239b99096), but also a small amount from the [shell](https://github.com/RC2014Z80/RC2014/blob/master/ROMs/CPM-IDE/acia85/main.c). The 8085 binaries (at 75.7kB) are now substantially smaller than equivalent functionality z80 binaries (at 79.9kB). So I think we've squeezed quite enough blood out of that rock for now. 😉

If I get time over the week, I'll rerun some benchmarking to see how the changes reflect in those numbers.

feilipu:
Closes #2179

PR #2188, opened 2023-02-05
GDB: Add tracecmds feature. Change x to allow specifying /length

empathicqubit:
Input script:

```sh
define hook-stop blarg
x /2 storyPC
x /10 instructionBuffer
end
trace on
break doneProcessInstruction
cont
cont
```

automatic_cmds

### DIFF
--- a/src/ticks/debugger.c
+++ b/src/ticks/debugger.c
@@ -175,7 +175,7 @@ static command commands[] = {
     { "set",       cmd_set,         "<hl/h/l/...> <value>", "Set registers" },
     { "out",       cmd_out,         "<address> <value>",    "Send to IO bus"},
     { "trace",     cmd_trace,       "<on/off>",             "Disassemble every instruction"},
-    { "define",    cmd_define,   "define {hook_stop|...}",        "Specify a command block:\n\t\t\t\t\thook-stop: commands to execute when trace occurs"},
+    { "define",    cmd_define,   "define {hook-stop|...}",        "Specify a command block:\n\t\t\t\t\thook-stop: commands to execute when trace occurs"},
     { "hotspot",   cmd_hotspot,     "<on/off>",             "Track address counts and write to hotspots file"},
     { "profiler",  cmd_profiler,    "[-f fun][-i iter]",    "start/stop profiling (-f function limit, -i iteration limit)"},
     { "list",      cmd_list,        "[<address>]",          "List the source code at location given or pc"},
@@ -421,7 +421,7 @@ static void debugger_print_prompt(char *prompt, size_t prompt_size, char *buf, s
 			snprintf(prompt, prompt_size, "\n" FNT_BCK "    $%04x    (%s)>" FNT_RST " ", bk.pc(), buf);
 		}
 	} else {
-		// Original output for non-active tty 
+		// Original output for non-active tty
         snprintf(prompt, prompt_size, " %04x (%s)>", bk.pc(), buf);
 	}
 }
--- a/src/ticks/debugger.c
+++ b/src/ticks/debugger.c
@@ -175,7 +175,7 @@ static command commands[] = {
     { "set",       cmd_set,         "<hl/h/l/...> <value>", "Set registers" },
     { "out",       cmd_out,         "<address> <value>",    "Send to IO bus"},
     { "trace",     cmd_trace,       "<on/off>",             "Disassemble every instruction"},
-    { "define",    cmd_define,   "define {hook-stop|...}",        "Specify a command block:\n\t\t\t\t\thook-stop: commands to execute when trace occurs"},
+    { "define",    cmd_define,   "define {hook-stop|...}",        "Specify a command block:\n          \t                    \t    hook-stop: commands to execute when trace occurs"},
     { "hotspot",   cmd_hotspot,     "<on/off>",             "Track address counts and write to hotspots file"},
     { "profiler",  cmd_profiler,    "[-f fun][-i iter]",    "start/stop profiling (-f function limit, -i iteration limit)"},
     { "list",      cmd_list,        "[<address>]",          "List the source code at location given or pc"},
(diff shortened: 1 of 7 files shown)


empathicqubit:
Input script:

```sh
tracecmds x /2 storyPC ; x /10 instructionBuffer
trace on
break doneProcessInstruction
cont
cont
```

automatic_cmds

suborb:
I really like this feature and want to extend it to everything (step, stepi, next, when a break/watch point fires)

empathicqubit:
Okay! Tracecmds runs when you do those things as the normal trace does. tracecmds will parse the input as provided, including invalid input, so you can give it anything, but commands which resume the execution will only do so after all of them are processed.

The script input takes any commands.

empathicqubit:
Okay, I guess I'm not passing up the return flag, but I could, if that's what you meant

empathicqubit:
What else I should change so you can merge it? I'm not sure I understood what you wanted

suborb:
Sorry, I've had a hectic IRL week and completely forgot about this.

I was suggesting that adding a capability to run commands after stepping/nexting would be really useful - often when I'm debugging I want to print the registers everytime I come back to the prompt.

desertkun:
Is `tracecmds` an official UI for regular gdb? Ideally we want to adjust to that UI as much as possible, so people with no context with z88dk gdb can instantly get a hold on how to use it.

If you're touching `x`, can you try go all the way, like gdb does? e.g `x/32xb` allows to specify not only size of the block, but split the block into pieces, depending on last "argument", could be bytes, shorts, ints, longs

desertkun:
I looked it up, I think introducing concept of functions (or blocks) https://stackoverflow.com/a/30999466 is how GDB does it.
```
define hook-stop
print x
end
```
GDB auto-calls `hook-stop` block on every break. This can be expanding further by using same code for `.gdbinit`, should we ever need it.

desertkun:
I like how you separated the console input and command execution. Does that apply for mi/2 interface as well?
But `debugger_evaluate_multiple` could be reworked to work on null-terminated `char**` array instead of splitting it by `;`, that way we can get more gdb-alike by using `define`.

empathicqubit:
Sorry I lost track of this. I might pick it up again soon

empathicqubit:
The block commands thingy is implemented now. The test script looks like this:

```sh
define hook-stop blarg
x /2 storyPC
x /10 instructionBuffer
end
trace on
break doneProcessInstruction
cont
cont
```

Can we do the format changes in a different PR? For now it seems to gracefully ignore it

empathicqubit:
Does this work or do you want to wait until the number formatting stuff is in?

desertkun:
Sorry I've missed the recent change. Let me take a look in few hours. Good stuff.

desertkun:
Works awesome, cool feature. Looks like lots of changes, looking thru them one. Quick comment if you don't mind, could you please replace tabs with 4 spaces to keep consistency?

src/ticks/debugger.c
```diff
@@ -170,10 +171,11 @@ static command commands[] = {
     { "break",     cmd_break,       "<address/label>",      "Handle breakpoints" },
     { "delete",    cmd_del_break,   "<breakpoint id>",      "Delete breakpoint(s)" },
     { "watch",     cmd_watch,       "<address/label>",      "Handle watchpoints" },
-    { "x",         cmd_examine,     "<address>",            "Examine memory" },
+    { "x",         cmd_examine,     "[/<length>] <address>","Examine memory" },
     { "set",       cmd_set,         "<hl/h/l/...> <value>", "Set registers" },
     { "out",       cmd_out,         "<address> <value>",    "Send to IO bus"},
     { "trace",     cmd_trace,       "<on/off>",             "Disassemble every instruction"},
+    { "define",    cmd_define,   "define {hook_stop|...}",        "Specify a command block:\n\t\t\t\t\thook-stop: commands to execute when trace occurs"},
```
desertkun, line 178: Should be `hook-stop` here? (hyphen instead of underscore)

desertkun:
Yeah, besides that hyphen fix and tabs, looks good.

empathicqubit:
I did a squash. I checked the diff and it seems correct but you might want to double check that somehow

desertkun:
I think you might have missed some places
image

empathicqubit:
The diff said it changed something 🤷‍♀️

empathicqubit:
Your editorconfig seems to want tabs

empathicqubit:
Should I change the editorconfig

empathicqubit:
https://github.com/z88dk/z88dk/blob/1fbc247efad40f3a6ae940f62e5f0985a9f8a840/src/.editorconfig#L6

desertkun:
Sorry I am not sure about that editor config, but the source code in all places has been 4 spaces. I think changing that in that file too would bee a good choice.

empathicqubit:
Okay. The reason I ask is because it is messing up VSCode. Legacy Visual Studio has an extension for it too, but I had trouble finding for 2022 for some reason. I'll fix it.

empathicqubit:
Document formatting wants to change a few other things too for some reason

empathicqubit:
It looks like there's a lot of lines with 2 space indent in ticks.c. Also the file is really huge and probably needs to be refactored somewhat

empathicqubit:
Crap

empathicqubit:
Sorry for the noise. It's hopefully fine now

empathicqubit:
I guess something from master broke the tests

desertkun:
I see you formatted the whole ticks.c, which I cannot feasibly verify. Since this seems like a lot of hussle over a minor thing, I am willing to go back to having some tabs. @suborb your opinion?

If you don't have those changes anymore, I can force-push into your branch, the old changes I pulled a few hours back.

suborb:
> I see you formatted the whole ticks.c, which I cannot feasibly verify. Since this seems like a lot of hussle over a minor thing, I am willing to go back to having some tabs. @suborb your opinion?

Revert ticks.c - it's hard enough to reason with without it being auto formatted! A few tabs slipping through is fine if we can restore it.

desertkun:
@empathicqubit what's your action item? If you revert ticks.c and make it still compile and work, that'd be great so we can merge this awesome change.

suborb:
FYI: Windows build failures are expected until the next nightly build.

empathicqubit:
Okay, I pulled back the old attempt and readded the editorconfig change

empathicqubit:
Thank you! 🥳

PR #2190, opened 2023-02-06
zsdcc - v4.3.0

### LINKED ISSUE #2186: Test current sdcc (before 4.3.0 release)
SDCC developers intend to do a 4.3.0 release soon. While the official release process with code freezes, etc hasn't been started yet, this seems like a good time to test if z88dk works well with SDCC from current svn, and report any found SDCC issues upstream, while there is still a chance to fix them before 4.3.0.

SDCC progress since 4.2.0 was mostly in improved standard compliance; a large number of bugs have been fixed, too.

feilipu:
Closes #2186

 -  testing with r13854

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,13 +32,13 @@ ifneq (, $(shell which ccache))
    CC := ccache $(CC)
 endif
 
-Z88DK_PATH	= $(shell pwd)
-SDCC_PATH	= $(Z88DK_PATH)/src/sdcc-build
-SDCC_VERSION    = 13131
+Z88DK_PATH		= $(shell pwd)
+SDCC_PATH		= $(Z88DK_PATH)/src/sdcc-build
+SDCC_VERSION	= 13854
 
 ifdef BUILD_SDCC
 ifdef BUILD_SDCC_HTTP
-SDCC_DEPS = zsdcc_r$(SDCC_VERSION)_src.tar.gz
+SDCC_DEPS		= zsdcc_r$(SDCC_VERSION)_src.tar.gz
 endif
 endif
 
--- a/changelog.txt
+++ b/changelog.txt
@@ -14,6 +14,7 @@ Detailed but incomplete changelist:
 - [sccz80] 8080, 8085, gbz80 opt-code-speed: add32, sub32, longcompare
 - [sccz80] Extended quikmult cases
 - [z80asm] Allow jr across sections
+- [zsdcc] Upgraded to SDCC 4.3.0 r13xxx
 - [zcc] +config at any place, support for -xc
 - [make] add cmake support
 
--- a/src/zsdcc/readme.md
+++ b/src/zsdcc/readme.md
@@ -8,7 +8,7 @@ For Linux users follow the instructions on the [installation page](https://githu
 
 `sdcc-z88dk.patch` is the current default standard patch.
 
-`sdcc-13131-backport-z88dk.patch` is the current zsdcc patch, retained for comparison and building against sdcc 4.2.0 r13131.
+`sdcc-13xxx-z88dk.patch` is the current zsdcc patch, retained for comparison and building against sdcc 4.3.0 r13xxx.
 
 `sdcc-9958-z88dk.patch` is the previous z88dk v1.99c zsdcc standard patch, retained for comparison and building against sdcc r9958.
 
--- a/src/zsdcc/sdcc-13854-z88dk.patch
+++ b/src/zsdcc/sdcc-13854-z88dk.patch
@@ -1,41 +1,6 @@
-Index: ChangeLog
-===================================================================
---- ChangeLog	(revision 13131)
-+++ ChangeLog	(working copy)
-@@ -1,3 +1,24 @@
-+2022-06-25 Philipp Klaus Krause <krauseph AT informatik.uni-freiburg.de>
-+
-+	* src/z80/gen.c,
-+	  support/regression/tests/bug-3458.c.in:
-+	  Fix bug #3458.
-+
-+2022-06-01 Philipp Klaus Krause <krauseph AT informatik.uni-freiburg.de>
-+
-+	* src/SDCCast.c:
-+	  Add an assertion to show the underlying issue of bug #3401.
-+
-+2022-05-13 Philipp Klaus Krause <krauseph AT informatik.uni-freiburg.de>
-+
-+	* src/z80/gen.c:
-+	  Fix bug #3388.
-+
-+2022-03-28 Philipp Klaus Krause <krauseph AT informatik.uni-freiburg.de>
-+
-+	* src/z80/ralloc2.cc:
-+	  Fix code generation issues in use of iy.
-+
- 2022-02-26 Philipp Klaus Krause <krauseph@informatik.uni-freiburg.de>
- 
- 	* src/SDCCsymt.c,
-@@ -48206,4 +48227,4 @@
- 	* as/mcs51/lklex.c:  accept everything as symbol name in rel-files, bug fix ID 452601
- 	* as/mcs51/lksym.c:  accept everything as symbol name in rel-files, bug fix ID 452601
- 
--$Revision: 13081 $
-+$Revision: 13131 $
 Index: src/SDCCasm.c
 ===================================================================
---- src/SDCCasm.c	(revision 13131)
+--- src/SDCCasm.c	(revision 13854)
 +++ src/SDCCasm.c	(working copy)
 @@ -403,8 +403,8 @@
  static const ASM_MAPPING _asxxxx_mapping[] = {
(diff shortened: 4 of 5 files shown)


feilipu:
@suborb could you put a http://nightly.z88dk.org/zsdcc/zsdcc_r13854_src.tar.gz file on the server, please?
And rerun the actions, if you could too please.

Makefile
```diff
@@ -118,8 +118,8 @@ endif
 
 bin/z88dk-zsdcc$(EXESUFFIX): $(SDCC_PATH)/bin/sdcc$(EXESUFFIX)
 ifdef BUILD_SDCC
-	cp $(SDCC_PATH)/bin/sdcc$(EXESUFFIX)  $(Z88DK_PATH)/bin/z88dk-zsdcc$(EXESUFFIX)
-	cp $(SDCC_PATH)/bin/sdcpp$(EXESUFFIX) $(Z88DK_PATH)/bin/z88dk-zsdcpp$(EXESUFFIX)
+	cp $(SDCC_PATH)/src/sdcc$(EXESUFFIX)  $(Z88DK_PATH)/bin/z88dk-zsdcc$(EXESUFFIX)
```
suborb, line 121: ?? Seems a bit odd

feilipu: The executables are no longer copied to `sdcc/bin` on creation. There are just scripts from `sdcc/bin` that call them in place.

So to find the actual executables you have to dig a bit deeper.

Had me guessing for a few minutes, why the `sdcc` was suddenly only 200 odd Bytes.

You don't normally "open" a bin file to check the contents, so I thought the build had gone bad on first pass.

feilipu:
Based on the comment from Felix https://github.com/z88dk/z88dk/issues/2186#issuecomment-1424784546, it looks like we'll have to "install" the sdcpp binary somewhere in the build tree using the gcc-12 make file first, and then rename/copy it to the final destination in z88dk.

I've had a read of the `~/sdcc/support/cpp` make file, and the top level sdcc makefile, and, based on the salad I'm reading, I'm afraid it is beyond me to produce a fix (let alone one that works cross platform).

If you want to take this on, please do.

Makefile
feilipu: See https://github.com/z88dk/z88dk/issues/2186#issuecomment-1424784546

feilipu:
Just thinking, I might be over thinking this.

The builds are all on VMs so a default `make install` and copy from the default location to `z88dk/bin` should work.

Hmm.

feilipu:
Hmm. Failing because the action doesn't run with root, so this will need to be fixed.

Also the `zsdcc` refers to `cc1`, which is probably related to `gcc` but to be confirmed, which will break things later in the build test.

IDK why the preprocessor needs to be hardwired to a specific `cc1`.

feilipu:
Of course these changes are going to trample on a user's vanilla `sdcc` installation, by overwriting the binaries in `/user/local/bin`, which is very bad.

feilipu:
Failing now because the action VM is not truly running at root, so the copy can't find where the binary was installed... :man_shrugging: 
```
cp /home/runner/work/z88dk/z88dk/src/sdcc-build/usr/local/bin/sdcc  /home/runner/work/z88dk/z88dk/bin/z88dk-zsdcc
cp: cannot stat '/home/runner/work/z88dk/z88dk/src/sdcc-build/usr/local/bin/sdcc': No such file or directory
make: *** [Makefile:122: bin/z88dk-zsdcc] Error 1
Error: Process completed with exit code 2.
```

ped7g:
try `make install` with something like `PREFIX` into some folder withing the container environment maybe even the target location where you were planning to copy it to? (just general idea, I didn't check z88dk setup, so not sure if you do support prefix install, and also no idea where is good folder in the VM as target location, but I guess you can avoid sudo this way by writing to folder where you have already access)

feilipu:
> try make install with something like PREFIX into some folder within the container environment maybe even the target location where you were planning to copy it to.

Good plan. But since I don't have the VMs to had to test before committing, I'll defer here to @suborb. Otherwise the history will be filled with my random fumbling.

Also need to find out why the `cc1` (gcc compiler) is hardwired to `sdcpp`.

suborb:
Not sure I've done anything beyond remove the `sudo`...

The cc1 problem is a blocker, I can't remember if `z88dk-ucpp` is good enough to prepare the input for sdcc but that might be a work around pending another sdcc release.

feilipu:
> The cc1 problem is a blocker!

I've subscribed to their related issues, to be aware of how and when this gets fixed.

Philipp also raised an issue that `sdcc` no longer works with debuggers, as it is a script and can't find the `sdcpp` binary. That will need to be fixed too.

suborb:
@feilipu I've switched the preprocessor for sdcc over to ucpp - it works for the "simple" code we've got in the repo, but it might fail on more complex ones - can you test against some of your libraries etc?

feilipu:
> can you test against some of your libraries etc?

So FatFS passes with no issues. Perhaps not a surprise since it is written to comply with C89.

But FreeRTOS generates issues with inline assembly. Specifically in [portmacro.h](https://github.com/feilipu/z88dk-libraries/blob/master/freertos/source/include/sdcc/portmacro.h) the syntax for zsdcc doesn't seem to work properly any more. I've tried some variants (like adding `"/n"`), but still haven't gotten to the bottom of the correct syntax required.

suborb:
I didn’t want to add \x87 handling back then but to unstick us I’ll add it in.

feilipu:
> Actually, there's a common form (works for both sccz80 + sdcc) that doesn't need any weird characters.

Ah. Ok will try that.

I think I was quite close to that form at one stage, just had an extra `\n` on the `__asm__` line which was messing things up.

feilipu:
Related if there's to be just one preprocessor `ucpp`, then for newlib perhaps there doesn't need to be two separate versions of the header files? The header files could be made with decorations for both compilers, and then unnecessary inputs (e.g. `__LIB__`,`__smallc`, and ` __preserves_regs()`) just filtered out or decorations added as needed.

feilipu:
> can you test against some of your libraries etc?

With the correct in-line assembly syntax, FreeRTOS now passes fine. it complies to its [specification of C99](https://www.freertos.org/FreeRTOS-Coding-Standard-and-Style-Guide.html).

The time library also works fine, and it [aspires to C90 standard](https://github.com/feilipu/z88dk-libraries/tree/master/time).

So, from my point of view `ucpp` will be more than adequate to replace `sdcpp` if it remains troublesome.
And, it will probably be better to have a single preprocessor which will allow for simplification in the future.

Perhaps the documentation just needs to clarify that the C standard for z88dk is C99 (rather than the [sdcc documentation](https://sourceforge.net/p/sdcc/wiki/Standard%20Compliance/) which allows various standards?

Perhaps the default incantation for testing or using `zsdcc` should always include `--std-c99`?

suborb:
> Related if there's to be just one preprocessor `ucpp`, then for newlib perhaps there doesn't need to be two separate versions of the header files? The header files could be made with decorations for both compilers, and then unnecessary inputs (e.g. `__LIB__`,`__smallc`, and ` __preserves_regs()`) just filtered out or decorations added as needed.

The entry points are slightly different for sdcc and sccz80 around `__z88dk_fastcall` so they'd need to be a bit of work there to harmonise the headers.

suborb:
> Perhaps the documentation just needs to clarify that the C standard for z88dk is C99 (rather than the [sdcc documentation](https://sourceforge.net/p/sdcc/wiki/Standard%20Compliance/) which allows various standards?
> 
> Perhaps the default incantation for testing or using `zsdcc` should always include `--std-c99`?

I'm not sure the pre-processor needs any changes for higher standards beyond having a `__STDC...` macro changed?

suborb:
@feilipu thanks for cleaning up...

Should I build some binaries or should we wait for the formal release?

feilipu:
Do you think anyone would use Windows or Mac binaries to test? If so then you could build them to see.

Otherwise, I think this PR just stays as draft, pending v4.3.0, until we can finalise and merge.

If we're bored, it might be fun to play with some RCs when they come, but that's probably unnecessary.

suborb:
Given how much I dislike building the binaries, let's just leave this as draft..

feilipu:
Watching SDCC [#3491](https://sourceforge.net/p/sdcc/bugs/3491/) and [#3555](https://sourceforge.net/p/sdcc/bugs/3555/).

Dark-Schneider-666:
Hi, I am getting this issue with SDCC 4.2.0:
[https://github.com/z88dk/z88dk/issues/2216](url)
That is SDCC is ignoring __smallc calling convention, it always uses its own.

Is there some pre-release binary for Mac with Apple Silicon for testing?

feilipu:
The best thing to do is to raise the issue on the SDCC Bug Tracker. That way Philipp and others will be able to fix it cleanly.

spth:
> I'm not sure the pre-processor needs any changes for higher standards beyond having a `__STDC...` macro changed?

There were changes in the preprocessor C90 vs. C99, in particular the // comment.

AFAIK, there were no big preprocessor changes C99 vs. C17. This allowed SDCC to keep its old preprocessor for a long time.

But in C23 vs. C17, there are big changes in the preprocessor.

So a year ago, SDCC developers looked at available options, and in the end went for a preprocessor based on current GCC. The integration of cpp into the rest of GCC had become tighter, which made things harder for SDCC, but it still seemed easier than using the clang preprocessor, which is even more tightly integrated into LLVM.
If we were to make the decision today, we likely would go for a Wave-based preprocessor (and we are actually considering switching to that for 4.4.0 or 4.5.0), since Wave has nearly everything we need, and is meant to be suitable for a standalone preprocessor. But when we made the choice a year ago, Wave development had been inactive for years, and there was no effort to support C23.

Dark-Schneider-666:
> The best thing to do is to raise the issue on the SDCC Bug Tracker. That way Philipp and others will be able to fix it cleanly.

Hi, sorry I am a bit lost, where would be that? Have seen one at sourceforge but not sure if it's there, as I think the z88dk version is a modified one, so not sure if the vanilla SDCC consider the z88dk calling conventions so the report could not be at the right place.

feilipu:
> > The best thing to do is to raise the issue on the SDCC Bug Tracker. That way Philipp and others will be able to fix it cleanly.

> Hi, sorry I am a bit lost, where would be that?

[Here for SDCC bugs](https://sourceforge.net/p/sdcc/bugs/).

> Have seen one at sourceforge but not sure if it's there, as I think the z88dk version is a modified one, so not sure if the vanilla SDCC consider the z88dk calling conventions so the report could not be at the right place.

It is only modified slightly, to allow it to works for us, so it is very close to original these days. But we only move forward on a stable build, i.e. 4.x.0 releases, so check it hasn't already been resolved.

So it is best if you can reproduce with the vanilla build too.

feilipu:
Will do updates and move to merge released 4.3.0 before end of month. Been (am) AFK for some time.

feilipu:
@suborb this is now ready to go, and can be merged.
If you like you can re-run the tests before merging, once the [sdcc-r14210](http://nightly.z88dk.org/zsdcc/zsdcc_r14210_src.tar.gz) file exists.

suborb:
Had the usual pain building for MacOS, the solution is here: https://z88dk.org/forum/viewtopic.php?p=21448#p21448

Will merge when this PR completes.

PR #2197, opened 2023-02-21
fix zx_pxy2saddr prototype and macro declaration in spectrum.h

jorgegv:
fix: remove spurious *mask parameter from zx_pxy2saddr prototype and macro in spectrum.h.

The following code does not compile with `master`:

```
// zcc +zx test1.c -lndos -o test1 -create-app

#include <stdio.h>
#include <spectrum.h>

void main( void ) {
    printf( "Hello world\n\n\n" );
    printf( "pxy2saddr( 128,64 ) = %d\n", zx_pxy2saddr( 128, 64 ) );
}
```
The prototype and macro declarations for `zx_pxy2saddr` are wrong in `spectrum.h`

I have checked the ASM implementations, and indeed they do _not_ expect the `*mask` parameter


suborb:
I think the old version did care, but you're right the code we're using now doesn't.

PR #2198, opened 2023-02-22
(rules) I've been sitting on these for a while

suborb:
Various tweaks spotted.

The big change is not using add hl,NNNN for Z80n


lib/z80rules.1
```diff
@@ -6889,11 +6889,212 @@ $notcpu 8085
...
+=
+	ld	a,(_%1)
+	pop	de
+	ld	(de),a
+	ld	l,a
+
+%notcpu gbz80
+	push	hl
+	ld	a,%1
+	pop	de
```
ped7g, line 6952: (curious without knowing anything about rules in z88dk):
this ends with both HL and DE equal to some value, while `ex de,hl` will not keep HL... does the compiler never emit code which would exploit copy of HL in DE, so the optimization is safe?

lib/z80rules.1
```diff
@@ -6889,11 +6889,212 @@ $notcpu 8085
...
+=
+	ld	a,(hl)
+	ld	hl,%1	;const
+	add	hl,sp
+	add	(hl)
+	ld	l,a
+	ld	a,h
+	adc	h
+	ld	h,a
```
ped7g, line 6974: this is different? the H is not zero anymore (after `add hl,sp`).
imho this should be rather something like
```
 ld l,a
 ld h,0
 rl h
```
which is not clear win... also original code could do `ld h,d` instead of `ld h,0` to save 3T and one byte.
And finally not sure what the initial `%check` is for, as value in H is not preserved after add hl,sp, so it could be anything IMHO?

lib/z80rules.1
```diff
@@ -6889,11 +6889,212 @@ $notcpu 8085
...
+	pop	de
+=
+	ld	hl,%1	;const
+	add	hl,sp
+	add	(hl)
+	ld	l,a
+	ld	a,h
+	adc	h
+	ld	h,a
+	pop	de
```
ped7g, line 6994: same as previous, seems to me incorrect due to H != 0

src/sccz80/codegen.c
```diff
@@ -5192,14 +5192,9 @@ void constbc(int32_t val)
...
-        ot("add\thl,");
-        outdec(val); nl();
-    } else {
-        ot("ld\tbc,");
-        outdec(val);
-        outstr("\n\tadd\thl,bc\n");
-    }
+    ot("ld\tbc,");
+    outdec(val);
+    outstr("\n\tadd\thl,bc\n");
```
ped7g, line 5197: and why not using `add hl,nnnn` with z80n? I mean, it feels correct change to me, because the function has name `addbchl`, but otherwise `add hl,nnnn` is more optimal on Z80N if resulting BC is not important.

ped7g (commented):
left few comments at particular rules lines, but keep in mind I have zero knowledge about how z88dk and rules files work, and what they are expected to achieve. It's clearly not meant as general asm optimizer rules, as in such case too many of them would break code as they produce different results, so take my comments as curious by stander asking dumb questions... :) I hope you don't mind.

lib/z80rules.1
suborb: A push hl/pop de cycle occurs when storing a value to the stack.

An example of where this comes in is, something simple like this:

```
unsigned char b;

int func()
{
        unsigned char c;

        c = b;
}
```

Which is obviously a noop if you've got a context sensitive optimiser...

At -O0 it's ridiculously verbose:

```
// Make space on stack
        dec     sp
// Calculate &a
        ld      hl,0    ;const
        add     hl,sp
// Save it the we store it
        push    hl
// Load b 
        ld      hl,_b
        ld      l,(hl)
        ld      h,0
        ld      h,0
// Get address of a back
        pop     de
// And store it
        ld      a,l
        ld      (de),a
        inc     sp
        ret
```

Which will now end up as:

```
        dec     sp
        ld      hl,0    ;const
        add     hl,sp
        ex      de,hl
        ld      a,(_b)
        ld      (de),a
// Optimiser is context less, so assumes that the value is needed, so loads into primary reg
        ld      l,a
        ld      h,0
        inc     sp
        ret
```

Which you could take a little further:

```
         ld      a,(_b)
        push    af
        inc     sp
        ld      l,a
        ld      h,0
        inc     sp
        ret
```

lib/z80rules.1
suborb: It is wrong, thanks for the spot - all of my test cases obviously just cared about the lower 8 which is why they worked!

src/sccz80/codegen.c
suborb: I found it was interfering with some of the other rules.

The plan was to do a replacement in z80n_rules.1 but that needs some thought since a long + constant addition usually looks something like this:

```
ld bc,nnnn
add hl,bc
ex de,hl
ld bc,nnnn
adc hl,bc
ex de, hl
```
 
Which needs the carry flag from the `add hl,bc`

suborb:
> left few comments at particular rules lines, but keep in mind I have zero knowledge about how z88dk and rules files work, and what they are expected to achieve. It's clearly not meant as general asm optimizer rules, as in such case too many of them would break code as they produce different results, so take my comments as curious by stander asking dumb questions... :) I hope you don't mind.

 Not at all, it's appreciated. This stuff is quite delicate and a fresh eye asking questions is always welcome.

src/sccz80/codegen.c
ped7g: then there's no replacement, `add hl,nnnn` preserves carry flag and any checks for overflow will be more costly than regular Z80 `add hl,r16`.

(btw the other z80n extension `add hl,a` destroys carry flag with meaningless value - probably result of typo in VHDL, unfinished real carry reserving 17 bits and writing carry from b16, but letting adder to set only 16 of them, so the b16 is "whatever" fpga has there in the temp).

ped7g (approved):

desertkun:
Haha I've opened pandora's box haven't I? 😄

PR #2205, opened 2023-03-11
Add a manually selected failback to boost::filesystem

suborb:
macOS (Mojave? and earlier) don't have std::filesystem in any form and:

1. libstdc++ is no longer supported
2. libc++ can't be statically linked
3. I think boost::filesystem is compatible for what we use it for

So producing nightly builds with this flag for macOS will gain us a bit more backwards compatibility to support older hardware


suborb:
Found a difference: boost throws an exception if a wildcard doesn't match

suborb:
Ageing out

PR #2207, opened 2023-03-12
(arkos) Rearrange slightly to remove duplication

suborb:
Make the SMC player work for ROM targets by default (providing an opt-in option to use the ROM player)

I'll be adding CPC eventually...


jorgegv:
Whoa, dude, hats-off to you! I have been reviewing your rearrangement, it makes full sense... and you added support for 2 more targets in a glimpse!

Great team work :-)

PR #2213, opened 2023-03-19
(tms9918) Place __tms9918_map_colour into the correct code section

brijohn:
The  __tms9918_map_colour function was not being placed in the correct section during compilation. This makes sure it is placed into the code_clib section like the other tms9918 functions.


suborb:
I see what I did there! Thank you for the fix

PR #2237, opened 2023-05-01
newlib: Introduce sysvars.h for target "zx"

jurajlutter:
Introduce sysvars for zx as it is for zxn.

system_variables.asm has content from zxn, is to be reviewed and adjusted accordingly.


suborb:
Many thanks!

PR #2253, opened 2023-05-08
(classic) Stop moving files about the place

suborb:
z80asm can dump into the output directory, so use that feature.

This means another blocker on the way to parallel builds is sorted.

Additionally, rename ti8x to ti83p to be consistent.


SimonEast:
I just landed here after wondering why `zcc +ti8x` command was no longer working. I've seen this command recommended in several forum threads and also here in Github discussions, but it no longer works.

I'm all for clean-up and consistency, but I wonder if there could be some kind of hint given to the user when they try to run `zcc +ti8x`, especially considering the mentions of it across the internet. Currently you get this cryptic message with no suggestion of what to try:

```
Can't open config file ..\z88dk\bin/../lib/config//ti8x.cfg
```

I assume the command is now `zcc +ti83p`?

suborb:
I've added an internal redirect to cope with this in this change https://github.com/z88dk/z88dk/commit/c88a00048353c254bbbff534024abd6e4f4e930c

I hadn't realised it was so widely documented to be honest!

Issue #2258, opened 2023-05-12
(math32/am9511) Callee and other issues


suborb:
I'm a little confused as to what's happened with math32 - as far as I can tell bits of the library can't work with sdcc/classic and the same functions also cause trouble with linking.

```
#include <math.h>
#include <stdio.h>

#pragma printf "%f"

int main()
{
        double y;
        double x = fmod(11, 2);
        printf("%lf\n",x);
}
```

Fails to link (see https://z88dk.org/forum/viewtopic.php?f=2&t=11786)

Correcting, so that the SDCC function is exported as defc _fmod_callee = fmod_callee then the result is wrong.

A bit of head scratching shows the header file:

```
#ifdef __SCCZ80
extern double_t __LIB__ fmod(double_t x,double_t y) __smallc;
extern double_t __LIB__ fmod_callee(double_t x,double_t y) __smallc __z88dk_callee;
#define fmod(a,b) fmod_callee(a,b)
#endif

#ifdef __SDCC
extern double_t fmod(double_t x,double_t y);
extern double_t fmod_callee(double_t x,double_t y) __z88dk_callee;
#define fmod(a,b) fmod_callee(a,b)
#endif
```

So the calling conventions differ which isn't how classic works.

```
find lm32/c/sccz80 | xargs grep sdcc
```

Shows a lot of references to sdcc functions which isn't what I was expecting.

So I'm really confused and I've missed something, @feilipu can you remember and enlighten me?

feilipu:
IIRC, the main issue is that most of the library is written in assembly, and but some functions are written in C [and pre-compiled by SDCC_IX](https://github.com/z88dk/z88dk/blob/master/libsrc/_DEVELOPMENT/math/float/math32/c/Makefile#L10) and linked as assembly. So sometimes there are native labels, and sometimes the linkage has to deal with SDCC produced labels.

So that resulted in quite a confusing mix of native z80 assembly and SDCC labels being used, depending on where the function originated. And that is why you are seeing SDCC label references.

Compounding that, my testing was being done mainly on newlib targets (`sdcc_ix`, `sdcc_iy`, and sccz80 `new`) means that classic+SDCC combined only got `test` target testing, or possibly not tested at all.

Unfortunately this issue (classic+SDCC) might also flow onto `am9511` and `float16` libraries, as I would have done testing the same way there too. Can dig into that later.

---

[...]

PR #2265, opened 2023-05-17
math32 am9511 - fmin/fmax and compare improvement

feilipu:
This PR is to sort out the callee and comparison functions for `math32` and `am9511` math libraries.
It also introduces a C implementation of `fmin` and `fmax` for the 8085 build.

At the moment the 8085 specific sccz80 `_am9511_fmin` and `_am9511_fmax` assembled C functions are not connected to the library functions. I'm not sure how best to do this cleanly, without introducing another special case.

Perhaps adding another set of 8085 specific binding files is needed?
But, that seems like overkill. Welcome consideration on how to resolve neatly.

Further resolution of #2258

Restored `am9511` callee usage of `pow`, `hypot`, `frexp`, and `ldexp`, as these have been properly (??) implemented in newlib, so should work for classic too.

Removed `am9511` callee usage of `atan2`, `fmod`, and `modf` as these don't work, and won't be fixed.


suborb:
I'm wondering if we do something like this:

```
float am9511_fmin_callee(float x,float y) __z88dk_callee
{
    if ( x < y )
        return x;
    return y;
}

float am9511_fmin(float x,float y)
{
   return am9511_fmin_callee(x,y);
}
```

And then (cam32_sccz80_fmin.asm)

```
IF __CPU_8085
EXTERN _am9511_fmin
defc cam32_sccz80_fmin = _am9511_fmn
ELSE
; Rest of the code
ENDIF
```

feilipu:
Got caught by compare mutating the parameters... :shrug:
So needed this.
```C
float am9511_fmin_callee(float x,float y) __z88dk_callee
{
    float i = x;
    float j = y;
    if ( i < j )
        return i;
    return j;
}
```
I don't know of a cleaner way to preserve the parameters... perhaps with `volatile` ?
Please adjust if you prefer.

Anyway it passes the #2267 extended test, now.

PR #2276, opened 2023-05-26
(classic) Moving from recursive make

suborb:
Starting the move away from a recursive make for classic.

This doesn't actually allow a -j build yet:

- [ ] msxdeps target needs to be removed (+svi and +msx)
- [ ] x81deps target needs to be removed (+zx80, +zx81, lambda)
- [ ] stdio not migrated
- [ ] Few other sub libraries

But, on the +side we get dependency tracking:

- [x] Change a file in classic and the z80_crt0.lib files get rebuilt
- [x] Change a file in newlib and the crt0.lib files get rebuilt
- [x] Each sub-library can be built by itself either from the subdirectory or top level

So that's already an improvement. This is already pretty huge, so get in this partial stage sooner rather than later.


suborb:
Squashing to avoid the messiness

PR #2290, opened 2023-06-04
build.sh: Use gmake on Darwin

jurajlutter:
On Darwin, /usr/bin/make is GNU make 3.x that can not build z88dk reliably.

I'm suggesting here to use gmake 4.x from homebrew (or other 3rd party package manager).

### DIFF
--- a/.github/workflows/build-on-macos.yml
+++ b/.github/workflows/build-on-macos.yml
@@ -18,7 +18,7 @@ jobs:
         submodules: 'true'
     - name: Install homebrew dependencies
       run: |
-        brew install perl dos2unix boost ragel re2c
+        brew install perl dos2unix boost ragel re2c make
     - name: Install perl prerequisites
       run: |
         export PATH=/usr/local/bin:$PATH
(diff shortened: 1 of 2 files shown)


jurajlutter:
Also for CI, GitHub Actions might need to be updated to also install gmake.

suborb:
What’s the issue? I’ve not spotted anything unexpected and my main machine is macOS.   Having said that I probably do things slightly differently.

jurajlutter:
> What’s the issue? I’ve not spotted anything unexpected and my main machine is macOS. Having said that I probably do things slightly differently.

I've had various issues buiding using built-in make from macOS. Like, it failed to properly expand dependencies list or to rebuild the dependencies. Also, test suite was failing in a random way. There were no issues when using gmake 4.x from homebrew. gmake 3.81 from xcode cmdline tools is just too old.

dbolli:
Compiling nightly build now fails if gmake is not installed :(

suborb:
Hopefully the above keeps everyone happy - prefer gmake and recommend its use.

PR #2326, opened 2023-06-23
r2ka - integer math routines

feilipu:
Some integer math routines for the `r2ka` architecture.

Abstracted the multiply linkages so I didn't have to build a complete set of routines (and it now echos the divide linkages).

I've no hardware to test these additions effectively.
So perhaps @suborb you could extend the tests to make sure these routines are properly tested, please?
And of course any library or make file tricks that I've missed.

__EDIT__ Extended the existing testing a little to cover `r2ka`, but not sure if 100% right.


feilipu:
Will merge this now.
- [ ] to do a `r2ka` version of `l_mulu_64_32x32`.

PR #2334, opened 2023-06-29
Fix build for FreeBSD

W8BSD:
Use $(MAKE) instead of make
Don't use -q with dos2unix (Not supported on FreeBSD 12 and 13)
Use a default CC of cc on *BSD instead of gcc


pauloscustodio:
Thanks for contributing.

PR #2348, opened 2023-07-07
TRS80 Model II: hiding cursor, reducing the game control opts

zx70:


suborb:
Probably need:

```
        $(MAKE) -C video/mc6845 TARGET=trs80m2
```

after adding the port settings to the .inc file

PR #2367, opened 2023-07-22
[laser500] fix joystick fire2 button stuck high

nippur72:
on laser500 the joystick "fire2" button is bit 4 not 5 of I/O port $27. Currently the fire2 is stuck high when calling the `joystick()` function.


suborb:
Oops. Thanks for this.

Issue #2368, opened 2023-07-24
Build fails on Ubuntu WSL with garbage path


atirut-w:
I did everything exactly as the instructions said, but the makefile seems to contain an error. I'd like to help investigate this, but that would be tomorrow evening. I will leave information here if someone also wants to take a crack at it.

OS: Ubuntu 22.04 WSL

Error sample:
```
mkdir -p -- /usr/local/home/atirut/Projects/z88dk/src/sdcc-build/sdcc-install /usr/local/home/atirut/Projects/z88dk/src/sdcc-build/sdcc-install
mkdir: cannot create directory ‘/usr/local/home’: Permission denied
mkdir: cannot create directory ‘/usr/local/home’: Permission denied
make[3]: *** [Makefile:2800: installdirs] Error 1
make[3]: Leaving directory '/home/atirut/Projects/z88dk/src/sdcc-build/support/sdbinutils'
make[2]: *** [Makefile:2537: install] Error 2
make[2]: Leaving directory '/home/atirut/Projects/z88dk/src/sdcc-build/support/sdbinutils'
make[2]: Entering directory '/home/atirut/Projects/z88dk/src/sdcc-build/doc'
install -d /usr/local/home/atirut/Projects/z88dk/src/sdcc-build/sdcc-install/share/doc/sdcc
install: cannot create directory ‘/usr/local/home’: Permission denied
```

**Info for git bisect**:
Bad commit: bd839f7c659abf9cc2bc868dde20d1e62869a70e
Last known good commit: a09ee3d7a100514c7a1d7c56a1e20955386ab0b7

feilipu:
The WSL probably has an issue with the `usr/local/home` path.

Please confirm that you can create that path on your build machine.

If not, then it is clearly WSL related, given the new sdcc v4.3.0 build requirements, and you might need to adjust.

suborb:
I‘m AFK at the moment, but we’ve got the same failure with snapcraft. 

I‘ll dig a bit when I’m back later on this week.

atirut-w:
> The WSL probably has an issue with creating the `/usr/local/home` path.

I can investigate later today but the build script should *not* be creating anything inside a protected path AFAIK. It seems the build script appended `/usr/local` to a sane path, as evidenced by the username. The compile innstructions does not call for sudo or root access neither.

suborb:
Reflecting a bit more - GitHub actions passing, snap and your build failing during an install phase. 

[...]

PR #2370, opened 2023-07-27
zsdcc - simplify (omit) install

feilipu:
Since the `make install` generates a fake binary, which just refers to the real binary found in `sdcc/src`, why not just copy the real binary to `z88dk-zsdcc` and omit the install step? Well, let's see if that works...


feilipu:
Passes tests.
Don't know whether it will also pass WSL Ubuntu #2368 or nightly's, but it should.

Issue #2410, opened 2023-10-10
Could somebody help me with compiling for the ti84+


HeronErin:
Hello,

Currently I am writing [TiConstructor](https://github.com/HeronErin/TiConstructor), a project for making programs/apps for the ti-84+ (non ce), and have been using the normal sdcc for compiling with my own crt0 files. But have been noticing that sdcc generates some pretty bad assembly, has some issues with syntax (ie struct->attr+1 not returning the incremented attribute, but instead the normal value), and overall a quite bad standered library.  So am working on transitioning to using zcc as the compiler.

However have not been able to get the calculator to accept anything generated by zcc, and at best giving me a "ERR: INVALID".

The following is a bare bones helloworld I am using for testing that should compile for both zcc and sdcc, and produces a valid 8xp with my old build script:
 ```c
#ifdef SDCC
#define bcall(__LABEL__) rst	40 \
                        .dw __LABEL__
// A bit of trash data so the program runs correctly
void _(){__asm ret
	     .ascii "0000000000" __endasm;}

#else
#pragma string name ad
#define bcall(__LABEL__) rst 40 \ defw __LABEL__

// A bit of trash data so the program runs correctly
void _(){
	#asm
	ret
	defm "0000000000"
	#endasm
}
#endif




void main()
{

	#ifdef SDCC
	__asm
	#else
	#asm
	#endif
		bcall(0x4540 ) ; _ClrLCDFull

		xor	a, a
		ld	(0x86D7), a
		ld	(0x86D8), a

		#ifdef SDCC
		ld a, #'!'
		#else
		ld a, '!'
		#endif

		push	ix
		bcall(0x455E)     ; _VPutMap
		pop	ix


		bcall(0x4972 )    ; _getkey

	#ifdef SDCC
	__endasm;
	#else
	#endasm
	#endif
	


}
```
And here is the build script that works to compile this for sdcc:
```bash
DIRECTORY=$(cd `dirname $0` && pwd)

cd x

export MAINC="../hello.c"
export OUT_NAME="testtex"


sdasz80 -p -g -o tios_crt0.rel $DIRECTORY/other_files/tios_crt0.s


sdcc -DSDCC --no-std-crt0 --code-loc 40347 --data-loc 0 --std-sdcc99 -mz80 --reserve-regs-iy -o $OUT_NAME.ihx tios_crt0.rel $MAINC


objcopy -Iihex -Obinary $OUT_NAME.ihx $OUT_NAME.bin
[...]

Issue #1622, opened 2020-10-29
Invalid .8xp file produced by ZCC


siraben:
Given the following C program and Makefile, building succeeds but when I try to run it on the TI-84+ calculator (emulated) under TI-OS, I get an `ERR: INVALID` file. Looks like the binary needs to have `$bb $6d` prefixed to it before the conversion to `.8xp` happens.

```c
#include <stdio.h>

int main()
{
  puts("hello");
  return 0;
}
```

```makefile
all:
    zcc +ti8x -startup=10 -create-app hello.c -o hello
```

zx70:
this is an old, never ending topic  😁
I tend to prefer to evolve appmake and keep all the possible output variants.  Could an option, say, named "-Cz--bb6d" suffice.?

siraben:
Sure, that would work, but I thought that appmake would generate the `.8xp` properly?

zx70:
with the ti83plus formats, we had claims about the right way to create the application since forever.  I think that depending on the transfer tools/emulators in use, those 2 bytes are necessary or superflous (perhaps addec by the tool itself?)..  if we change it,  someone else will complain. 
So,  I checked the code and such workaround is in place already,  and I forgot about that  :)

please try by adding:
-Cz--altfmt
to your build command line

zx70:
@siraben ..did it work?

empathicqubit:
If I try with 

```
zcc +ti8x -subtype=mirage -startup=2 -Cz--altfmt -Cz--comment=what -create-app world.c
```

It successfully transfers to Wabbit, but there is weird junk in the name, and it does not appear in Mirage. However, if I try the same command, but edit the crt0 to include the extra bytes: 

```
IF (startup=2)
	DEFINE MirageOS			;Used by greyscale interrupt etc.
	DEFINE NOT_DEFAULT_SHELL	;Else we would use Ion
	;org	$9D95
	org	$9D93
	defb         $BB,$6D
```

It seems to execute correctly in Mirage but still has the weird junk in the name. If I remove the switch with the above edit, Wabbit refuses to transfer the file again. If I use binpac8x with the above edit, that also works. If I use binpac8x without the edit, then it does not appear in Mirage.

empathicqubit:
This is a really cool tool btw. Is it possible to attach a debugger to an emulator using the gdb stuff? I was going to write my own debugger if there wasn't anything, but I'm still trying to figure out what the best tools are for that.

zx70:
[...]

PR #2414, opened 2023-10-12
Fix no stub("asm(") programs for the ti83p

HeronErin:
For some reason the crt0 file had commented out the lines that add the AsmPrgm token to the beginning of the binary. This is required for both the ti83p and ti84p or the calculator will refuse to run the program and display "ERR: INVALID". It seems that "-create-app" still results in a garbled name, however this makes the binary compatible with binpac8x.py. But it should be noted that "-create-app" was broken before this PR, and #1622 mentions that it was a problem back then as well.

Related issues: #2410 and likely #1622


HeronErin:
After viewing the source code for "-Cz--altfmt" I saw that it is supposed to do this, however it wasn't working for me. This PR is not large enough to get ti-83 compiling well enough

Issue #2423, opened 2023-10-17
ez80 - ld.lis opcodes


feilipu:
There is an option to load a 2 bytes into an ez80 24 bit register using the `ld.lis` opcode.
I don't think these have been implemented.

For reference, the `BCU`, `DEU`, `HLU` registers are loaded with 0x00, while the normal 16 bit registers are loaded with the following two bytes.

feilipu:
Hmm. Not sure now... in um077.pdf on Page 230 the opcode table says it is not possible.
Yet on Page 20 the below example is given.

```
LD.LIS HL, 3456h ;HL[23:0] ¨ {00h, 3456h}. .IS
                 ;directs eZ80 to fetch only 16-
                 ;bits of data. .L uses all 3 bytes
                 ;of HL register.
```
I don't have an Agon to test, I'm afraid.

TheByteAttic:
You should ask L Break Into Program (Dean), the firmware designer.

feilipu:
Dean @breakintoprogram do you know the answer to this question?

feilipu:
OK, so Dean has answered on X and it seems that `ld.lis rr,mm` opcodes are real.
https://twitter.com/BreakIntoProg/status/1714272497120493677

PR #2424, opened 2023-10-17
ez80 - use mnemonics for lp_ functions

feilipu:
Converted `lp_` functions to use z88dk-z80asm mnemonics for the ez80. No change in function.

Pending #2423 to use `ld.lis`, if possible.


libsrc/_DEVELOPMENT/l/sccz80/4-ez80/lp_gchar.asm
```diff
@@ -4,18 +4,16 @@
...
-    defb    $5b
-    ld      a,(hl)
-    ret
+    push.l    hl       ; push 3 bytes (HLU is undefined in Z80 MEMORY Mode)
+    ld.lil    hl,2     ; point to pushed HLU by loading 16 bit 0x0002 to HL, HLU is written to 0x00.
+    add.l     hl,sp    ; by adding the 24 bit SPL to 24 bit HL
+    ld.l      (hl),e   ; store the E register at the HLU location in stack
+    pop.l     hl       ; get a correct (far pointer) long address into 24 bit HL
+    ld.l      a,(hl)   ; get a byte from that location into A.
+    ret.l              ; return correctly from a long call
```
suborb, line 19: These functions are called from z80 side, so this should be a short ret?

feilipu: You're right if it can be guaranteed that it will only be called from a z80 short call.

But the guidelines in the datasheet suggest...

4. Any code block that may be called from either Z80 mode or ADL mode must be called with a suffix to save the calling code’s memory mode on the SPL stack.
5. Any routine that may be called from either mode must return with a suffixed `RET.L` instruction to restore the calling code’s memory mode from the SPL stack.

I don't know how these will be used, but if they're intrinsic couldn't the be called from code running in either mode?  :man_shrugging:
But doing so would break everything else, unless everything was assembled in ADL mode.

I'll change it when we find out about the `ld.lis rr,mm` correctness (or not).

PR #2425, opened 2023-10-17
Fixed double label

polluks:


suborb:
Thanks, that was a dumb copy/paste.

PR #2428, opened 2023-10-19
Increase concurrency when building libraries

suborb:
I _think_ I've tracked down the majority of the dependency/ordering issues when building classic.

The problems arise when building the same code for different configurations - ideally target/ code should be put into separate sub-directories, but that's a large task to tackle.

So, bump in concurrency for a couple of the CI builds - we get a nice speed bump and can enjoy any intermittent errors that come out of it.

### DIFF
--- a/.github/workflows/build-on-macos.yml
+++ b/.github/workflows/build-on-macos.yml
@@ -31,7 +31,7 @@ jobs:
     - name: Build libraries
       run: |
         export PATH=/usr/local/bin:$PATH
-        ./build.sh -v -b
+        MAKE_CONCURRENCY=-j3 ./build.sh -v -b
     - name: Build examples
       run: |
         export PATH=/usr/local/bin:$PATH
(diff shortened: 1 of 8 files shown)


suborb:
This odd, the PR built fine, but merged and we repeatedly get this error on msys:

https://github.com/z88dk/z88dk/actions/runs/6585852909/job/17897400245

I can see that z80asm has had -v added in the past - do you have suspicions @pauloscustodio ?

pauloscustodio:
I do not remember, and a quick search on the closed issues did not reveal anything relevant. You can try to remove the ```Z80ASM=-v``` and retest

suborb:
Looking further, this is failing at the appmake stage and without an error message.

So, digging further, I spotted some reading beyond the bounds of an allocated block, so maybe that's triggering the equivalent of a bus error. I've pushed a fix and we'll see.

suborb:
Obviously a bit of a heisenbug since the tixx commit before mine passed. I'll keep an eye on it.

PR #2433, opened 2023-10-27
Added support for ti83p/ti84p flash apps.

### LINKED ISSUE #2410: Could somebody help me with compiling for the ti84+

HeronErin:
I have added support to building both single-page and multi-page flash apps for the ti83p/ti84p. A lot of things had to be added, so feel free to look through and see if I have left any bugs, or better yet tell me a better way of doing it.  The crypto stuff for signing apps was mostly taken from [spasm-ng](https://github.com/alberthdev/spasm-ng) and the rest I wrote. 

It should be noted that at this point puts() is not functioning for flash apps, but fputc_cons() is, so something needs fixed in the standard library somewhere. 

With the recent changes to ram programs and this PR, I should be able to close #2410

You can compile the following hello world app with `zcc +ti8xk main2.c -o main.bin -create-app -Cz"-n test"`
```c
#include <stdio.h>

// Call the system routine that waits for a user to press a key
void pressAnyKey() __ti_bcall(0x4972);

int main(){
    fputc_cons('H');
    fputc_cons('e');
    fputc_cons('l');
    fputc_cons('l');
    fputc_cons('o');

    pressAnyKey();


    return 0;
}
```

-----------------------------------------------------------------------------------------

And here is an example of compiling multi-page apps, you compile them as multiple different binaries and combine them with appmake. 
Makefile:
```make
all: prep build.8xk clean


prep:
	touch ti_branch_table.inc

clean:
	-@rm page_1.o main.o *.map


main.o: main.c page_1.o
	zcc +ti8xk main.c -o main.o -gen-map-file
page_1.o : page_1.c
	zcc +ti8xk page_1.c -subtype=xpage -o page_1.o -gen-map-file


build.8xk: page_1.o main.o
	z88dk-appmake +ti8xk --combine-pages --output out.8xk --name multipage --other-pages main.o,page_1.o
```
main.c :
```c
#define page_0
#include "multipage.h"
#include <stdio.h>

// Add in the branch table create in multipage.h (it should be noted that since pragmas are ran before function decorators, this must be pre-existing)
#pragma data_from_file branch_table ti_branch_table.inc


void pressAnyKey() __ti_bcall(0x4972); 


int main(){
	fputc_cons('0'+mpage_add(5, 4));

	pressAnyKey();


} 
```
page_1.c :
```c
#define page_1
#include "multipage.h"

// Since bcalls add to the stack, __ti_multipage_callable must pop them off the stack inorder for the params to work, and then reset the stack before the return, all without changing any registers
char add(char x, char y) __ti_multipage_callable   {
    return x+y;
}
```
multipage.h
```c
#pragma once

#ifndef page_1
// Can't read a label map while that label map is being created

char mpage_add(char x, char y) __ti_branch(1, "_add", "page_1.map"){}

#endif

```

Also, appmake now has these build args `-lgmp -lssl -lcrypto` so The GNU Multiple Precision Library along with OpenSSL are now needed as build dependencies.


HeronErin:
Just having some issues fixing the github workflows, not great at multiplatfom compiler stuff.

HeronErin:
Ok, now it is just not building :( 
With I knew what I was doing here

suborb:
Sorry, I've been away for a few days - I've now approved building.

It looks like you need install libgmp in the GitHub workflow files - easiest is to focus on ubuntu initially.

lib/target/ti83p/classic/ti83papp_crt0.asm
```diff
@@ -0,0 +1,268 @@
...
+	defc	__CPU_CLOCK = 6000000
+        INCLUDE "crt/classic/crt_rules.inc"
+	
+
+
+; Header data
+	DEFINE ASM
+	DEFINE NOT_DEFAULT_SHELL
+
+	org $4000
```
suborb, line 38: Is `CRT_ORG_BSS` needed to place variables into an area of RAM?

src/sccz80/codegen.c
```diff
@@ -5610,7 +5627,84 @@ void gen_interrupt_leave(SYMBOL *func)
     nl();
 }
 
+void gen_ti_multipage_enter(void){
+        int from_bcall = getlabel();
+        int not_from_bcall = getlabel();
+
+        // ol("di");
+        ol("exx");
```
suborb, line 5635: I'm not sure how frequently this sequence is likely to be seen, but can it be pushed into a library routine (see the handling of `l_debug_push_frame`/`l_debug_pop_frame`) that way if it's used more than once the size is reduced and it becomes a bit more generic and the mechanism could be used for other targets.

suborb:
I‘m getting my head around the workflow. There might be a way to get appmake to create the table and vector the bcalls which is less intrusive.

src/appmake/Makefile
```diff
@@ -11,7 +11,7 @@ else
 endif
 
 CC			?= gcc
-CFLAGS		+= -Wall -std=gnu11 -MMD -pedantic -I../common
+CFLAGS		+= -Wall -std=gnu11 -MMD -pedantic -I../common -lgmp -lssl -lcrypto
```
suborb, line 14: These `-l` should be on LDFLAGS

suborb: And for macOS -lssl -lcrypto aren't required.

src/appmake/ti8xk.c
```diff
@@ -0,0 +1,547 @@
...
+ */ 
+
+
+#include <ctype.h>
+#include <stdlib.h>
+#include <string.h>
+#if defined(__APPLE__)
+#  define COMMON_DIGEST_FOR_OPENSSL
+#  include <CommonCrypto/CommonDigest.h>
+#  define SHA1 CC_SHA1
```
suborb, line 21: MD5 is used in this file, so this should be:

```
#define MD5 CC_MD5
```

src/appmake/Makefile
suborb: Given that we only use MD5, I'd be tempted to pull in the reference implementation as source code rather than have this extra dependency.

lib/target/ti83p/classic/ti83papp_crt0.asm
suborb: I think this might also be the reason why `puts()` is failing - the std* streams aren't being setup because they're in flash rather than ram

src/appmake/ti8xk.c
suborb: I'm impressed that you got this to work! I'm not keen on the number of custom annotations and sccz80 changes you had to make so it would be nice to reduce them and allow use of the other compilers - I just need to figure out how.

I understand the need to compile banks to bin files - it looks like the paging isn't entirely under user control so we can't do our normal trick of a common/home and a banked page (which does make things simpler and allows more flexible interbank calling).

I'm putting down my thoughts below, it won't work at the moment - I do need to figure out how to get the BCALL target exports working nicely.

## crt0 extension

Introduce a link time variable eg `CRT0_BCALL_TABLE_SIZE` which defines the amount of reserved space for bcalls. Default it to 10 or something like that.

Add the following for each page:

```
    SECTION BANK_1
    org 0x014000
    SECTION CODE_1
    SECTION RODATA_1
    SECTION DATA_1
    SECTION BSS_1
    SECTION BANK_1_END
```

## Compiling files

Use `#pragma bank NN` in the file to move the address to 0x00014000, 0x00024000 as appropriate. This matches up with the way banking works on other machines.

Generate the binary with the -g flag which generates a defc file so we know the address of functions (**problem**: this contains all symbols, not just the ones we want to export).

The theory is that we should be able to include that .defc file in later compilations so we get the address of functions.  **Note** this won't work at the moment. 

## Function decorators

* If we use the decorator `__banked` then function calls are generated with:
```
call banked_call
defw addr
defw bank
```

We can add a flag which switches this to a rst:

```
rst $28
defw addr
defw bank
```

I'll need to come up with a way to add a function epilogue/prologue call to sort out the stack.


## Appmake extensions

- Search the binary files for the rst opcode. Pick up the address that follows and check that it exists in any of the `.map` files. Insert into table. Check table hasn't overflowed. Fix up call site (note, this wastes 2 bytes).

src/sccz80/codegen.c
HeronErin: As this is something that would be used often, it would make sense to moved to a library routine. Although I doubt it would useful to other targets due to the use of the Floating Point Stack.

src/appmake/Makefile
HeronErin: I was thinking the same thing with MD5, as requiring the installing of openssl just for one target seems like a waste if it can be done another way.  With there was a good alternative with GMP though.

src/sccz80/codegen.c
suborb: That's fine - it can be located in a TI specific library with a generic name: `l_fixup_banked_prologue`etc

HeronErin:
Ok, I have now gotten the docker working along with the "build-on-ubuntu" and "build-on-macos" workflows.

HeronErin:
Have everything but msbuild working. Have not idea what to do with msbuild

suborb:
Think we might have to skip msbuild - I can't see a chocolatey package for gmp.

The msvc files aren't actually used to produce a build - I think Paulo just uses them for local development, so I would:

1. Remove ti8xk.c from the project files
2. Protect the change in appmake.h with `#ifndef _MSC_VER`

HeronErin:
I'm trying to offload as much as possible to appmake (so multiple recompiles are not needed as they should have all already happened at the time of appmake) but I'm not sure the best way to do the branch table. Is there a way I can do a "for" loop in the crt file? (to zero fill it until appmake replaces it) Or would it be better just to add the size to the ORG and insert the bytes later?


Also is there a better way to 'link' the pages? I would like to just copy the function signature to a .h file and it detect the function in another page, right now I have the compiler search the .map file for the functions label, but this means you must compile the pages in a specific order and you can't use any function in a future page file.  
I know I probably will have to make appmake search for `rst $28 `opcodes and replace it with the correct branch table index, but how can it get that function label to begin with? Is there support in z88dk for dynamically linked libraries, then I could just do that and treat it like normal banked functions (with `rst $28` instead of a `call banked_function`)

suborb:
> Also is there a better way to 'link' the pages? I would like to just copy the function signature to a .h file and it detect the function in another page, right now I have the compiler search the .map file for the functions label, but this means you must compile the pages in a specific order and you can't use any function in a future page file.
I know I probably will have to make appmake search for rst $28 opcodes and replace it with the correct branch table index, but how can it get that function label to begin with? Is there support in z88dk for dynamically linked libraries, then I could just do that and treat it like normal banked functions (with rst $28 instead of a call banked_function)

This is is the annoying bit that I can't quite figure out how to do. Normally with _banked we compile to object files and the linker takes care of binding the address in. 

The two problems are:

* when compiling to .bin file we lose the ability to forward call within banks.
* when compiling via .o files we can have forward calls, but the support routines are not in the banks so nothing works

So the dynamic idea (or rather an extra stage of linking handled by appmake), with a ".dynamic" file created alongside with the patch offsets using .bin file method seems like it might be a neat solution.

It will require linker changes, so paging @pauloscustodio for some ideas.

suborb:
> I'm trying to offload as much as possible to appmake (so multiple recompiles are not needed as they should have all already happened at the time of appmake) but I'm not sure the best way to do the branch table. Is there a way I can do a "for" loop in the crt file? (to zero fill it until appmake replaces it) Or would it be better just to add the size to the ORG and insert the bytes later?

This was what I meant by:  "Introduce a link time variable eg CRT0_BCALL_TABLE_SIZE which defines the amount of reserved space for bcalls. Default it to 10 or something like that."

So we reserve zero-ed space in the crt and appmake then populates it. The variable can then be configured with the option `-pragma-define:CRT0_BCALL_TABLE_SIZE=nn` on the link command line.

HeronErin:
Sorry, what I meant to ask is **how** do I reserve a variable amount of space. Something like
.repeat defb 0 CRT0_BCALL_TABLE_SIZE times

HeronErin:
> > I'm trying to offload as much as possible to appmake (so multiple recompiles are not needed as they should have all already happened at the time of appmake) but I'm not sure the best way to do the branch table. Is there a way I can do a "for" loop in the crt file? (to zero fill it until appmake replaces it) Or would it be better just to add the size to the ORG and insert the bytes later?
> 
> This was what I meant by: "Introduce a link time variable eg CRT0_BCALL_TABLE_SIZE which defines the amount of reserved space for bcalls. Default it to 10 or something like that."
> 
> So we reserve zero-ed space in the crt and appmake then populates it. The variable can then be configured with the option `-pragma-define:CRT0_BCALL_TABLE_SIZE=nn` on the link command line.



> > I'm trying to offload as much as possible to appmake (so multiple recompiles are not needed as they should have all already happened at the time of appmake) but I'm not sure the best way to do the branch table. Is there a way I can do a "for" loop in the crt file? (to zero fill it until appmake replaces it) Or would it be better just to add the size to the ORG and insert the bytes later?
> 
> This was what I meant by: "Introduce a link time variable eg CRT0_BCALL_TABLE_SIZE which defines the amount of reserved space for bcalls. Default it to 10 or something like that."
> 
> So we reserve zero-ed space in the crt and appmake then populates it. The variable can then be configured with the option `-pragma-define:CRT0_BCALL_TABLE_SIZE=nn` on the link command line.

Sorry, what I meant to ask is **how** do I reserve a variable amount of space. Something like
.repeat defb 0 CRT0_BCALL_TABLE_SIZE times

pauloscustodio:
> .repeat defb 0 CRT0_BCALL_TABLE_SIZE times

You can write ```defs CRT0_BCALL_TABLE_SIZE, 0``` to reserve N zero bytes.

pauloscustodio:
> This is is the annoying bit that I can't quite figure out how to do. Normally with _banked we compile to object files and the linker takes care of binding the address in.
> 
> The two problems are:
> 
>     when compiling to .bin file we lose the ability to forward call within banks.
>     when compiling via .o files we can have forward calls, but the support routines are not in the banks so nothing works

I don't understand what is the problem. When you link banked code with banks at 0x014000, 0x024000, ..., each of these banks generate a separate bin file. The main code at address 0x000000 generates the fixed memory that is always paged in. The call to the banked_call function (in the fixed memory) with the ```defq func_name``` argument takes care of setting up the appropriate bank and jumping to the address in the function address, and the linker takes care of patching in the correct addresses.

suborb:
> I don't understand what is the problem. When you link banked code with banks at 0x014000, 0x024000, ..., each of these banks generate a separate bin file. The main code at address 0x000000 generates the fixed memory that is always paged in. The call to the banked_call function (in the fixed memory) with the `defq func_name` argument takes care of setting up the appropriate bank and jumping to the address in the function address, and the linker takes care of patching in the correct addresses.

The banking for the TI is setup slightly differently - the main page is paged out and the call is vectored via a table in the application header (so an index into that table) so each bank has to be self contained and the linker patching can't work.

pauloscustodio:
@HeronErin , what is the banking mechanism of the TI calculators? You create a table of function pointers (that include bank and address) and refer to them by index after a ```rst $38``` instruction?

pauloscustodio:
> Is there a way I can do a "for" loop in the crt file?

You can have in the crt:
```
   jp after_table
.branch_table
   defs BRANCH_TABLE_SIZE*3, 0
.after_table
```
and invoke zcc with ```-DBRANCH_TABLE_SIZE=10```

src/sccz80/codegen.c
```diff
@@ -1160,15 +1161,28 @@ void gen_hl_call(Type *functype, int module, int address)
...
     }
-    ol("call\tbanked_call");
-    ot("defq\t"); outname(sym->name, dopref(sym)); nl();
-}
+    if ( c_banked_style == BANKED_STYLE_TICALC ) {
+        ol("rst\t$28");            // BCALL, meaning the system will handle paging for us
+
+        outfmt(".%s%x%s%s\n",    // This label can be exported as a .map for appmake
+            BANKED_SYMBOL_PREFIX, 
+            (functype->funcattrs.ticalc_banked_call_count)++, // All labels must be unique
```
suborb, line 1175: Just call getlabel() which gives you an incrementing count

src/sccz80/codegen.c
```diff
@@ -1491,6 +1505,10 @@ void gen_leave_function(Kind vartype, char type, int incritical)
 
     if (type)
         setcond(type);
+    if ((currfn->flags&BANKED)==BANKED && c_banked_style == BANKED_STYLE_TICALC){
+        ol("EXTERN l_fixup_banked_epilogue");
```
suborb, line 1509: These `EXTERN` can go into z88dk/lib/z80_crt0.hdr

src/zpragma/zpragma.c
```diff
@@ -191,6 +195,51 @@ void write_redirect(char *sname, char *value)
     fclose(fp);
 }
 
+void data_from_file(char *line){
```
suborb, line 198: Is this still needed?

src/appmake/appmake.h
```diff
@@ -525,6 +536,10 @@ struct {
       "DGOS, KCS variant for Sorcerer Exidy and Microbee, also WAV format",
       NULL,
       sorcerer_exec,    &sorcerer_options },
+    { "bin2var",   "ti8xk",       "(C) 2023 z88dk, HeronErin, Spasm-ng",
```
suborb, line 539: This bit needs to be in `#ifndef _MSC_VER`

lib/target/ti83p/classic/ti83papp_crt0.asm
```diff
@@ -0,0 +1,294 @@
+;	Stub for the TI 83+ calculator for building as an app
+;
+;  This is preferable to ti83papp.asm
```
suborb, line 3: `ti83papp.asm` just doesn't work anymore - it looks like it was written before we had sections. So feel free to delete or drop this one into its place.

suborb:
@HeronErin this is really nicely done - thank you.

Just a couple of extra cleanup things to think about:

- The old ti83app.asm doesn't work, so can your crt0 replace it?
- Once that is done, the extra config file can go away and using this becomes a SUBTYPE in ti83p.cfg
- And then rename the appmake target from +ti83pk to +ti83papp (and "executable" name to bin2app rather than bin2var)
- An example in examples/ would be great - it would be built on every commit so this doesn't rot over rtime

src/zpragma/zpragma.c
HeronErin: No it is not needed, pragma data_from_file was from before I moved the brany

HeronErin:
That was a difficult use after free bug to track down, but I finally think that this PR is close to being done. 

@zx70 It seems that [jsTIfied](https://www.cemetech.net/projects/jstified/) is still not excepting the 8xp files generated by z88dk. Here is the [Makefile](https://github.com/HeronErin/z88dk/blob/master/examples/ticalc/Makefile) and [C file](https://github.com/HeronErin/z88dk/blob/master/examples/ticalc/world.c) I am using in [examples/ticalc](https://github.com/HeronErin/z88dk/tree/master/examples/ticalc).

zx70:
> @zx70 It seems that [jsTIfied](https://www.cemetech.net/projects/jstified/) is still not excepting the 8xp files generated by z88dk.

Ok, I'll begin looking at it but isn't it better leaving to you the field of the TI calculators sections for a while ?  I wouldn't want to collide with your work.

zx70:
It look like it builds.
Do you think we can already pull it?
IMO if it's a WIP but in a reasonable decent state, it is a valid option.

HeronErin:
I think it is ready. I am going to do some testing to find any issues, but thinks seem to be working

suborb:
I've been AFK for a few days. 

@HeronErin thanks for doing the cleanup this is all looking good.

I'll merge when I've updated the nightly machine to have gmp for the various machines we build for.

suborb:
I've had a bit of fun creating a fat library for arm64/x86_64 macOS - but I've now got something that runs on my machines despite a few linker warnings.

PR #2454, opened 2023-12-05
Fix 8xp programs

HeronErin:
Fixed tixx.c in appmake so that now 8xp programs are accepted by both [jstified](https://www.cemetech.net/projects/jstified/) and my physical calculator. 

* I found that although spasm makes use of a `trailer83`, it is not actually something used for the TI-84, and is not used in other programs for building .8xp files. 
* The AsmProg token was placed at the wrong part of the file. It should be in the binary contents, not the header.
* Also made a change to allow for genname() to parse out the file name if the output dir is in another directory
* Corrected the length to be correct
* Also did some refactoring to make variable names less ambiguous

### DIFF
--- a/examples/ticalc/Makefile
+++ b/examples/ticalc/Makefile
@@ -3,7 +3,7 @@
 # Note: just building hello worlds
 
 
-all: world82.82p world83.83p world83.8xk world83.8xp world85.85s world86.86p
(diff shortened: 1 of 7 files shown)


HeronErin:
Fixes: #1622

suborb:
Apologies, I've been off doing other things the past few days.

Thanks for the update!

PR #2461, opened 2023-12-18
Nabu analog paddle support

brijohn:
This updates the NABU target to support receiving and decoding the analog data sent from the keyboard when using paddle controllers. I've also included an example program I used to test keyboard/controller input devices.

### DIFF
--- a/examples/nabu/inputtest.c
+++ b/examples/nabu/inputtest.c
@@ -1,0 +1,121 @@
+#include <conio.h>
+#include <games.h>
+#include <arch/nabu/paddles.h>
+#include <font/font.h>
+#include <sys/ioctl.h>
+
+#define MODE_TEXT 0
+
+// decode and print joystick byte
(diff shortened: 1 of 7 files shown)


examples/nabu/inputtest.c
```diff
@@ -0,0 +1,122 @@
...
+	if (b & MOVE_FIRE) {
+		state[4] = 'F';
+	}
+	printf("%s", state);
+}
+
+void main() {
+
+	uint8_t mode = MODE_TEXT;
+	void *font_ptr = &font_6x8_ascii;
```
suborb, line 43: Do you mind using either `font_8x8_einstein_system` or `font_8x8_msx_system`? Both of them should work correctly in text mode without being cut off.

I'll add a 6x8 category after the next release to avoid having to guess which ones should work.

suborb:
Thank you very much for this.

PR #2482, opened 2024-01-25
rcm2024 - initial headers

feilipu:
I'm adding headers for the rcm2024 (a r2ka CPU machine, with RC2014 bus).

I'm using the newlib mechanism to generate headers for private asm, public asm, and C, for both the r2ka CPU and for the optional rcm2024 bus interface modules

Not too comfortable putting the target related headers into the include/arch/r2ka directory. Probably there should be separate target related header configurations directory in classic (like newlib), where a target can pick its own header configuration files, and mash them together with the CPU related header configurations.


suborb:
I'm resigned to the fact that although they're called targets, the headers are in the arch/ folder - this was chosen to match newlib.

If the ppide etc headers aren't just for the r2ka then there's an argument to moving them into a common arch/bus/rcbus directory

feilipu:
> If the ppide etc headers aren't just for the r2ka then there's an argument to moving them into a common arch/bus/rcbus directory.

I've been mulling this argument (for too long now).

There are three cases.
1. The intrinsic CPU registers for z180, r2ka, r4k, etc. These will never change but will appear for every target, though for 8080/5 and z80 these are minimal.
2. Hardware or devices with registers or typical implementations like ppide or tms9918 etc. These will probably also be constant but may appear across multiple targets.
3. The target itself that needs to have a CPU frequency, how the I/O and memory is configured, etc. These will be target specific (by definition).

So I am uncomfortable to smash them all in the `include/arch/` directory, though that's where they all are now.
And, since there's a `config` directory needed for the m4 files to generate headers, it will need to hide somewhere too. Perhaps that means an `arch/device/` directory?

suborb:
@feilipu I know you've lost interest in the Rabbit, I'm guessing this should be closed since you're not going to take this any further?

feilipu:
Yes. Realising that the Rabbit can't successfully run CP/M or any standard Z80 code is a deal breaker for me.

And since I won't be able to do anything sensible until end 2026, best to just close this. I'll do it now.

PR #2486, opened 2024-01-30
add missing SP1 fastcall stubs after source migration to classic

jorgegv:
Hey @suborb, I just copied the `_fastcall` stubs from `sdcc_ix`, since they are the most restrictive (IX is preserved in them) and I think I recall using this same criteria when adapting the newlib sources to classic the first time.

### DIFF
--- a/libsrc/_DEVELOPMENT/temp/sp1/zx/c/sccz80/sp1_DeleteSpr_fastcall.asm
+++ b/libsrc/_DEVELOPMENT/temp/sp1/zx/c/sccz80/sp1_DeleteSpr_fastcall.asm
@@ -1,0 +1,11 @@
+
+; void sp1_DeleteSpr_fastcall(struct sp1_ss *s)
+
+SECTION code_clib
+SECTION code_temp_sp1
+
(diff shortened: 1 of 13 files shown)


suborb:
Thanks for picking this up - do you want to add the rotation table export to this PR?

jorgegv:
> Thanks for picking this up - do you want to add the rotation table export to this PR?

Sure, that's why I added "WIP" to the title... :-)

Regarding that, I was thinking: why not just add the `customize.asm` SP1 definitions to the library instead of near the CRT?

jorgegv:
@suborb my head is exploding now... I was thinking of adding an include for `config_provate.inc` somewhere, I thought that `sp1_Initialize` would be the sensible place... and there is already an `INCLUDE "config_private.inc"` in `libsrc/_DEVELOPMENT/temp/sp1/zx/updater/asm_sp1_Initialize.asm`! This would explain why the examples work fine... but then why don't the symbols from config_private appear in the generated library?

suborb:
Thanks @jorgegv

PR #2494, opened 2024-02-08
Clang fixes for zxspectrum next header

vmorilla:
Fixes in zxn.h for CLANG compiler


suborb:
Thanks for the fix @vmorilla

PR #2519, opened 2024-04-20
z80 crt: define __SDCC_IX for correct __sdcc_enter_ix

jorgegv:
define __SDCC_IX for correct __sdcc_enter_ix when using --opt-code-size

### DIFF
--- a/libsrc/z80_crt0s/Makefile
+++ b/libsrc/z80_crt0s/Makefile
@@ -28,23 +28,23 @@ obj/gbz80-crt0:
 	@touch $@
 
 obj/z180-crt0:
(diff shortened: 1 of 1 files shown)


suborb:
For reference, it would need to be added for eZ80, Rabbits, KC160 and Z180.

However, I've no idea what's going on with that test failure so I'll dig.

suborb:
Oh, wow. It's l_compare_de_hl.asm which swaps the arguments round when compiled with one of the __SDCC flags.

I presume this fixed your problem? If so I'll sort this PR out.

Thanks for raising it in the first place!

suborb:
I'm going to merge into a temporary branch to save code dictation!

PR #2520, opened 2024-04-20
Continuation of #2519 - sdcc_enter_ix

suborb:
* Classic should be compiled with sdcc_ix
* crt0 routines weren't being
* As a result --opt-code-size pulled in a routine with ix/iy swapped


jorgegv:
The problem I mentioned does not happen any more (expected), now IX is used correctly.

PR #2522, opened 2024-04-21
(classic) Generalisation of __far support

suborb:
__far support hasn't really been used for 20+ years, so a few bugs crept in during the sccz80 updates.

* Create a general arch where only a small number of low level functions are needed for each target
* Generalise the far string routines and add a couple of mem* routines
* Rename functions so that they are `strcmpf` rather than `strcmp_far` (uncertain about this, makes eyeballing code difficult)
* Import the allocator from sdcc in - realloc is a much simpler version to avoid some of the benchmark issues we saw
* Temporary names for the low level routines (need a rename to represent their purpose rather than where they appear in the code)
* As a bonus add a cpm library for the z80retro! - just the VDP code, untested at present


suborb:
I still need to add machine initialisation code + examples for +msx and +zx, but I'll merge this bit now since the core work is done.

PR #2524, opened 2024-04-26
docker: publish a tagged image when a new tag is pushed (e.g. v2.2, v2.3...)

jorgegv:


jorgegv:
Shamelessly copied from the existing one, but adjusted for the trigger actions and the needed docker tag

suborb:
Thanks, nothing quite like having 3 copies of effectively the same file!

PR #2530, opened 2024-05-09
Correct extended DSK signature

damieng:
The disk signature for generated .dsk files is slightly incorrect causing the image to not be loadable into various tools such as Spectaculator and Disk Image Manager.

The spec at https://www.cpcwiki.eu/index.php/Format:DSK_disk_image_file_format indicates that offset 00 should include the sequence:

`"EXTENDED CPC DSK File\r\nDisk-Info\r\n"`

Unfortunately it is generated with the wrong case right now as the word File has been capitalized. Manually editing the signature in generated disk images indicated the image is otherwise correct.

### DIFF
--- a/src/appmake/cpmdisk.c
+++ b/src/appmake/cpmdisk.c
@@ -1,4 +1,4 @@
--/*
+/*
  * Disc image handling
  */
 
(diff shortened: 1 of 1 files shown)


suborb:
Good spot, thank you.

PR #2560, opened 2024-06-10
cpm - 8085 stdio fnctl library build

feilipu:
As quite a lot of the classic library stdio and fnctl code is still in C, it is worth doing a specific CP/M 8085 build to make use of the improvements in sccz80 and 8085 libraries (vs 8080).

### DIFF
--- a/lib/config/cpm.cfg
+++ b/lib/config/cpm.cfg
@@ -15,7 +15,7 @@ CLIB    default -lcpm_clib -D__Z80__ -D__Z80 -Cc-standard-escape-chars -LDESTDIR
 CLIB    ixiy -lcpmixiy_clib -D__Z80__ -D__Z80 -mz80_ixiy -startuplib=z80iy_crt0 -Ca-IXIY -Cl-IXIY -Cc-standard-escape-chars -LDESTDIR/lib/clibs/ixiy
 CLIB    z180 -lcpmz180_clib -startuplib=z180_crt0 -Cc-standard-escape-chars -LDESTDIR/lib/clibs/z180 -LDESTDIR/lib/clibs/z180 -mz180 -D__Z180
 CLIB    8080 -lcpm8080_clib -Cc-standard-escape-chars -m8080 -D__8080 -startuplib=8080_crt0 -pragma-define:CLIB_DISABLE_FGETS_CURSOR=1 -LDESTDIR/lib/clibs/8080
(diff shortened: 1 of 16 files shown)


libsrc/stdio/fread.c
```diff
@@ -4,7 +4,7 @@
 #define STDIO_ASM
 #include <stdio.h>
 
-#ifdef __8080
+#if CPU_8080 || CPU_8085
```
suborb, line 7: __8080 || __8085 is preferred I think

lib/config/cpm.cfg
```diff
@@ -15,7 +15,7 @@ CLIB    default -lcpm_clib -D__Z80__ -D__Z80 -Cc-standard-escape-chars -LDESTDIR
 CLIB    ixiy -lcpmixiy_clib -D__Z80__ -D__Z80 -mz80_ixiy -startuplib=z80iy_crt0 -Ca-IXIY -Cl-IXIY -Cc-standard-escape-chars -LDESTDIR/lib/clibs/ixiy
 CLIB    z180 -lcpmz180_clib -startuplib=z180_crt0 -Cc-standard-escape-chars -LDESTDIR/lib/clibs/z180 -LDESTDIR/lib/clibs/z180 -mz180 -D__Z180
 CLIB    8080 -lcpm8080_clib -Cc-standard-escape-chars -m8080 -D__8080 -startuplib=8080_crt0 -pragma-define:CLIB_DISABLE_FGETS_CURSOR=1 -LDESTDIR/lib/clibs/8080
-CLIB    8085 -lcpm8080_clib -Cc-standard-escape-chars -m8085 -D__8085 -startuplib=8085_crt0 -l8085_opt -pragma-define:CLIB_DISABLE_FGETS_CURSOR=1 -LDESTDIR/lib/clibs/8085
```
suborb, line 18: Can you update +test as well?

feilipu: I think the +test target `config/test.cfg` already has a 8085 specific build, but I've fixed a bit of formatting on the way past.
Have I missed what you intended?

libsrc/stdio/fread.c
feilipu: resolved with https://github.com/z88dk/z88dk/pull/2560/commits/4442b58a6164bfc4e250735af40ea4b625895b87

feilipu:
And done. Found the itchy `atol_fastcall` spot.

lib/config/cpm.cfg
suborb: Ah yeah, it's okay...just ignore me! :)

PR #2562, opened 2024-06-10
genmath_iy on KC and Primo renamed to genmath_zx81

zx70:


suborb:
Genmath_zx81 preserves af' as well as the ixiy swap - do those machines need it?

zx70:
> Genmath_zx81 preserves af' as well as the ixiy swap - do those machines need it?
> 
> Probably needs to be genmath_ixiy instead.

Ok, I'll double check.  The library was missing my local environment, I thought it had been renamed.

PR #2582, opened 2024-07-01
End address of Z1013 header must point to last the address  of binary

hechtus:
The end address of the so called Z1013 'Headersave' header must point to the last byte of the binary instead of one Byte beyond.

### DIFF
--- a/src/appmake/z1013.c
+++ b/src/appmake/z1013.c
@@ -99,7 +99,7 @@ int z1013_exec(char* target)
     }
 
     writeword(pos, fpout);
-    writeword(pos + len, fpout);
+    writeword(pos + len - 1, fpout);
     writeword(pos, fpout);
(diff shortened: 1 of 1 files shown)


suborb:
Thanks, for future reference:

```
   DATA:  13H Blockadresse des gelesenen Blocks
   ZILAD: 25H zu lesende Blockadresse
   SOIL:  16H Start of Input-Line
   ARG 1: 1BH Anfangsadresse File
   ARG 2: 1DH Endadresse File
   ARG 3: 23H Startadresse File
   DATA+2:15H Zeichenzähler bei Nameneingab
```

from: https://hc-ddr.hucki.net/wiki/doku.php/z1013/software/headersave

suborb:
Thanks for the contribution.

PR #2585, opened 2024-07-18
8085 - add saved_hl - remove saved_de

feilipu:
Found compiling for 8085 using `__z88dk_callee` the `saved_hl` and `saved_de` locations are used by sccz80, but are not provided in 8085 crt0 for CP/M and other targets.

Improvement - there could be a `IF __8080 |  __8085` protection around the locations in the CP/M crt0 to create these locations only for 8080 or 8085, but I've been unable to find the right solution to make it work.

Also, additional 8080 and 8085 targets should have this added to their crt0.

### DIFF
--- a/lib/crt/classic/crt_section.inc
+++ b/lib/crt/classic/crt_section.inc
@@ -117,7 +117,6 @@ _heap:
 ENDIF
 
 IF CLIB_BALLOC_TABLE_SIZE > 0
-
     ; create balloc table
     SECTION data_alloc_balloc
     PUBLIC  __balloc_array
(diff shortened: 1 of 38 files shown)


suborb:
I think it's `__CPU_INTEL__` that you need?

feilipu:
> I think it's __CPU_INTEL__ that you need?

Yes. Thanks.
Added the conditional now to both cpm and test libraries.

And added to all the targets which mention 8080 or 8085 in their configuration (without the conditional).

I'll merge later if no further improvements.

suborb:
Sorry, I'm going to mess you around.

* `saved_de` doesn't seem to be used at all so all references to it can be removed
* `saved_hl` for classic can be put into `crt_section.inc` in `bss_crt`

feilipu:
No problem.
I guess then it can be removed from the specific classic target crt0 code?

suborb:
> I guess then it can be removed from the specific classic target crt0 code?

Yes

PR #2594, opened 2024-09-18
Added generation of .aqx files for Aquarius+

iratahack:
Aquarius+ BASIC has a new feature, the capability to "run" a binary file directly from BASIC using the "run" command.

- e.g. run main.aqx

The .aqx file format is a header followed by the binary image. The header consists of the following.

| Name | Size | Value|
|--------|--------|--------|
|Signature| 10 bytes|AQPLUSEXEC|
|Load address| 2 bytes | Populated by appmake |
|Binary length| 2 bytes | Populated by appmake |
|Exec address| 2 bytes| Populated by appmake|

A new subtype has been added to the aquarius target 'aqx' which will pass the --aqx parameter to appmake. Also, appmake for the Aquarius now accepts --crt0file and --org parameters.

### DIFF
--- a/lib/config/aquarius.cfg
+++ b/lib/config/aquarius.cfg
@@ -17,6 +17,7 @@ CLIB		aqplus -laqplus -laquarius_clib -D__AQUARIUSPLUS__ -pragma-define:CLIB_AQU
 
 SUBTYPE         none
 SUBTYPE		default -Cz+aquarius 
+SUBTYPE		aqplus -Cz+aquarius -Cz--aqx
 SUBTYPE		banked -Cz+noop 
 SUBTYPE		rom -Cz+rom -Cz--romsize=8192 -pragma-define:CLIB_AQUARIUS_ROM=1 -Cz--rombase=0xe000 -D__ROM__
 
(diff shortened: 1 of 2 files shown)


lib/config/aquarius.cfg
```diff
@@ -17,6 +17,7 @@ CLIB		aqplus -laqplus -laquarius_clib -D__AQUARIUSPLUS__ -pragma-define:CLIB_AQU
 
 SUBTYPE         none
 SUBTYPE		default -Cz+aquarius 
+SUBTYPE		aqplus -Cz+aquarius -Cz--aqx
```
suborb, line 20: This ought to be aqx

suborb:
Have you got a pointer to the docs for this feature? I can't seem to find anything (albeit it was a  cursory search)

lib/config/aquarius.cfg
iratahack: Instead of aqplus?

suborb: Yes please!

iratahack:
> Have you got a pointer to the docs for this feature? I can't seem to find anything (albeit it was a cursory search)

@RevCurtisP litreally implemented it last night.

lib/config/aquarius.cfg
iratahack: Got it. No problem!

suborb:
Thanks for this - I should check in and catch up with what's happening in AQ+ land at some point soon.

iratahack:
> Thanks for this - I should check in and catch up with what's happening in AQ+ land at some point soon.

Lots of cool stuff going on. Including a port of Castle Escape ;) written using z88dk of course https://iratahack.itch.io/castle-escape-aquarius

PR #2598, opened 2024-09-19
Aquarius+ specific native console I/O routines.

iratahack:
These changes are needed to fix problems when calling printf() or gets() from __banked functions on the Aq+.

1. Created Aq+ specific routines for _fputc_cons_native_ and _fgetc_cons_native_ and modified them so calls to the ROM are performed via _ext_call_ which correctly handles saving and restoring the memory bank paged in at $c000. _ext_call_ does clobber IX so these routines now save and restore the IX register also.
2. The _libsrc_ _Makefile_ has been updated to include _-DSTANDARDESCAPECHARS_ when building the sources for the aqplus_clib library so CR/LF are handled correctly.
3. The _getk_ routine calls the _KEYASC_ ROM routine ($1f01, actually KEYASC+1) which is available in both Aq and Aq+ ROMs and does not need to be called via _ext_call_. However, the IX register is clobbered so _getk_ now saves and restores IX.

### DIFF
--- a/libsrc/_DEVELOPMENT/sdcc_opt.1
+++ b/libsrc/_DEVELOPMENT/sdcc_opt.1
@@ -606,6 +606,14 @@ __xinit_%0
 	call	banked_call
 	defq	%2
 
+	ld	e,%1
+	ld	hl,%2
+	jp  ___sdcc_bcall_ehl
+=
+	call	banked_call
+	defq	%2
+	ret
+
 ;; 
 ;; ez80 instructions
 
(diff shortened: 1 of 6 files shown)


libsrc/_DEVELOPMENT/sdcc_opt.1
iratahack, line 1: Handle banked call at the end of a function when using SDCC.

suborb:
Thank you once again - adding banked support and far heap was a lot of fun and I'm curious to see what people will do with it.

PR #2602, opened 2024-09-23
zx: Add “-clib=noclib” variant for cases when zx_clib library is not present

desertkun:
I have a rare setup when I build zx game without CLIB (hot-pluggable module). Modules in this case rely on main executable's clib, so when themselves are built, clib is not needed.

This change adds `-clib=noclib` variant so I can build w/o clib for zx when needed.


suborb:
I think I know how you’re using this - ie with an export file for linking. But most of the clib is actually within the z80 lib so all that’s being excluded here is stdio and Zx specific code - is that what you expect?

desertkun:
Currently when you would try to build a module with 
```
-nostdlib --no-crt
``` 
it would still try to link `zx_clib` which I don't need (and don't even have). So yeah it's a mishmash, got tired of keeping it in my stashes :)

Since code is "targets" zx, I need +zx so could would invoke zx's stdlib, but stdlib is not there — it's provided by an asm mixin with a thousand of defc statements.

desertkun:
@suborb do you have any objections?

PR #2603, opened 2024-09-24
Added aqx support to appmake for the Aquarius+ and other updates.

iratahack:
# Change 1

An aqx file is a BASIC program with a binary image appended.

In this case the BASIC program consists of:

- 10 X=USR($xxxx)

Where xxxx is the address of CRT_ORG_CODE. The binary code must have a load address >= $39B0, which the default is. The load address can also be changed with _#pragma output CRT_ORG_CODE=addr_. The space between the end of the BASIC program and the binary file is padded with zeros.

The aqx format differes from the aqex format in that it has a BASIC protion. aqex files are pure binary with a prepended header.

# Change 2

Separate the code used by _loadbanks_ to read the SD card via the ESP device so it can be called by the developer.

# Change 3

Allow all memory pages to be included in banking starting at USER_RAM (32). Memory pages 32 -> 63 are mapped to banks 0 -> 31 giving the user access to all RAM pages.

# Change 4

Enabled VTII, WYZ and Arkos sound examples for Aq+

# Change 5

Added IM1 interrupt support based of VBLANK IRQ.

# Change 6

Added support for joystick 1 & 2.


suborb:
> Allow all memory pages to be included in banking starting at USER_RAM (32). Memory pages 32 -> 63 are mapped to banks 0 -> 31 giving the user access to all RAM pages.

I've got some (non-committed) test programs to validate the mapping between `__banked` and `__far` - it uses the `l_far_mapaddr` routine which might need some tuning to cope with this change.

I've not got access to that machine until the weekend so I'll be able to validate then.

iratahack:
> > Allow all memory pages to be included in banking starting at USER_RAM (32). Memory pages 32 -> 63 are mapped to banks 0 -> 31 giving the user access to all RAM pages.
> 
> I've got some (non-committed) test programs to validate the mapping between `__banked` and `__far` - it uses the `l_far_mapaddr` routine which might need some tuning to cope with this change.
> 
> I've not got access to that machine until the weekend so I'll be able to validate then.

I think I have addressed l_far_mapaddr in the latest check-in and also the initialization of the far heap in setup_far_heap.

iratahack:
> > > Allow all memory pages to be included in banking starting at USER_RAM (32). Memory pages 32 -> 63 are mapped to banks 0 -> 31 giving the user access to all RAM pages.
> > 
> > 
> > I've got some (non-committed) test programs to validate the mapping between `__banked` and `__far` - it uses the `l_far_mapaddr` routine which might need some tuning to cope with this change.
> > I've not got access to that machine until the weekend so I'll be able to validate then.
> 
> I think I have addressed l_far_mapaddr in the latest check-in and also the initialization of the far heap in setup_far_heap.

I've removed most of the remapping of bank numbers. This makes the code cleaner IMO. The only time mapping is used is when loading banks from the xxx_BANK_nn.bin files when 32->36 are mapped to 0->31. All other code uses the actual aquarius RAM page numbers. Hopefully this didn't break any of your test cases.

libsrc/target/aquarius/interrupt/im1_init.asm
```diff
@@ -0,0 +1,64 @@
...
+        ld      (INTJMP), a
+        ld      hl, asm_im1_handler
+        ld      (INTJMP+1), hl
+
+        ; Unmask and clear VBLANK IRQ
+        ld      a, $01
+        out     (IO_IRQMASK), a
+        out     (IO_IRQSTAT), a
+
+        ei
```
suborb, line 37: Should there not be a di whilst we're messing with the vectors?

To preserve the previous interrupt state, this is the pattern we use:

```
call l_push_di. ; saves enabled state
.... ; critical code
call l_pop_ei ; restores enabled state
```

libsrc/target/aquarius/interrupt/im1_init.asm
```diff
@@ -0,0 +1,64 @@
...
+
+        pop     hl
+        pop     af
+        ei
+        ret
+
+
+        SECTION bss_driver
+
+existing_int:
```
suborb, line 63: Should there be a `SECTION code_crt_exit` to copy back the vectors?

libsrc/target/aquarius/interrupt/im1_init.asm
```diff
@@ -0,0 +1,64 @@
...
+
+        EXTERN  asm_interrupt_handler
+        EXTERN  im1_vectors
+        EXTERN  l_pop_ei
+        EXTERN  l_push_di
+        extern  tick_count_isr
+
+        defc    Z80_OPCODE_JP=$c3
+        defc    EXISTING_INT_SIZE=3
+        defc    IO_IRQMASK=$ee
```
suborb, line 13: The defc from here could go into aqplus.def

libsrc/target/aquarius/interrupt/im1_init.asm
iratahack: Hi @suborb , thanks for taking a look. To be safe there absolutely be a di before messing with the vector I will add this.

However, I'm hesitant to use l_push_di and l_pop_ei due to the fact that examples like the sound example call im1_init but do not perform an explicit ei, so my assumption is im1_init enables interrupts on exit. What are your thoughts?

libsrc/target/aquarius/interrupt/im1_init.asm
iratahack: Great idea! I wasn't exactly sure how to handle this but code_crt_exit is it. I'll updated.

libsrc/target/aquarius/interrupt/im1_init.asm
iratahack: I've been talking to the AQ+ guys. There are two inc files I think we should pull in from their source base which will handle all of the Aq+ specific devices. sbasic.inc and plus.inc, where is the best place to put these so they can be used by z88dk and developer code?

lib/target/aquarius/classic/aqplusram.asm
```diff
@@ -65,8 +65,11 @@ setup_far_heap:
     push    af                  ;Save it
     srl     a                   ;/4
     srl     a
-    ld      de,+((((CLIB_FARHEAP_FIRST - AQPLUS_FIRST_BANK) >> 2) +1) % 256)
-    ld      hl,0
+    ;
+    ; Bank number is e:hl[15:14], offset is hl[13:0]
+    ;
+    ld      de,CLIB_FARHEAP_FIRST >> 2
```
suborb, line 71: This looks suspicious (if CLIB_FARHEAP_FIRST = 35)

The (virtual) far address space start from `0x010000` and the original intent was trying to map this virtual address onto a physical address, such that it's coherent with bank mappings. 

The original code assumed that the bank corresponding to AQPLUS_FIRST_BANK corresponded to far pointer 0x010000. Thus the shifting was trying to find the far pointer corresponding to CLIB_FARHEAP_FIRST.

include/games.h
```diff
@@ -300,7 +300,7 @@ extern const unsigned char *joystick_type[];
 
 #ifdef __AQUARIUS__
 #ifdef DEFINE_JOYSTICK_TYPE
-	const unsigned char *joystick_type[] = {"QAOP-MN", "8246-05", "hjkl-sd", "Cursor"};
+	const unsigned char *joystick_type[] = {"Joystick 1", "Joystick 2", "QAOP-MN", "8246-05", "hjkl-sd", "Cursor"};
```
suborb, line 303: This is just for the Plus?

iratahack: No. This save code should also work for the Aquarius.

libsrc/target/aquarius/interrupt/im1_init.asm
suborb: That's a good point - yes, an ei would be correct then.

lib/target/aquarius/classic/aqplusram.asm
iratahack: I left it at 35 since that's what it was originally, there is no reason for it to be 35. I can change it to AQPLUS_FIRST_BANK to match the banking.

libsrc/target/aquarius/interrupt/im1_init.asm
suborb: Plonk them in lib/target/aquarius/def Then the `OPTIONS` line in aquarius.cfg can have `-Ca-IDESTDIR/lib/target/aquarius/def` added to it - that way it's just `INCLUDE "plus.inc"` in files.

include/games.h
suborb: It uses the AY registers - I thought a regular Aquarius didn't have an AY?

iratahack: https://en.wikipedia.org/wiki/Mattel_Aquarius it had a single AY, here's the cheatsheet https://www.vdsteenoven.com/aquarius/hndcntrl.html

lib/target/aquarius/classic/aqplusram.asm
suborb: Actually, the 35 value was chosen based on this:

> Pages 32 through 47 is User RAM and is free for use by BASIC programs.

The assumption being that page 32,33,34 are going to be used by the main program loading, and that we can start banking into 0xc000 from page 35 onwards.

For automatic configuration, it can't be 32 since that will corrupt "main memory" when the far heap is initialised. Given that you've now harmonised the banks, 36 makes more sense since it's the start of a 64k chunk. 

So I think we need to ensure that CLIB_FARHEAP_FIRST is a multiple of 4, then restoring the following should work:

```
    ld      de,+((((CLIB_FARHEAP_FIRST - AQPLUS_FIRST_BANK) >> 2) +1) % 256)
    ld      hl,0
```

So:

* CLIB_FARHEAP_FIRST = 32, then de = 1, so far pointer = 0x010000
* CLIB_FARHEAP_FIRST = 36, then de = 2, so far pointer = 0x020000

include/games.h
suborb: Ah, the Mini-Expander add on.

lib/target/aquarius/classic/aqplusram.asm
iratahack: Yes, this makes sense.

iratahack: The +1 will break the mapping code as pages are no longer numbered 1-n they are 0x20-0x3f. Because of this there is no need to do CLIB_FARHEAP_FIRST - AQPLUS_FIRST_BANK it should be CLIB_FARHEAP_FIRST. So when you shift 0x24 right twice you get 0x90000, so the code below should suffice.

```
ld    de,CLIB_FARHEAP_FIRST >> 2
ld    hl. 0
```

iratahack:
I have fixed the compilation error due to a bad include file path. My far heap and banked test passes successfully as does the z88dk banked example.

libsrc/target/aquarius/far/l_far_mapaddr.asm
```diff
@@ -14,13 +14,11 @@ l_far_mapaddr:
     ld      a,e    ; If banked address is 0x00nnnn, then address absolutely
     and     a 
     ret     z
-    dec     e      ; __banked starts from 1
```
iratahack, line 17: No need to +1 or -1 as actual RAM page numbers are now used. Just need to shift the page number.

libsrc/target/aquarius/far/__far_page.asm
```diff
@@ -15,21 +15,19 @@ __far_page:
     and     a
     jr      z,localfar
     ld      d,e
-    dec     d
```
iratahack, line 18: No need to -1 as actual RAM page numbers are now used. Just need to shift the page number.

libsrc/target/aquarius/far/__far_page.asm
```diff
@@ -15,21 +15,19 @@ __far_page:
...
-    ; e is now which bank we should look at
+    ; d is now which bank we should look at
     ld      a,b
     and     @00111111      ;Take mod 16384
     or      @11000000	   ;Map to 0xc000 page
     ld      h,a
     ld      l,c
     ; hl = offset within bank
     ld      a,d
-    add     AQPLUS_FIRST_BANK	;0x01xxxx starts at AQPLUS_FIRST_BANK
```
iratahack, line 32: No need to +AQPLUS_FIRST_BANK as actual RAM page numbers are now used. Just need to shift the page number.

lib/target/aquarius/classic/aqplusram.asm
```diff
@@ -65,7 +67,10 @@ setup_far_heap:
     push    af                  ;Save it
     srl     a                   ;/4
     srl     a
-    ld      de,+((((CLIB_FARHEAP_FIRST - AQPLUS_FIRST_BANK) >> 2) +1) % 256)
+    ;
+    ; Bank number is e:hl[15:14], offset is hl[13:0]
+    ;
+    ld      de,CLIB_FARHEAP_FIRST >> 2
```
iratahack, line 73: If hl is forced to 0x0 it means that the starting far heap page number must be a multiple of 4.
Adding the line below means this restriction goes away.

```
    ld    hl, CLIB_FARHEAP_FIRST <<14
```
I don't mind either way, I just wanted to point it out.

lib/target/aquarius/classic/aqplusram.asm
suborb: Sorry yes, my confusion.

We should still modulus 4 to make life simpler and just add a 64k chunks to the heap apart from any remnant at the end.

lib/target/aquarius/classic/aqplusram.asm
suborb: As noted above, multiple of 4 is simpler.

lib/target/aquarius/classic/aqplusram.asm
iratahack: OK, I think this is good to go.

suborb:
Agreed, thanks for the changes and simplification of the memory map.

PR #2610, opened 2024-10-01
Updated sound example apps.

iratahack:
Updated WYX example app to skip 1 out of every 6 raster interrupts for the Aquarius since it's interrupts are 60Hz and the music was written for 50Hz. Also updated the data for the effects and made the application a little more user friendly.


iratahack:
@suborb , I updated the code formatting.

PR #2617, opened 2024-10-06
Ensure hardware and stdio are sync'd during init.

iratahack:
@suborb , hopefully this addresses the issue you identified. My testing does show that when adding --generic-console, the correct crt0_init code is pulled in.

### DIFF
--- a/libsrc/target/aquarius/stdio/generic_console.asm
+++ b/libsrc/target/aquarius/stdio/generic_console.asm
@@ -18,8 +18,6 @@
     EXTERN      scrollup_BITMAP
     EXTERN      scrollup_TEXT
 
-    EXTERN      generic_console_ioctl
-
     EXTERN      __aquarius_mode
 
     INCLUDE     "ioctl.def"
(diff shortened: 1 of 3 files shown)


libsrc/target/aquarius/stdio/generic_console.asm
```diff
@@ -57,3 +62,26 @@ generic_console_scrollup:
     bit     2,(hl)
     jp      nz,scrollup_BITMAP
     ret
+
+
+    SECTION code_crt_init
```
iratahack, line 67: Would it be better to move this code to one of the crt as files surrounded by IF __HAVE_GENCON and __AQUARIUSPLUS__ so the code isn't even included to save a few bytes and not do the runtime check for CLIB_AQUARIUS_PLUS

suborb: That would work

iratahack: That me add this change to the PR.

suborb:
Thanks again for this

PR #2620, opened 2024-10-11
Review ZX getting started guide

derekfountain:
I noticed that recent updates to sdcc have made it a bit more pedantic, and the example code in the ZX getting started guide now produces some (harmless) warnings.

I fixed the example code and took the opportunity to check all the documents, fixing a bit a link rot and one or two other things.


suborb:
Thanks @derekfountain

PR #2625, opened 2024-10-17
Genie III LOWE CP/M disk format and good skew support in DMK containers

zx70:


zx70:
This pull fixes also the sector "codes":  1024=3, 512=2, 256=1, 128=0

PR #2652, opened 2024-12-14
IDE integration support fixes.

desertkun:
Currently, CLion (and maybe other) IDE fails to properly recognize z88dk CMake projects, lile [z88dk-gdb-ide-test](https://github.com/speccytools/z88dk-gdb-ide-test), because for several reasons:

- z88dk-ucpp generates like 1000 of empty lines on the preprocessing output for no reason: CLion C Preprocessor (called ReSharper) fails with stack-overflow because of the way it parses regular expressions. So I modified uccp to no produce those empty lines. See the bottom to explanation.
- zcc's `+<x>` command line option interferes with basic CMake self-identification. So I added an exception when `CMakeCCompilerId.c` w/ no `+<x>` is called, to assume `zx`.
- Ignore `-MD`, `-MT` and `-MF` w/o error.
- Verbosity for zcc decreced, unless `-vv` is added. It is interferes with post-processing: basially CLion stops respecting defines that compiler generates (like `SPECTRUM` for zx), because of `PROCESSING ...` which gcc does NOT log with verbosity enabled.

EXPLANATION

CLion feed ucpp the following file for preprocessing to observe preprocessor abilities.

```
void ____CIDR_command() {
#define ___CIDR_FEATURES_START
#if defined(__STDC_VERSION__)
{ int __attribute__ ((unused)) ____CIDR_command_SET_LANGUAGE_STANDARD, ____CIDR_command_STANDARD_C = ((void) __STDC_VERSION__, 0); }
#endif
#if defined(__cplusplus)
{ int __attribute__ ((unused)) ____CIDR_command_SET_LANGUAGE_STANDARD, ____CIDR_command_STANDARD_CPP = ((void) __cplusplus, 0); }
#endif
#ifdef __FILE_NAME__
{ int __attribute__ ((unused)) ____CIDR_command_ADD_DEFINITION, ____CIDR_command___CIDR_macro__FILE_NAME__ = ((void) "_file_name_short_", 0); }
#endif
#if defined(__has_builtin)
#define __CIDR_HELPER_DEFINE_0
#endif
#if defined(__CIDR_HELPER_DEFINE_0)
{ int __attribute__ ((unused)) ____CIDR_command_SET_FEATURE, ____CIDR_command_HAS_BUILTIN = ((void) 1, 0); }
#else
{ int __attribute__ ((unused)) ____CIDR_command_SET_FEATURE, ____CIDR_command_HAS_BUILTIN = ((void) 0, 0); }
#endif
#if defined(__has_feature)
#define __CIDR_HELPER_DEFINE_1
#endif

... and it goes on
```

ucpp produces the following

```
void ____CIDR_command() {

{ int __attribute__ ((unused)) ____CIDR_command_SET_LANGUAGE_STANDARD, ____CIDR_command_STANDARD_C = ((void)  201710L , 0); }






{ int __attribute__ ((unused)) ____CIDR_command_ADD_DEFINITION, ____CIDR_command___CIDR_builtin__has_include = ((void) 0, 0); }



{ int __attribute__ ((unused)) ____CIDR_command_ADD_DEFINITION, ____CIDR_command___CIDR_builtin__has_include_next = ((void) 0, 0); }



{ int __attribute__ ((unused)) ____CIDR_command_ADD_DEFINITION, ____CIDR_command___CIDR_builtin__has_cpp_attribute = ((void) 0, 0); }



{ int __attribute__ ((unused)) ____CIDR_command_SET_FEATURE, ____CIDR_command_C_VARIABLE_DECLARATION_RULES_ARE_C89_COMPATIBLE = ((void) 0, 0); }










<2000 of empty lines which crashes CLion>
```

### DIFF
--- a/src/ucpp/cpp.c
+++ b/src/ucpp/cpp.c
@@ -2462,7 +2462,6 @@ static int parse_opt(int argc, char *argv[], struct lexer_state *ls)
 		} else if (!strcmp(argv[i], "-Z")) {
 			no_special_macros = 1;
 		} else if (!strcmp(argv[i], "-d")) {
-			ls->flags &= ~KEEP_OUTPUT;
 			print_defs = 1;
 		} else if (!strcmp(argv[i], "-e")) {
 			ls->flags &= ~KEEP_OUTPUT;
--- a/src/ucpp/lexer.c
+++ b/src/ucpp/lexer.c
@@ -450,6 +450,8 @@ void flush_output(struct lexer_state *ls)
 }
 #endif
 
+static char nl_flag = 0;
+
 /*
  * Output one character; flush the buffer if needed.
  * This function should not be called, except by put_char().
@@ -467,6 +469,23 @@ static inline void write_char(struct lexer_state *ls, unsigned char c)
 #endif
 	if (c == '\n') {
 		ls->oline ++;
+	} else {
+		nl_flag = 0;
+	}
+}
+
+/*
+ * Output one character; flush the buffer if needed.
+ * This function should not be called, except by put_char().
+ */
+static inline void write_char_nl(struct lexer_state *ls)
+{
+	if (nl_flag) {
+		ls->oline ++;
+		return;
+	} else {
+		write_char(ls, '\n');
+		nl_flag = 1;
 	}
 }
 
@@ -809,7 +828,7 @@ static inline int read_token(struct lexer_state *ls)
 		shift_state = 0;
 	}
 	if (!(ls->flags & LEXER) && (ls->flags & KEEP_OUTPUT))
-		for (; ls->line > ls->oline;) put_char(ls, '\n');
+		for (; ls->line > ls->oline;) write_char_nl(ls);
 	do {
 		c = next_char(ls);
 		if (c < 0) {
--- a/src/ucpp/macro.c
+++ b/src/ucpp/macro.c
@@ -113,6 +113,28 @@ static inline int check_special_macro(char *name)
 int c99_compliant = 1;
 int c99_hosted = 1;
 
+static void define_static_macro(const char *name, const char *value)
+{
+	struct macro *m;
+
+#ifndef LOW_MEM
+	struct token t;
+#endif
+	m = new_macro();
+#ifdef LOW_MEM
+	m->cval.t = getmem(strlen(value) + 2);
+	m->cval.t[0] = NUMBER;
+	mmv(m->cval.t + 1, value, strlen(value) + 1);
+	m->cval.length = strlen(value) + 2;
+#else
+	t.type = NUMBER;
+	t.line = 0;
+	t.name = sdup(value);
+	aol(m->val.t, m->val.nt, t, TOKEN_LIST_MEMG);
+#endif
+	HTT_put(&macros, m, name);
+}
+
 /*
  * add the special macros to the macro table
  */
@@ -147,6 +169,9 @@ static void add_special_macros(void)
 #endif
 		HTT_put(&macros, m, "__STDC_VERSION__");
 	}
+
+	define_static_macro("__SIZEOF_SIZE_T__", "2");
+
 	if (c99_hosted) {
 #ifndef LOW_MEM
 		struct token t;
--- a/src/zcc/zcc.c
+++ b/src/zcc/zcc.c
@@ -382,6 +382,7 @@ static char  *c_altmathflags = NULL;        /* "-math-z88 -D__NATIVE_MATH__"; */
 static char  *c_startuplib = "z80_crt0";
 static char  *c_genmathlib = "genmath@{ZCC_LIBCPU}";
 static int    c_stylecpp = outspecified;
+static char  *c_swallow_mf = NULL;
 
 static char  *c_extension = NULL;
 static char  *c_assembler = NULL;
@@ -568,6 +569,9 @@ static option options[] = {
     { 0, "lstcwd", OPT_BOOL|OPT_DOUBLE_DASH,  "Paths in .lst files are relative to the current working dir" , &lstcwd, NULL, 0},
     { 0, "custom-copt-rules", OPT_STRING,  "Custom user copt rules" , &c_coptrules_user, NULL, 0},
     { 'M', NULL, OPT_BOOL|OPT_PRIVATE,  "Swallow -M option in configs" , &swallow_M, NULL, 0},
+    { 0, "MD", OPT_BOOL|OPT_PRIVATE,  "Ignore -MD" , &swallow_M, NULL, 0},
+    { 0, "MT", OPT_BOOL|OPT_PRIVATE,  "Ignore -MT" , &c_swallow_mf, NULL, 0},
+    { 0, "MF", OPT_STRING|OPT_PRIVATE,  "Ignore -MF" , &c_swallow_mf, NULL, 0},
     { 0, "vn", OPT_BOOL_FALSE|OPT_PRIVATE,  "Turn off command tracing" , &verbose, NULL, 0},
     { 0, "no-cleanup", OPT_BOOL_FALSE, "Don't cleanup temporary files", &cleanup, NULL, 0 },
     { 0, "", 0, NULL },
@@ -768,8 +772,8 @@ int process(char *suffix, char *nextsuffix, char *processor, char *extraargs, en
     }
 
     if (verbose) {
-        printf("%s\n", buffer);
-        fflush(stdout);
+        fprintf(stderr, "%s\n", buffer);
+        fflush(stderr);
     }
 
     status = system(buffer);


feilipu:
Is there another way to fix the verbosity issue with `-v`, other than changing the current behaviour?

This proposed change means I'd have to revisit every guide, documentation, and C file, containing a hint or instruction containing `-v` and update them to `-vv`. Not looking forward to that...

desertkun:
The problem is zcc's behavior differs from gcc in this regard. I can do some shenanigans like suppress some of verbosity when -MT/MD are passed which do nothing for it otherwise, but it basically creates a delicate relationship between zcc and clion which I rather avoid.

The following lines break CLions parsing:

PROCESSING x
<commands being executed, likd ucpp>

gcc does not do it. Obviously outputting commands being executed has value and it saved me once, but I am trying to make zcc to comfrorm to gcc to make it a more adoptable compiler.

suborb:
What am I missing? Isn't it a case of not specifying -v and then the compiler output is in a form that can be parsed?

desertkun:
This pr includes 3 things, -v is one of them. Problem is I don't have control over CLion specifying the -v parameter, and I want z88dk projects to work in CLion out of the box.

suborb:
> Problem is I don't have control over CLion specifying the -v parameter

That was the bit I was missing.

Given gcc -v output:

```
dom@zebedee z88dk % gcc -v arr.c
Apple clang version 15.0.0 (clang-1500.3.9.4)
Target: arm64-apple-darwin23.4.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
 "/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang" -cc1 -triple arm64-apple-macosx14.0.0 -Wundef-prefix=TARGET_OS_ -Wdeprecated-objc-isa-usage -Werror=deprecated-objc-isa-usage -Werror=implicit-function-declaration -emit-obj -mrelax-all --mrelax-relocations -disable-free -clear-ast-before-backend -disable-llvm-verifier -discard-value-names -main-file-name arr.c -mrelocation-model pic -pic-level 2 -mframe-pointer=non-leaf -fno-strict-return -ffp-contract=on -fno-rounding-math -funwind-tables=1 -fobjc-msgsend-selector-stubs -target-sdk-version=14.4 -fvisibility-inlines-hidden-static-local-var -target-cpu apple-m1 -target-feature +v8.5a -target-feature +crc -target-feature +lse -target-feature +rdm -target-feature +crypto -target-feature +dotprod -target-feature +fp-armv8 -target-feature +neon -target-feature +fp16fml -target-feature +ras -target-feature +rcpc -target-feature +zcm -target-feature +zcz -target-feature +fullfp16 -target-feature +sm4 -target-feature +sha3 -target-feature +sha2 -target-feature +aes -target-abi darwinpcs -debugger-tuning=lldb -target-linker-version 1053.12 -v -fcoverage-compilation-dir=/Users/dom/z88dk -resource-dir /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/clang/15.0.0 -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk -I/opt/homebrew/include -I. -I/usr/local/include -internal-isystem /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/local/include -internal-isystem /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/clang/15.0.0/include -internal-externc-isystem /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include -internal-externc-isystem /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/include -Wno-reorder-init-list -Wno-implicit-int-float-conversion -Wno-c99-designator -Wno-final-dtor-non-final-class -Wno-extra-semi-stmt -Wno-misleading-indentation -Wno-quoted-include-in-framework-header -Wno-implicit-fallthrough -Wno-enum-enum-conversion -Wno-enum-float-conversion -Wno-elaborated-enum-base -Wno-reserved-identifier -Wno-gnu-folding-constant -fdebug-compilation-dir=/Users/dom/z88dk -ferror-limit 19 -stack-protector 1 -fstack-check -mdarwin-stkchk-strong-link -fblocks -fencode-extended-block-signature -fregister-global-dtors-with-atexit -fgnuc-version=4.2.1 -fmax-type-align=16 -fcommon -fcolor-diagnostics -clang-vendor-feature=+disableNonDependentMemberExprInCurrentInstantiation -fno-odr-hash-protocols -clang-vendor-feature=+enableAggressiveVLAFolding -clang-vendor-feature=+revert09abecef7bbf -clang-vendor-feature=+thisNoAlignAttr -clang-vendor-feature=+thisNoNullAttr -mllvm -disable-aligned-alloc-awareness=1 -D__GCC_HAVE_DWARF2_CFI_ASM=1 -o /var/folders/rf/bs86hlbs7hv_ql55ckzj3b700000gn/T/arr-1e0721.o -x c arr.c
clang -cc1 version 15.0.0 (clang-1500.3.9.4) default target arm64-apple-darwin23.4.0
ignoring nonexistent directory "/usr/local/include"
ignoring nonexistent directory "/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/local/include"
ignoring nonexistent directory "/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/Library/Frameworks"
#include "..." search starts here:
#include <...> search starts here:
 /opt/homebrew/include
 .
 /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/clang/15.0.0/include
 /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include
 /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/include
 /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/System/Library/Frameworks (framework directory)
End of search list.
 "/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ld" -demangle -lto_library /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/libLTO.dylib -no_deduplicate -dynamic -arch arm64 -platform_version macos 14.0.0 14.4 -syslibroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk -o a.out -L/opt/homebrew/lib -L. -L/usr/local/lib /var/folders/rf/bs86hlbs7hv_ql55ckzj3b700000gn/T/arr-1e0721.o -lSystem /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/clang/15.0.0/lib/darwin/libclang_rt.osx.a
Undefined symbols for architecture arm64:
  "_func", referenced from:
      _func2 in arr-1e0721.o
  "_main", referenced from:
      <initial-undefines>
ld: symbol(s) not found for architecture arm64
```

Maybe just sticking a colon after processed might be enough?

desertkun:
@suborb I found the root cause. First, the minimally reproducible example: it is focused around preprocessor only (`-E`)
```bash
gcc -xc -g -fpch-preprocess -v -dD -E -D___CIDR_DEFINITIONS_END cidr-command.txt
# or
zcc +zx -xc -debug -fpch-preprocess -v -dD -E -D___CIDR_DEFINITIONS_END cidr-command.txt
```
[cidr-command.txt](https://github.com/user-attachments/files/18140718/cidr-command.txt)

On the first impression, output is similar, *BUT* gcc actually outputs verbose logs in stderr!

```bash
gcc -xc -g -fpch-preprocess -v -dD -E -D___CIDR_DEFINITIONS_END cidr-command.txt 2> error.log
# this produces just definitions, not extra actions!

# but error.logs shows all them logs
cat error.log
Using built-in specs.
COLLECT_GCC=cc
OFFLOAD_TARGET_NAMES=nvptx-none:amdgcn-amdhsa
OFFLOAD_TARGET_DEFAULT=1
Target: x86_64-linux-gnu
Configured with: ..
Thread model: posix
... etc
```

ROOT CAUSE

CLion defines `___CIDR_DEFINITIONS_END` as a magic word to detect the preprocessor output.

At some point the definition ends up in the output like so
```
#define ___CIDR_DEFINITIONS_END 1
```

However, zcc is happy to report that to stdout, it forwards all those arguments to z88dk-ucpp:

```
PROCESSING cidr-command.txt
z88dk-ucpp -iquote"." -D__SPECTRUM -DSPECTRUM -D__SPECTRUM__ -D__Z80 -D__Z80__ -D___CIDR_DEFINITIONS_END -D__Z88DK  -DSCCZ80 -DSMALL_C -D__SCCZ80 -I"/usr/local/share/z88dk//include/arch/zx" -isystem"/usr/local/share/z88dk//include"   -d  "cidr-command" "/tmp/tmpXXru6iV9.i2"
```

.. which CLion detects as a sign that the marker is there and any defines should stop proceessing.

gcc instead does a similar thing to stderr.

SO,

I changed the behavior similar to gcc's to output those commands in stderr, such as it does not confuse the IDE.

desertkun:
@feilipu no `-vv` required with this change (stderr, similar to gcc, for verbose output)

feilipu:
Looks better. Thanks.

But is it not a bit limited to have the only target specification that works with CLion be tied to `+zx`?

I understand the point that the + syntax is somehow problematic, but perhaps requiring an alternative way to specify the target would be more flexible (if the default + is not present)?

IDK. Perhaps `-target=zx` or otherwise similar to gcc syntax?

If someone wants to use any other target with CLion then the default `+zx` would be an issue, wouldn't it?

desertkun:
@feilipu `CMakeCCompilerId.c` is different in this regard. CMake and CLion (and I assume other IDEs) use it to inspect the basic properties of a compiler (like name, C standard, etc), and not the platform the compiler is going be building for.

When this program is compiled, it's not even executed (and cannot be in case of zcc), but by having a successful compilation CLion can assume the compiler is working.

See example of a typical CMakeCCompilerId.c
[CMakeCCompilerId.c.txt](https://github.com/user-attachments/files/18141918/CMakeCCompilerId.c.txt)

On linux, the resulting binary would be inspected for values, I would assume, using `strings` or something like that.
With zcc, it could not even be inspected (unless I turn every stone around to understand how) but failure to generate that file prevents CLion from properly recognizing zcc as a grown-up compiler.

Having that said, since i've been working on this fix for a year now, it seems like CLion more or less properly builds and syncs the project w/o CMakeCCompilerId.c change, so I dropped for now. But to illustrate my point:

image

The +<profile> parameter interferece with basic self-indentification of a <compiler>.
image
So I guess I have to revisit this. Maybe detect that we're being self-identified, and assume "+test"? Or assume "+test" every time none specified, but issue a warning?

feilipu:
Notwithstanding I don't know about CLion, if I understand your point is that the default target would only be used in a single specific case to test for compliance.

So in that case `+test` would a be a more reasonable default target solution.

But, it still feels like a bit of a bodge to bake it in to the code. I can't really comment further. 🤷

desertkun:
Anyway, I removed that from the PR. Would appreciate the approval.

suborb (approved):

feilipu (approved):

feilipu:
Does the target specification using `+` still need a solution to keep CLion happy?

desertkun:
Yes, as a separate effort. Ideally we fugure out a way to make +x notation optonal altogether.

PR #2653, opened 2024-12-16
Updated to latest Aquarius+ kernel headers.

iratahack:
Updated kernel.inc and kernel.h for the Aquarius+ target to match the latest ROM release.


suborb:
Ugh, I hate it when this sort of thing happens.

We have a similar issue with the LM80c (https://github.com/z88dk/z88dk/blob/master/lib/config/lm80c.cfg) with the ROM entry points moving around. It might be best if we do a similar thing for the Aquarius.

iratahack:
@suborb , take a look at the updated pull request. I think it's probably a good idea to prepare for this even though the entry points used have not changed. I have defaulted to what is currently the latest version (0.23q) but it can be overridden with the -clib option. The developers say the entry points should be stable now and only new entries will be added so hopefully there won't be too much churn.

suborb:
Great - thank you.

PR #2665, opened 2025-02-07
rc2014 - 16550 UART

feilipu:
Spencer has launched new [16550 UART Modules](https://rc2014.co.uk/modules/dual-serial-module-16c2550/), available in  1 and 2 port versions. It is likely that in the medium term these new modules will become the default serial ports for the RC2014.

This reorganisation deletes the cpm and hbios subtypes, as they are just a replication of the same named cpm and hbios targets. They were originally added to support additional RC2014 specific hardware but, that has never happened, so they're redundant.

__EDIT__ need to reinstate cpm subtype as it is used to reach down to the FatFS library and IDE drives (under CP/M).

The goal is to support both the [BASIC](https://github.com/RC2014Z80/RC2014/tree/master/ROMs/CPM-IDE) and [CP/M-IDE](https://github.com/RC2014Z80/RC2014/tree/master/ROMs/CPM-IDE) platforms using the 16550 UART modules, through these drivers.


libsrc/_DEVELOPMENT/target/rc2014/startup/rc2014_crt_128.asm.m4
```diff
@@ -0,0 +1,397 @@
...
+
+    IF CRT_ABPASS > 0
+
+        ld a,h
+        ld b,l
+        jp CRT_ABPASS
+
+    ENDIF
+
+    ei
```
jagarsoft, line 316: If CRT_ABPASS > 0, shouldn't EI instruction be execute too?

feilipu: Thanks for the input.

I'll have a look at this when I'm back on it next week (afk currently).

But there's still quite a bit of work to do here, including reinstating the cpm subtype (required to access FatFS drives on the rc2014. Doh!).

Once I build / test the hardware then I can write and test the actual drivers too.

feilipu: I've moved the `ei` before a potential jump to the BASIC `ABPASS` function.
There would be no reason that interrupts would have been disabled at that point, so I probably is just to be sure.
https://github.com/z88dk/z88dk/pull/2665/commits/b23f470dc6c8c2876d00f7c34ed7ce291cc97514

feilipu:
This is now complete enough to merge.

TO DO:
- handle framing / overrun errors (nice to have).
- for classic lib - connect ttyin / ttyout for 8085 CPU (newlib is already dual port).

PR #2685, opened 2025-03-18
[WIP] Agon light MOS 3 alpha api

TurBoss:
This are small changes taken from the agon-mos 3 alpha 2 [PR](https://github.com/AgonConsole8/agon-mos/pull/123)

> Whilst APIs were added in MOS 3 alpha 2 to expose new underlying functionality, unfortunately some mistakes were made in their creation. A notable flaw was the use of the A register as an API parameter - it isn't actually possible to use the A register as that is used to specify which API call is being made with RST 8.
>
>This PR aims to address that by changing the APIs to use different registers.
>
>There may also be a few enhancements too, such as the addition of a flag byte to the "unpack RTC" API to control whether/how an RTC refresh call will be made to the VDP.

I placed the WIP tag because some parts are not well tested

Thanks to @stevesims


TurBoss:
Thanks

PR #2692, opened 2025-04-02
Workaround for fgets()

zx70:


suborb:
@zx70 What's the problem?

zx70:
This workaround solved the problem with this program: 
https://github.com/z88dk/z88dk-ext/blob/master/misc/spline.c

..if built for CP/M.
Oddly I couldn't reproduce the same behaviour on the ZX spectrum of by providing a test program ( e.g. hacking clock.c)

The problem: without this workaround the input sequence never ends, a linefeed character is put in the input buffer.

suborb:
If I remember correctly gets calls onto fgets for convenience. However fgets should leave linefeeds in the buffer and gets should remove them. 

So in the empty case the result is wrong. It’s probably mixed up with a bit of lf/cr mapping. 

I‘ll have a dig.

zx70:
I'm still wondering why my name is on top of that file  :D    I didn't remember I had put my hands on that portion of the library.
Thank you for the fix, it looks like something is still missing because I can't get the right behaviour even from the test case with the mentioned program on CP/M.
By the way, I understand it was too cumbersome to test it right, e.g. it crashes on other targets.
I'll first set up a clean test program, then hopefully I'll be able to get backwards to the root cause.

suborb:
It should now behave the same way as a host compile. 

That code looks like it’s looking for EOF to determine whether to skip the atof of the number/end input.  CTRL+d will simulate that.

zx70:
I think I get it, time to experiment a little.

zx70:
I put a tiny extra check in it, basing on the C documentation I could see online, an empty string should result in a NULL ptr.

suborb:
> I put a tiny extra check in it, basing on the C documentation I could see online, an empty string should result in a NULL ptr.

Have you got a reference for that? I thought that was a possibility based on the example but couldn’t find anywhere trusted that indicated it. K&R says null for error or eof only.

zx70:
I'm not accurate in checking the language references, but that's what I noticed in the code a couple of times, so I trusted the first webpage I found telling about gets 
IBM is probably more authoritative, and it confirms it 

https://www.ibm.com/docs/en/i/7.5?topic=functions-gets-read-line

suborb:
I read that as agreeing with me:

zx70:
Oh, right.  
Apparently some old program checked for a null pointer on the console input, perhaps if was considered as a "no characters provided" condition.
BTW I understand that even on the programmer perspective having to deal with a null pointer rather than a null string is a nightmare.

zx70:
...it fixed a long lasting problem on clisp.c  ;)

PR #2700, opened 2025-04-17
[zx81] VT ANSI console fixed

zx70:


zx70:
+ text emulation in mid resolution graphics (udg mode).

Issue #2717, opened 2025-05-11
SDCC build fails on Fedora 42


jorgegv:
Z88DK fails build on Fedora 42 when building SDCC.

There is an open upstream bug (https://sourceforge.net/p/sdcc/bugs/3846/) due to a mismatched `elf()` function declaration in SDCC sources. It compiles fine with older compilers, but on Fedora 42, GCC 15 defaults to compiling in C23 mode, and SDCC build fails.

The proposed upstream workaround is to compile SDCC with `-std=c17` option, but since this would also require patching the SDCC Makefile, I have opted for a more direct fix with a patch that directly fixes the mismatched declaration in  SDCC sources when preparing for building.

Pull request is here: https://github.com/z88dk/z88dk/pull/2718

jorgegv:
Wel, so far I'm building it with C23 on Fedora 42, and this one is the only error, so it seems it will not be very much affected...?

feilipu:
closed by #2718.

PR #2718, opened 2025-05-11
Fix/sdcc elf declaration - bug #2717

jorgegv:


feilipu:
If this bug persists through to v4.6.0, move it into the main patch file.

PR #2724, opened 2025-05-15
Upgrade to arkos3

jorgegv:
Arkos Tracker 3 was released in March 2024. The changes mainly involve making the tracker program multiplatform, but there are almost no changes to the players, so the upgrade should have been a trivial one...

...if it was not for @suborb haviing done a reformat on the autogenerated player sources, instead of the original player source under `suport/arkos`, as it was indicated in the generated player files! :-)

Just kidding, @suborb :-) Now, if you share the tool you used for the reformat, I can add the reformat step to the player generation Makefile, so that the final players can be created automatically whenever Arkos 4 sees the light.

BTW, the regenerated players compile fine for all targets, but I just tested them for ZX and CPC, which are my most familiar platforms and the ones I have setup my emulators for. Some help would be appreciated for testing in other platforms, e.g. Aquarius+ (@iratahack ?) and MSX.


suborb:
Oh god. Did I? It wouldn’t have been intentional. Just a clean up side effect. 

It’ll be asmstyle.pl in the bin folder that was used.

jorgegv:
> Oh god. Did I? It wouldn’t have been intentional. Just a clean up side effect.
> 
> It’ll be asmstyle.pl in the bin folder that was used.

Applied to the generated players, and added to the Makefile for future regeneration ;-)

Thanks for the quick answer...

suborb:
Sorry, I got a little obsessed with CP/M file io performance so took my eye off this one.

Let me take a look.

suborb:
I'm just going to merge and test MSX that way.

PR #2737, opened 2025-05-27
Don't apply optimization with RES instruction for 8080 target

quasiengineer:
8080/8085 has no such instruction, so it could break compilation for that targets.


feilipu:
Good catch!

Issue #2725, opened 2025-05-17
(classic) CP/M file io performance


suborb:
Pulling out an issue from the thread here: https://github.com/z88dk/z88dk-ext/pull/6#discussion_r2089616650

In summary CP/M file performance has always been an issue.

* Before 2023, when writing a byte we read the sector, modified the byte, rewrote the sector
* In 2023, a sector cache was added to reduce the number of read/write cycles. On real hardware this showed a significant performance improvement
* In the linked thread, byte by byte writing is being used, I was surprised that a userland sector cache made such a significant improvement and I've been digging to understand why.

To investigate this issue I've added support for CPM random file access to ticks in 0b7d9f00a9e6749c3871a0a272e52de71a9b82d1. This allows us to use the debugging tools with ticks (hotspots etc) to analyse what's going on.

There's 3 operating modes in test tool:

* Unbuffered (-U) which relies on buffering within the library, calling via stdio to fcntl on a byte-by-byte basis
* Buffered, userland buffering, calling via stdio to fcntl in 128 byte blocks 
* Direct, calling to fcntl on a byte-by-byte basis

The tool is basically reading a 1 megabyte file and then writing it to a different file.

[buf.c](https://github.com/user-attachments/files/20264538/buf.c.txt)

CP/M file io is fairly complex, so I've baselined against `+test` which is just a thin wrapper around the host, so it's ideal for it. Some measurements:

| Mode | Tick count (10e6)  |
|-|-|
| Unbuffered | 1441 |
| Buffered | 510 |

And for CP/M we have:

| Mode | Tick count (10e6)  |
|-|-|
| Unbuffered (before changes) | 14545 |
| Unbuffered (after changes)  | 5760  |
| Direct (after changes) | 4495  |
| Buffered (after changes) | 643 |

So we can see that the stdio component adds roughly 1000 (again CP/M variant is a bit more complex since It handles text mode). All the extra is in the `write()`/`read()` implementation.

Summary of changes:

| Mode | Tick count (10e6)  |
|-|-|
| 1 Unbuffered (before changes) | 14545 |
| 2 Return &fcb rather than index | 12986 |
| 3 Cache file position -> record number | 12708 |
| 4 Optimise size==1 case to avoid memcpy | 12628 |
| 5 Reorder local variables to make fcb easier to access in write | 12569 |
| 6 Move read fcb as above | 11435 |
| 7 Optimisation rule + inline cache contents check| 10595 |
[...]

PR #2738, opened 2025-05-28
cpm fcntl - sequential read & write for #2725

feilipu:
Modified `fcntl` routines to use BDOS sequential read and write calls.
Relies on the `record_nr` value being up to date.


feilipu:
Interesting outcome. Things are working now. But not for large files.

The instrumentation shows that the 0x2000 record is being skipped, not being read. Yet it is being written.
That causes the end of file to be lost.

```txt

RD - S2 81, EX 1f, CR 7d, RC 80, record_nr 00001ffd
RD - S2 81, EX 1f, CR 7d, RC 80, record_nr 00001ffd

RD - S2 81, EX 1f, CR 7e, RC 80, record_nr 00001ffe
RD - S2 81, EX 1f, CR 7e, RC 80, record_nr 00001ffe

RD - S2 81, EX 1f, CR 7f, RC 80, record_nr 00001fff
RD - S2 81, EX 1f, CR 7f, RC 80, record_nr 00001fff

WR - S2 81, EX 1e, CR 00, RC 00, record_nr 00001f00
WR - S2 81, EX 1e, CR 00, RC 00, record_nr 00001f00

WR - S2 01, EX 1e, CR 01, RC 01, record_nr 00001f01
WR - S2 01, EX 1e, CR 01, RC 01, record_nr 00001f01

WR - S2 01, EX 1e, CR 02, RC 02, record_nr 00001f02
WR - S2 01, EX 1e, CR 02, RC 02, record_nr 00001f02

...


WR - S2 01, EX 1f, CR 7d, RC 7d, record_nr 00001ffd
WR - S2 01, EX 1f, CR 7d, RC 7d, record_nr 00001ffd

WR - S2 01, EX 1f, CR 7e, RC 7e, record_nr 00001ffe
WR - S2 01, EX 1f, CR 7e, RC 7e, record_nr 00001ffe

WR - S2 01, EX 1f, CR 7f, RC 7f, record_nr 00001fff
WR - S2 01, EX 1f, CR 7f, RC 7f, record_nr 00001fff

RD - S2 82, EX 00, CR 01, RC 80, record_nr 00002001
RD - S2 82, EX 00, CR 01, RC 80, record_nr 00002001

RD - S2 82, EX 00, CR 02, RC 80, record_nr 00002002
RD - S2 82, EX 00, CR 02, RC 80, record_nr 00002002

RD - S2 82, EX 00, CR 03, RC 80, record_nr 00002003
RD - S2 82, EX 00, CR 03, RC 80, record_nr 00002003

RD - S2 82, EX 00, CR 04, RC 80, record_nr 00002004
RD - S2 82, EX 00, CR 04, RC 80, record_nr 00002004

...

RD - S2 82, EX 01, CR 7e, RC 80, record_nr 000020fe
RD - S2 82, EX 01, CR 7e, RC 80, record_nr 000020fe

RD - S2 82, EX 01, CR 7f, RC 80, record_nr 000020ff
RD - S2 82, EX 01, CR 7f, RC 80, record_nr 000020ff

WR - S2 82, EX 00, CR 00, RC 00, record_nr 00002000
WR - S2 82, EX 00, CR 00, RC 00, record_nr 00002000

WR - S2 02, EX 00, CR 01, RC 01, record_nr 00002001
WR - S2 02, EX 00, CR 01, RC 01, record_nr 00002001

WR - S2 02, EX 00, CR 02, RC 02, record_nr 00002002
WR - S2 02, EX 00, CR 02, RC 02, record_nr 00002002

WR - S2 02, EX 00, CR 03, RC 03, record_nr 00002003
WR - S2 02, EX 00, CR 03, RC 03, record_nr 00002003

```

suborb:
Recalling the tests from #2725, running with -B with an alternate handling of `ungetc()` and `fread()`

| Target | Test | Ticks * 10e6 |
|-|-|-|
| +cpm | Existing | 643 |
| +cpm | In progress | 584 |
| +test | Existing | 509 |
| +test | In progress |  504 |

This change has made the byte reading case a fraction slower so a bit more work needed.

suborb:
If you merge my branch above into yours then you should get a little bit of a boost.

feilipu:
The new `ungetc` code is working correctly, as far as I can see. The logic no longer goes into the cache path unnecessarily. Good news.

Bad news is that the sequential calls are not working for files larger than one extent. This is because the records `RC` byte is not being filled in correctly with the actually used records in a following extent `EX` when it is being opened. In the BDOS `OPENIT` should be doing this, but it is not being done as far as I can see.

The random calls don't have this problem. Yet they do exactly the same logic within the BDOS. So it is something wrong with my code.

I'm noticing that my code seems to let `CR` go to 0x80, which it should never do. Also `CR` should never be higher than `RC`. Perhaps that's why? Will need some more code review to see why this is not behaving as I'd expect.

feilipu:
So it seems that BDOS uses an out of range event for `CR` to trigger opening the next valid extent, when using the sequential read function. This will fill the `RC` with the correct number of used records in the new extent. So when `CR` is set to 0x80 (with `EX` 0x01) it is actually asking to read `EX` 0x02 and `CR` 0x00.

`RD in  - S2 80, EX 01, CR 80, RC 80, cached_record ffffffff, record_nr 0100, rwptr 00008000`

This "glitch" out of range hook is not used for write, or for the random write or read (or special random write) functions.

Given that the glitch over this threshold into the new extent is required to properly load the FCB with the new `RC` I doubt that it makes sense to use the sequential functions to support direct access (i.e. lseek). The random access functions will do this using their BDOS internal `POSITION` function, with the inherent inefficiency that they use this function on every call.

```c
void setrecord_rd(struct fcb *fc) __z88dk_fastcall
{
    uint16_t record_nr = fc->record_nr;

    fc->current_record = (char)record_nr & 0x7F;

    if( (record_nr % 0x80 == 0) && (record_nr != 0) ) {
        fc->current_record = 0x80;
        record_nr -= 0x80;
        }

    fc->extent = (uint8_t)(record_nr >> 7) & 0x1F;

    fc->s2 &= 0x80;     // preserve the clean/dirty bit 7
    fc->s2 |= (char)(record_nr >> 12) & 0x0F;
}

```
Here a new extent is being opened with only 0x38 records, so the `RC` is updated properly.
```
RD in  - S2 80, EX 01, CR 7c, RC 80, cached_record ffffffff, record_nr 00fc, rwptr 00007e00, offset 0, size 128, len 512
RD out - S2 80, EX 01, CR 7c, RC 80, cached_record ffffffff, record_nr 00fc, rwptr 00007e00, offset 0, size 128, len 512

RD in  - S2 80, EX 01, CR 7d, RC 80, cached_record ffffffff, record_nr 00fd, rwptr 00007e80, offset 0, size 128, len 384
RD out - S2 80, EX 01, CR 7d, RC 80, cached_record ffffffff, record_nr 00fd, rwptr 00007e80, offset 0, size 128, len 384

RD in  - S2 80, EX 01, CR 7e, RC 80, cached_record ffffffff, record_nr 00fe, rwptr 00007f00, offset 0, size 128, len 256
RD out - S2 80, EX 01, CR 7e, RC 80, cached_record ffffffff, record_nr 00fe, rwptr 00007f00, offset 0, size 128, len 256

RD in  - S2 80, EX 01, CR 7f, RC 80, cached_record ffffffff, record_nr 00ff, rwptr 00007f80, offset 0, size 128, len 128
RD out - S2 80, EX 01, CR 7f, RC 80, cached_record ffffffff, record_nr 00ff, rwptr 00007f80, offset 0, size 128, len 128

WR in  - S2 00, EX 01, CR 7c, RC 7c, cached_record ffffffff, record_nr 00fc, rwptr 00007e00, offset 0, size 128, len 512
WR out - S2 00, EX 01, CR 7c, RC 7c, cached_record ffffffff, record_nr 00fc, rwptr 00007e00, offset 0, size 128, len 512

WR in  - S2 00, EX 01, CR 7d, RC 7d, cached_record ffffffff, record_nr 00fd, rwptr 00007e80, offset 0, size 128, len 384
WR out - S2 00, EX 01, CR 7d, RC 7d, cached_record ffffffff, record_nr 00fd, rwptr 00007e80, offset 0, size 128, len 384

WR in  - S2 00, EX 01, CR 7e, RC 7e, cached_record ffffffff, record_nr 00fe, rwptr 00007f00, offset 0, size 128, len 256
WR out - S2 00, EX 01, CR 7e, RC 7e, cached_record ffffffff, record_nr 00fe, rwptr 00007f00, offset 0, size 128, len 256

WR in  - S2 00, EX 01, CR 7f, RC 7f, cached_record ffffffff, record_nr 00ff, rwptr 00007f80, offset 0, size 128, len 128
WR out - S2 00, EX 01, CR 7f, RC 7f, cached_record ffffffff, record_nr 00ff, rwptr 00007f80, offset 0, size 128, len 128

RD in  - S2 80, EX 01, CR 80, RC 80, cached_record ffffffff, record_nr 0100, rwptr 00008000, offset 0, size 128, len 512
RD out - S2 80, EX 01, CR 80, RC 80, cached_record ffffffff, record_nr 0100, rwptr 00008000, offset 0, size 128, len 512

RD in  - S2 80, EX 02, CR 01, RC 38, cached_record ffffffff, record_nr 0101, rwptr 00008080, offset 0, size 128, len 384
RD out - S2 80, EX 02, CR 01, RC 38, cached_record ffffffff, record_nr 0101, rwptr 00008080, offset 0, size 128, len 384

RD in  - S2 80, EX 02, CR 02, RC 38, cached_record ffffffff, record_nr 0102, rwptr 00008100, offset 0, size 128, len 256
RD out - S2 80, EX 02, CR 02, RC 38, cached_record ffffffff, record_nr 0102, rwptr 00008100, offset 0, size 128, len 256

RD in  - S2 80, EX 02, CR 03, RC 38, cached_record ffffffff, record_nr 0103, rwptr 00008180, offset 0, size 128, len 128
RD out - S2 80, EX 02, CR 03, RC 38, cached_record ffffffff, record_nr 0103, rwptr 00008180, offset 0, size 128, len 128
```

feilipu:
I'm still mulling on scenarios where the there would be a failure, and since there are never "gaps" in a file (partially filled extents) I don't have actual concrete examples.

Since the two BDOS call types are available at no cost, it might still be worth keeping options open. 

But I'm away for a few weeks, so I'll think on it some more before killing this off.

feilipu:
AI says it can happen.

The following scenarios can result in a non-last extent having fewer than 128 records:
### 1	File Truncation:
If a file is truncated (e.g., by a program reducing its size), a non-last extent may end up with fewer than 128 records. For example:
- A file originally has multiple extents, each filled with 128 records. A program truncates the file to a smaller size, causing an intermediate extent to have fewer records.
- The BDOS updates the RC field in the directory entry to reflect the new number of used records.
- How it happens: Using BDOS function 22 (Set File Size) or similar operations, the file size is adjusted, and the RC field for the affected extent is set to a value less than 0x80. For instance, a database or editor might truncate a file after deleting records.
### 2	Sparse or Non-Sequential Writes:
CP/M supports random record access via BDOS functions 33 (Read Random) and 34 (Write Random). If a program writes non-sequentially, it may leave gaps in an extent, resulting in fewer than 128 records being used. For example:
- A program writes 60 records to the first extent, then skips to a later extent for further writes, leaving the first extent with RC = 0x3C (60 records).
- This is common in applications like databases that use indexed or sparse file structures.
- How it happens: The BDOS updates the RC field based on the highest record written in the extent, even if earlier records are unwritten, resulting in a partially filled extent.
### 3	File System Errors or Manual Modification:
Errors or manual edits to the file system (e.g., via a disk editor or debugger) can result in an extent with an RC value less than 0x80. For example:
- A corrupted directory entry might incorrectly set the RC field to a lower value.
- A user or utility manually modifies the FCB or directory entry, altering the record count.
- How it happens: The RC field in the directory entry is directly modified, either by error or intentionally, causing CP/M to interpret the extent as having fewer records.
### 4	Large Block Sizes with Limited Allocation:
On disks with large block sizes (e.g., 8 KB), an extent might reference fewer blocks than needed to store 128 records. If only a subset of blocks is allocated, the extent may have fewer records. For example:
- A disk with 8 KB blocks might allocate only one block to an extent, limiting it to 64 records (8 KB ÷ 128 bytes = 64 records).
- This is rare and depends on disk format or allocation constraints.
- How it happens: The BDOS allocates fewer blocks than needed for 128 records, and the RC field reflects the actual number of records in the allocated space.
### 5	Application-Specific File Handling:
Certain applications (e.g., databases or specialized software) may intentionally create extents with fewer than 128 records to align with their data structures. For example:
- An application writes only a few records per extent to maintain a specific file layout, leaving intermediate extents partially filled.
- How it happens: The application uses BDOS functions to control record writes precisely, setting RC to reflect only the records used.

## Why This Is Uncommon
CP/M’s file system is designed to maximize extent usage, filling each extent with 128 records before creating a new one. The last extent typically holds fewer records, as it contains the file’s remainder. The scenarios above are exceptions, arising from:
- Programmatic file manipulation (truncation, sparse writes).
- Errors or manual intervention.
- Disk configurations with large block sizes.

Example
Consider a file with 200 records (25,600 bytes) across two extents:
- Extent 0: Records 0–127 (RC = 0x80, 128 records).
- Extent 1: Records 128–199 (RC = 0x48, 72 records). If the file is truncated to 150 records:
- Extent 0: Unchanged (RC = 0x80, 128 records).
- Extent 1: Updated to records 128–149 (RC = 0x16, 22 records). If a third extent exists (e.g., due to earlier writes), Extent 1 is no longer the last extent and has fewer than 128 records.

## Technical Notes
- Directory Entry: The RC field (byte 15) explicitly tracks the number of used records (0x00 to 0x80). If RC < 0x80, the extent is partially filled.
- BDOS Operations: Functions like Write Sequential (15), Write Random (34), or Set File Size (22) can result in partial extents by updating RC.
- Allocation: The AL field lists block pointers, but RC determines the actual data extent, allowing partial usage even with allocated blocks.
## Conclusion
In CP/M 2.2, non-last extents can have fewer than 0x80 records due to:
- [1] File truncation reducing the record count.
- [2] Sparse or random writes leaving gaps.
- [3] File system errors or manual edits altering RC.
- [4] Large block sizes limiting allocated records.
- [5] Application-specific partial extent usage.

These cases are exceptions, driven by specific operations or errors.

zx70:
My cent on this topic, the reason for me rolling back the code in the past was this program failing: 

https://github.com/z88dk/z88dk-ext/blob/master/os-related/lar.c

There could be something to fix also on the file redirection (command line parser, etc) but honestly I could never track the behaviour to describe a bug, of just to document the right way to use it if there's just some limit to work around.  I'm mentioning it because choosing the right programs to test the file library matters.

feilipu:
So the computer tells us that there are situations where there can be less than 128 records contained in an extent (that is not the last extent). But that these situations are exceptions to the normal state. But lets think about whether they're actually relevant.

> In CP/M 2.2, non-last extents can have fewer than 0x80 records due to:

> [1] File truncation reducing the record count.
> [2] Sparse or random writes leaving gaps.
> [3] File system errors or manual edits altering RC.
> [4] Large block sizes limiting allocated records.
> [5] Application-specific partial extent usage.

> These cases are exceptions, driven by specific operations or errors.

---

Analysis / Opinion
 
- [1] File truncation reducing the record count.

If a file is truncated, this wouldn't change any intermediate extents. Yes the last extent could now have fewer than 128 records, but this is the normal situation.

- [2] Sparse or random writes leaving gaps.

Reading the BDOS code, the random read and write functions do not provide a safety net for this situation. Any record number put into the random record bytes will be directly decoded into a representation of the CR EX and S2 bytes. So there is no change in behaviour here.

- [3] File system errors or manual edits altering RC.

Yes, this can happen anywhere.

- [4] Large block sizes limiting allocated records.

This might be an issue, particularly for some floppy disk types (e.g. Double Density 5.25" Floppy ???).
Need to understand further.

- [5] Application-specific partial extent usage.

Yes, this can happen anywhere.

---

Ok, noting that there are no safety nets to having non-standard record numbers in intermediate extents we can now think about whether doing this would be a win.

The random access method is to load the fcb record number into the R0, R1, and R2 bytes using `_putoffset.c`,  and the BDOS `POSITION` function is then called by the BDOS API to properly fill the CR, EX and S2 bytes. I asked the computer to calculate what this means in Z80 T-states, and it says about __565 T-states__ excluding called functions.

So to do the same thing using the sequential operations, the `set_record` function must be called to load the fcb record number directly into CR, EX, and S2 bytes. I asked again, and it says that the current implementations require between __750 and 896 T-States__ excluding called functions.

So currently, it would be slower to use the sequential BDOS functions. Unless the `set_record` function is rewritten to be better. That is a fairly easy job, so I'll tackle that next.

But on balance, I'm now less convinced there's a win here.

---

How many T-States does `_putoffset.c` use?
The number of T-states required, including CALL and RET operations but excluding called functions, is 217 T-states.

How many T-States does BDOS `POSITION` use?
The number of T-states required, including CALL and RET operations but excluding called functions, is 348 T-states.

How many T-States does `set_record_rd` use?
The number of T-states required, including CALL and RET operations but excluding called functions, is 896 T-states.

How many T-States does `set_record_wr` use?
The number of T-states required, including CALL and RET operations but excluding called functions, is 750 T-states.

feilipu:
@suborb I'm playing around with the `_putoffset.c` function trying to get an improved implementation. The current sccz80 result is about 220 T-States. I think this can be cut in half by simply iterating through the bytes of the long variable.

The code idea comes from elsewhere, but looks like this in C.

```c
void _putoffset(unsigned char *where,long offset)
{
    unsigned char *bytes = (unsigned char *)&offset;

    *where++ = *bytes++;
    *where++ = *bytes++;
    *where = *bytes;
}
```
I was hoping to see a result like this; which is about 118 T-States.

```c
; Function _putoffset flags 0x00000280 __smallc 
; void _putoffset(unsigned char * where, long offset)
; parameter 'long offset' at sp+2 size(4)
; parameter 'unsigned char * where' at sp+6 size(2
void _putoffset(unsigned char *where, long offset)
{
    #asm
    ld hl,6
    add hl,sp
    ld e,(hl)
    inc hl
    ld d,(hl)
    ld hl,2
    add hl,sp
    ld a,(hl+)
    ld (de+),a
    ld a,(hl+)
    ld (de+),a
    ld a,(hl)
    ld (de),a
    #endasm
}
```

But that is not what actually happens. The actual result is quite different and far from optimal.
Do you have any thoughts on a C grammar that might force this kind of assembly result?

Otherwise, perhaps for sccz80 8080/z80/gbz80 and 8085 builds it might be worth just using inline assembly?

```c
; Function _putoffset flags 0x00000280 __smallc 
; void _putoffset(unsigned char * where, long offset)
; parameter 'long offset' at sp+2 size(4)
; parameter 'unsigned char * where' at sp+6 size(2)
void _putoffset(unsigned char *where, long offset)
{
    #asm
    ld de,sp+6
    ld hl,(de)
    ld de,sp+2
    ld a,(de+)
    ld (hl+),a
    ld a,(de+)
    ld (hl+),a
    ld a,(de)
    ld (hl),a
    #endasm
}
```

---

I'm looking at `_putoffset.c` now, because I can't write a `setrecord.c` function (for sequential BDOS access) to be as efficient as the BDOS assembly implementation of `POSITION` for random access. It gets too messy to try to write it in assembly across all the different machine types.

So fixing `_putoffset.c` by removing over 100 T-states per 128 bytes read or written seems like a reasonable result that will have no side effects.

suborb:
I'd personally write it in assembler, but this is an interesting trick:

```
struct x {
        unsigned char buf[3];
};


void _putoffset(unsigned char *where,long offset) {
        *(struct x *)where = (struct x)&offset;

}
```

feilipu:
> I'd personally write it in assembler, but this is an interesting trick.

Yes. It does generate much better code in the most part. Good trick!
But, even that better code can be improved a little, so I'll do that as a result.

But there are some issues.
1. sdcc - baulks at the cast. `error 36: illegal cast (cast cannot be aggregate)`.
2. sccz80 - 8085 compilation issues an `ldir` instruction.

Resulting assembly for z80 is 110 T-states, and for 8085 it is 98 T-States.
Commit resolving this is [here](https://github.com/z88dk/z88dk/commit/a2b4fdb58414536d5d311d71d1d92913babdc873).

PR #2772, opened 2025-08-22
Banked calls for zxn target

vmorilla:
Implementation of banked calls for ZX Spectrum Next architecture following the discussion in https://www.z88dk.org/forum/viewtopic.php?t=12059

The segment to be used for banked calls can be configured with the pragma CRT_BANKING_SEGMENT:
- 0: MMU 0 and 1 (0x0000 - 0x3fff)
- 1: MMU 2 and 3 (0x4000 - 0x7fff)
- 2: MMU 4 and 5 (0x8000 - 0xbfff)
- 3: MMU 6 and 7 (0xc000 - 0xffff)


vmorilla:
I screwed something... I'm working on fixing it. This only works with the default MMUs (6 & 7)

libsrc/_DEVELOPMENT/target/zxn/zxn_rules.inc
vmorilla, line 1: This file includes the default values for the two pragma symbols defined:
- CRT_BANKING_SEGMENT
- CRT_BANKING_STACK_SIZE

vmorilla:
Now it should be working... please review the proposed distribution of the code in CRT files

libsrc/_DEVELOPMENT/target/zxn/zxn_crt.asm.m4
```diff
@@ -1048,3 +1048,6 @@ ifelse(__STARTUP, 799,
 
    include(`startup/zxn_crt_799.asm.m4')
 ')
+
+INCLUDE "startup/zxn_crt_banked.asm"
```
suborb, line 1052: This change adds the contents of zxn_crt_banked for all compilations - increasing their size.

I think to minimise disruption to historic projects I'd prefer that the file is conditionally included based on `CRT_BANKING_SEGMENT` - that is for banked calls to function `CRT_BANKING_SEGMENT` needs to be defined by the user.

vmorilla:
With the latest commit, the banking call implementation is included in the CRT only if CRT_BANKING_SEGMENT is defined, as suggested by suborb.

suborb:
Thanks you. Apologies for the delay. 

I‘ll add this into classic as well.

PR #2773, opened 2025-08-31
Some missing zxn definitions for sprites and tilemap registers

vmorilla:
Given the high number of registers and flags in the Spec Next architecture, I am suggesting a simplified approach to maintain them. In this approach, the definitions are included in .h files that are included from a .m4 script. The macro takes care of adding the content of the .h to config_zxn.h and transforming those definitions to the syntax required by the ASM files config_zxn_public.inc and config_zxn_private.inc.


vmorilla:
Removed dependency on sed on m4 macros

suborb:
Thanks.

PR #2776, opened 2025-09-06
(zxn) Move banked_call out of the crt for newlib

suborb:
* Add it to the classic library
* Create the banking stack on the stack as we do for other targets
* Harmonise the stack size pragma with other targets (so `CLIB_BANKING_STACK_SIZE` rather than `CRT_...`)
* Change `CRT_BANKING_SEGMENT` to `CLIB_BANKING_SEGMENT` for consistency

### DIFF
--- a/lib/target/zxn/classic/zxn_crt0.asm
+++ b/lib/target/zxn/classic/zxn_crt0.asm
@@ -26,6 +26,10 @@
     PUBLIC  __IO_NEXTREG_REG
     defc __IO_NEXTREG_REG = 0x243b
 
+    ; Always use segment 3 - otherwise we have to redo all the orgs for the banks
+    PUBLIC  CLIB_BANKING_SEGMENT
+    defc    CLIB_BANKING_SEGMENT = 3
+
 
 
     PUBLIC    _FRAMES
--- a/libsrc/_DEVELOPMENT/target/zxn/driver/banking/zxn_banked_call.asm
+++ b/libsrc/_DEVELOPMENT/target/zxn/driver/banking/zxn_banked_call.asm
@@ -2,14 +2,11 @@
    SECTION code_driver
    PUBLIC banked_call
    EXTERN l_jphl, __IO_NEXTREG_REG
-   EXTERN CLIB_BANKING_STACK_SIZE, CRT_BANKING_SEGMENT
+   EXTERN CLIB_BANKING_STACK_SIZE, CLIB_BANKING_SEGMENT
 
-   defc banking_mmu_low = 80 + CRT_BANKING_SEGMENT * 2
-   defc banking_mmu_high = 81 + CRT_BANKING_SEGMENT * 2
+   defc banking_mmu_low = 80 + CLIB_BANKING_SEGMENT * 2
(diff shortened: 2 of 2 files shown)


suborb:
@vmorilla do you mind verifying that I've not broken your use case with this refactor?

vmorilla:
Sure. Happy to. I will come back to you this weekend

vmorilla:
It works. Funny thing is your implementation is indeed similar to my first attempt (following your advice). However, I faced the problem that changing the value of CLIB_BANKING_SEGMENT made no difference: it always got the default one. I came to the (I guess wrong) conclusion that the drivers in the clib where compiled together with the z88dk, whereas the CRT code gets compiled with the user project. That's why I moved the code to the CRT... but I guess I had an error somewhere else...

So, I will take the chance to ask you a couple of questions:
- Are the drivers in the CLIB compiled also as part of the project?
- In case banking is not used... is the code for handling banked calls gets discarded (thus not increasing the size of the project) unlike the code at the CRT?

suborb:
The segment should be evaluated at link time since it’s extern - it’s a trick we use in a lot of places. 

I’ll check though.

vmorilla:
> The segment should be evaluated at link time since it’s extern - it’s a trick we use in a lot of places.
> 
> I’ll check though.

Sorry, check my updated message. I verified it works

suborb:
>It works. Funny thing is your implementation is indeed similar to my first attempt (following your advice). However, I faced the problem that changing the value of CLIB_BANKING_SEGMENT made no difference: it always got the default one. I came to the (I guess wrong) conclusion that the drivers in the clib where compiled together with the z88dk, whereas the CRT code gets compiled with the user project. That's why I moved the code to the CRT... but I guess I had an error somewhere else...

Thanks for testing it out.

I wonder if you were including a file that defined CLIB_BANKING_SEGMENT? - In which case when assembled an expression "relocation" wouldn't be generated.

These expression relocations are really powerful - if you run z88dk-z80nm -a zxn.lib you'll see them, the critical ones here are:

```
    E = $0000 $0000 0: banking_mmu_low := 80+CLIB_BANKING_SEGMENT*2 (section code_driver) (file target/zxn/driver/banking/zxn_banked_call.asm:7)
    E = $0000 $0000 0: banking_mmu_high := 81+CLIB_BANKING_SEGMENT*2 (section code_driver) (file target/zxn/driver/banking/zxn_banked_call.asm:8)
```

Most linkers only allow `symbol +/- value` which is quite limiting.

> So, I will take the chance to ask you a couple of questions:

>Are the drivers in the CLIB compiled also as part of the project?

The files in target/zxn/driver/ are included in the crt so are compiled as part of the user program. Everything else isn't assembled, but expressions are evaluated at link time.

> In case banking is not used... is the code for handling banked calls gets discarded (thus not increasing the size of the project) unlike the code at the CRT?

Yup, because it's in the library it's only pulled in when used.

vmorilla:
> I wonder if you were including a file that defined CLIB_BANKING_SEGMENT? - In which case when assembled an expression "relocation" wouldn't be generated.

Precisely. That was the issue. Thank you!

Issue #2786, opened 2025-09-29
Please create a new numbered release


mcphail:
As mentioned in your documentation, there is substantial and ongoing change in this project. Using the bleeding edge can result in breakage of old code bases, and makes it difficult for collaborators on a project to ensure everyone is using the same build chain and tools. Version 2.3 was released way back in 2023 and is not a good representation of current z88dk documentation and capabilities.

Can you please make a new numbered/snapshot release, and continue to make these regularly?

Many thanks

feilipu:
List of files affected by a release, to save thinking time. 

https://github.com/z88dk/z88dk/issues/2371#issuecomment-1838176970

suborb:
TODO:

- [x] Version bumps #2787 
- [x] Merge #2787
- [x] Write release email
- [x] Finalise date in changelog.txt
- [x] Freeze tree
- [x] Wait for a nightly with all the changes
- [x] Create GitHub release
- [x] Upload nightly artifacts
- [x] Check sourceforge has mirrored releases/re-upload there
- [x] Promote snapcraft build from build
- [x] Promote docker image from nightly
- [x] Announce release on major target forums - zx, zxn, rc2014, hbios, etc

mcphail:
Thanks for looking at this so quickly!

PR #2787, opened 2025-09-30
Issue #2786: Version bumps in prep

suborb:
Think this is all the version numbers


feilipu:
All the new library files that I can picture, that need changed.

PR #2790, opened 2025-10-05
Draft: Start the library flattening

suborb:
All very draft and WiP (and may fail completely when I realise I've missed something)

Just an internal reorg at this point: 

- [x] Move _DEVELOPMENT/target into _DEVELOPMENT/newlib/target
- [x] Move _DEVELOPMENT/target/*.inc into lib/crt/newlib/
- [ ] Move crts from _DEVELOPMENT/target into lib/target/[machine]/newlib
- [ ] Pull down classic makefiles into _DEVELOPMENT
- [ ] Pull down libsrc/target into _DEVELOPMENT/target
- [ ] Move classic only libraries in libsrc into _DEVELOPMENT/classic
- [ ] Move _DEVELOPMENT/arch into _DEVELOPMENT/target/
- [ ] Incorporate more of _DEVELOPMENT/arch into classic targets
- [ ] Move _DEVELOPMENT/* up to libsrc/

At this point there should be no net change for users (apart from some paths changing)

Stage 2:

Some thoughts:

* Trying to remove the need for users to rebuild libraries
* Reducing newlib down to "just a different set of crts"

- [ ] Split newlib target libraries into machine specific and CPU libraries (i.e. arch specific + stdlib (-stdio)) - can these be shared with classic CPU_crt0.lib files?
- [ ] Create extra libraries for eg rc2014 IDE hardware
- [ ] Review use of m4 - can z80asm now do all of the things needed?

Note: Windows builds will be broken since directories have changes

---8<---

Notes whilst going through:

- input library needs name harmonising 
- compress/zx7 sharing not possible: d12da0c0cb3e9461f1937ab899592d39b5482897


suborb:
Closing for the moment since this is going to be very noisy

PR #2800, opened 2025-10-29
Proposal: Flattening newlib and classic libsrc

suborb:
Just a reorg at this point:

- [x] Move _DEVELOPMENT/target into _DEVELOPMENT/newlib/target
- [x] Move _DEVELOPMENT/target/*.inc into lib/crt/newlib/
- [x] Move crts from _DEVELOPMENT/target into lib/target/[machine]/newlib
- [x] Pull down classic makefiles into _DEVELOPMENT
- [x] Pull down libsrc/target into _DEVELOPMENT/target
- [x] Move classic only libraries in libsrc into _DEVELOPMENT/classic
- [x] Move _DEVELOPMENT/arch into _DEVELOPMENT/target/
- [x] Incorporate more of _DEVELOPMENT/arch into classic targets
- [x] Move _DEVELOPMENT/* up to libsrc/
- [x] Reduce use of recursive make for classic Sinclair targets
- [x] Classic ZX/TS2068/ZXN classic code into a single directory rather than scatter it across 3 different directories
- [x] Classic list files for Sinclair simplified - the makefile holds the dependencies
- [x] Classic +sms has access to SMSlib and other libraries from newlib
- [x] Classic +zx and +zxn has access to all the multicolour libraries

Notes:

* The use of `_GLOBS` and `_GLOBS_ex` - `_GLOBS` in `.mak` files is used as an argument to z80asm so that it does the expansion rather than the shell (the shell inevitably runs out of argument space). `_GLOBS_ex` (expanded) is used by make to track dependencies.
* Still some bits of zx-libs and sms-libs still to merge in
* `classic` and `newlib` directories contain code that is only used in those libraries

Next stage:

* Split newlib into a processor library + target specific library?

At this point:

- Flattened library structure (libsrc/_DEVELOPMENT has been banished)
- Newlib crt includes pushed into lib/crt/newlib
- Setjmp is the same implementation across all libaries
- `-clib=new` has gained some of the classic functions for ULA+

Basically at this point there should be no real change for the end user (beyond any non-makefile building of the library being broken). I'd suggest looking at the tree structure rather than the actual code changes (they're mostly search/replace in anywise). Tree can be viewed: https://github.com/z88dk/z88dk/tree/final_merge

Not planning on merging this immediately - I want to do more cleanup in a separate MR but would appreciate feedback.

### DIFF
--- a/include/cpm.h
+++ b/include/cpm.h
@@ -148,6 +148,8 @@ extern struct dpb __LIB__  *get_dpb(int drive)  __z88dk_fastcall;
 #define CPM_ILOG 24              /* get bit map of logged in disks */
 #define CPM_IDRV 25              /* interrogate drive number */
 #define CPM_SDMA 26              /* set DMA address for i/o */
+
+/* The following functions are available on CP/M v2 and later */
 #define CPM_SUID 32              /* set/get user id */
 #define CPM_RRAN 33              /* read random record */
 #define CPM_WRAN 34              /* write random record */
@@ -155,8 +157,8 @@ extern struct dpb __LIB__  *get_dpb(int drive)  __z88dk_fastcall;
 #define CPM_RREC 36              /* update random access pointer */
 #define CPM_SRDS 37              /* selectively reset disc drives */
 #define CPM_WRZF 40              /* write random record with zero fill */
-#define CPM_DSEG 51              /* set DMA segment */
 
+/* Available only if the graphics extension has been applied */
 #define CPM_GSX 115              /* enter GSX */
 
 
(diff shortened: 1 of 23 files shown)


suborb:
Stage 2 is forming here: https://github.com/z88dk/z88dk/tree/cleanup_target - the aim there is to tidy up zx/zxn/ts2068/sam which all share code.

suborb:
I'm running out of steam on target cleanup and everything I've tried compiles and runs.  I'll merge this in after the weekend.

dbolli:
Great work so far and it's really appreciated 👍

I downloaded yesterday's nightly build and so far no issues.

Cheers,
Derek.

suborb:
> Great work so far and it's really appreciated 👍
> 
> I downloaded yesterday's nightly build and so far no issues.

Nightly tracks the master branch unfortunately.

suborb:
Might as well do it now!

Issue #2815, opened 2025-11-22
(newlib) Use one set of headers


suborb:
Summary:
 
* In earlier years sccz80 choked when sdcc's function decorators were used
* Classic has proven via the use of `<sys/compiler.h>` that a single set of headers can be used for both sdcc and sccz80.
* The `-clib=clang_ix`/`-clib=clang_iy` options aren't used (and that particular way of using clang hasn't been tested in years)
* The `ez80-clang` integration in classic has been more recently tested and is much easier to get running

Proposal:

* [x] Generate a single set of headers for newlib that with appropriate pre-processor macros works for both `-clib=new` and `-clib=sdcc`
* [x] Remove clang support from newlib

suborb:
Merged via #2819

Issue #2814, opened 2025-11-22
(newlib) Library rationalisation


suborb:
Following on from that monster PR, I'm planning a few much smaller ones.
 
* A large % of the code for sdcc_ix, sdcc_iy and sccz80 variants of newlib is identical.
* sdcc_iy is the recommended sdcc library to use
* sdcc and sccz80 use different entry points to libraries
* classic caters for machines with a single index register

With these points in mind, I'm proposing:

* Discontinuing sdcc_ix
* Creating a single library that contains `c/sccz80` and `c/sdcc_iy` along with `z80/` files

Care should be taken with the following functions (that take callbacks):

* qsort
* bsearch
* ....

In an ideal world I'd keep sdcc_ix rather than sdcc_iy to allow more code sharing for single-index register machines with classic. I can't obviously see where the overhead of a `push/pop I*` dominates the runtime of a function that uses `i*`

feilipu:
> * sdcc_iy is the recommended sdcc library to use

Good outcome. 
Though as a rule I have avoided using index registers where possible because of the (current) complexity of two libraries using different index registers.

> In an ideal world I'd keep sdcc_ix rather than sdcc_iy to allow more code sharing for single-index register machines with classic. I can't obviously see where the overhead of a `push/pop I*` dominates the runtime of a function that uses `i*`

Wouldn't some of the overhead be hidden (i.e. not obvious) in the function call set up, where the library was originated from C?

Anyway, not an important point at this stage.

👍

feilipu:
Just as a note for further action, the `z88dk-lib` tool will need to be adjusted too, to install third party libraries in the correct (new) locations, and generate headers from provided proto files.

suborb:
> Wouldn't some of the overhead be hidden (i.e. not obvious) in the function call set up, where the library was originated from C?

Broadly that's going to be a push/pop ix in c/sdcc_ix for the handful of functions that use an index register. It makes me wonder if most of the reported performance difference is coming from `--reserve-regs-iy` and sdcc generating alternate code.

feilipu:
> Broadly that's going to be a push/pop ix in c/sdcc_ix for the handful of functions that use an index register. It makes me wonder if most of the reported performance difference is coming from `--reserve-regs-iy` and sdcc generating alternate code.

Sorry for my confusion, but I thought that we used `—reserve-regs-iy` for both library versions?

[...]

PR #2817, opened 2025-11-23
(newlib) Remove sdcc ix and clang

suborb:
This is part of #2814 and #2815


suborb:
Closing this one - will do the clang removal first.

PR #2818, opened 2025-11-25
Increase flexibility on the MSX platform

pvmm:
Add the following optional defines:
* CRT_ON_EXIT allows cartridge programs to return control to MSX-BASIC, which is necessasry for call extensions;
* REGISTER_SP allow the MSX to define the stack by itself (useful for cartridges programs that return control to MSX-BASIC);

Reference: https://www.z88dk.org/forum/viewtopic.php?p=24313#p24313

### DIFF
--- a/lib/config/msx.cfg
+++ b/lib/config/msx.cfg
@@ -20,6 +20,7 @@ SUBTYPE		diskio -Cz+msx -Cz--disk -startup=1
 SUBTYPE		wav -Cz+msx -Cz--fmsx -Cz--audio -Cz--fast -startup=1 -lndos
 SUBTYPE		wavio -Cz+msx -Cz--fmsx -Cz--audio -Cz--fast -startup=1 
 SUBTYPE		rom	-Cz+msxrom  -startup=3 -lndos -D__ROM__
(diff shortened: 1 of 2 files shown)


pvmm:
Let me know if it needs improvement.

suborb:
There's a hierarchy for setting these, for example here: https://github.com/z88dk/z88dk/blob/master/lib/crt/classic/crt_rules.inc#L53

So if the user defines `CRT_` that has priority, followed by `TAR_` (defined by the crt) then `DEF_` which is the fallback.

So, actually these values can be injected in from `-pragma-define` command line arguments. For convenience it might be handy to define an alias in the .cfg that sets these values.

pvmm:
> So, actually these values can be injected in from `-pragma-define` command line arguments. For convenience it might be handy to define an alias in the .cfg that sets these values.

Here is a proposal of a new SUBTYPE in msx.cfg:
```
SUBTYPE         callext -Cz+msxrom  -startup=3 -lndos -D__ROM__  -pragma-define:CRT_MSX_CUSTOM_HEADER=1 -pragma-define:CRT_ON_EXIT=0x10002 -pragma-define:CRT_REGISTER_SP=-1
```

suborb:
@pvmm that looks good

pvmm:
I've tested it and my code does the same but now it's more concise.

pvmm:
I have changed CRT_REGISTER_SP to REGISTER_SP because it is what I've found in other targets. **cpm.cfg** for instance.

suborb:
I've stolen the key bit of your PR and pushed it directly into master - there shouldn't be any need for conditionals in the .asm

PR #2819, opened 2025-11-26
Issue #2815: Common headers

suborb:
* Move to a shared set of headers for newlib: `<sys/compiler.h>` sorts out the annotations for the different compilers
* z88dk-lib updated to not refer to sdcc/sccz80 for header location


suborb:
Merging this one since it fixes an issue when building the library (the .. in the path meant that we didn't assemble correctly for each library variant).

PR #2820, opened 2025-11-27
Re-aliasing of sdcc_iy + code removal

suborb:
As per the benchmarks in #2814 this removes sdcc_iy code

`-clib=sdcc_iy` is retained as an option for compatibility and turns on `--reserve-regs-iy` which makes contributes mainly to the historically reported performance difference.

This paves the way for a single library (eventually).


suborb:
Here we go...

PR #2828, opened 2025-12-06
Disable peephole optimization that could break i8085 code

quasiengineer:
Code to reproduce bug:
```c
uint8_t t[] = { 0x11 };

int main() {
  uint8_t * ptr = &t;
  uint8_t tmp;

  tmp = ptr[0];

  printByte(ptr[0]);
}
```

This optimization relies that `l_gint` does not modify `DE` register, but there is bunch of i8085-specific optimizations that replaces `l_gint` call with the code, that actually modifies `DE` (https://github.com/z88dk/z88dk/blob/master/lib/arch/8085/8085_rules.1)

### DIFF
--- a/lib/z80rules.1
+++ b/lib/z80rules.1
@@ -7222,6 +7222,7 @@ $notcpu 8085
 	dec	sp
 	push	hl
 
+%notcpu 8085
 %title Avoid going via stack
 	push	hl
 	ld	hl,%1	;const


feilipu:
Agreed.

There should be a i8085 specific optimisation that will work if offset is less than 255 and use the extended de stack access instructions. But avoiding the stack usage won't be possible.

suborb:
Thanks!

PR #2835, opened 2025-12-10
z88dk-gdb: support for debugging over device file, fixing CLion integration

desertkun:
This PR does few things:
1. Fixes IM2 mode of "gdb" so CLion can finally properly inspect and debug a emulator or a real spectrum.
2. Allows z88dk-gdb to connect over a device file (over USB CDC) on top of connecting thru tcp/ip

I am working on a cartridge for Spectrum, called [Spectranext](https://spectranext.net/) a successor of Spectranet, which among other things, allows to debug physical ZX Spectrum natively and without software running on it. Currently, the only way to debug real spectrum is with Spectranet AND additional software, which means only spectranet ecosystem can be debugged. Spectranext, however, can remain being a debugger while being disabled, which means you can debug whatever you have on your spectrum.

Spectranext uses Wi-FI so you can debug over traditional tcp, but also it exposes a USB CDC device, which can be debugged that way. I find device file debugging to be much less cumbersome and a recommended way.

@suborb if you have a real spectrum, I can ship you this device, so you can see it in action.

However, I am in a bit of catch-22, this cartridge is very much in alpha development, but in order to make z88dk part of its SDK, I must make sure z88dk can properly work with it, without developers having to complile z88dk themselves, so I am asking to get this merged first, wait for next release (maybe have an intermediary one?) so then I can make it part of SDK and get closer to device release and make some progress o Spectrum scene :)

### DIFF
--- a/src/ticks/debugger_gdb_packets.c
+++ b/src/ticks/debugger_gdb_packets.c
@@ -20,6 +20,8 @@ struct packet_buf
     int end;
 } in = {0}, out = {0};
 
+static uint8_t is_serial_connection = 0;
+
 uint8_t *inbuf_get()
 {
     return in.buf;
@@ -70,7 +72,13 @@ int read_data_once(sock_t sockfd)
     ssize_t nread;
     uint8_t buf[4096];
 
+#ifndef WIN32
+    // On Unix, use read() for both sockets and serial devices (sockets are file descriptors)
+    nread = read(sockfd, buf, sizeof(buf));
+#else
+    // On Windows, use recv() for sockets
     nread = recv(sockfd, buf, sizeof(buf), 0);
+#endif
     if (nread <= 0)
     {
         return -1;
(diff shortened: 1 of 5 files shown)


src/ticks/debugger_gdb_packets.c
```diff
@@ -86,7 +94,13 @@ void write_flush(sock_t sockfd)
     while (write_index < out.end)
     {
         ssize_t nwritten;
+#ifndef WIN32
```
suborb: The project I used it in has dropped off the internet, but I used a set of macros in it which sorted out on the winsock differences.

There's a gist here with the essence: https://gist.github.com/suborb/cc1a4828ae30247e21fe5064f23f44a2

suborb:
Sounds like an interesting piece of hardware, I never got around to getting hold of Spectranet at the time which was a shame.

The next release is "planned" in my head after the library merges have been completed which may well take a while. We're still building nightly and dockerhub should be up-to-date so there's no need for people to build it themselves - just provide a minimum version after this is merged?

desertkun:
This is what I am "doing" right now – https://github.com/spectranext/spectranext-sdk/blob/master/install.sh#L31 – I want to make it really easy for the devs who are probably not familiar with what z88dk is and pull it myself, so all you have to do is call a script and then open a project in an IDE. 

I will definately make do with nightlies for now. How do you make peoples macs not to complain about binaries, do you sign them?
What are you thoughs on the next frontier of publishing on brew? That'd be killer.

suborb:
The Mac builds are both signed and notarised so there's no warning message.

I've no plans to make a brew recipe, but that's not to stop anyone else from submitting one.

suborb:
Is this one ready for merge?

desertkun:
Apologies, did not react to your comment yet. I need to fix it, will do that over weekends.

desertkun:
@suborb I've taken a look at your gist and it does not apply here: on windows, I still must use win32 apis to open COM device. I am willing to wait for nightly and test it on on my windows machine after this PR is merged, because otherwise I do not have an environment to build for windows.

So, looks good.

Issue #2837, opened 2025-12-13
[x07][feature] Conio support? Even just cursor position and non-blocking keyboard input.


Fabrizio-Caruso:
Any chance to see some initial gotoxy support and getk?

The reason for this ticket is that a new coding jam on the X-07 has been launched:
https://discord.com/channels/679205550399160351/679205550843494420
https://www.facebook.com/groups/rpufos/posts/2025640488222230/

So many people could try Z88DK (including myself) on this target.

On the other hand some people will dig and learn more about some of the technical details that are not clear now.

suborb:
Now that we can reliably load programs things are looking up. Mame/Takeda is fairly easy to read.

One thing that I can't see emulated is the TV display (x820 device?) so we'll be stuck with 120x32 display.

Fabrizio-Caruso:
Great! I am searching for a capable emulator with support for the video module expansion .

Anyway even 128x32 would be enough for many games. Ideally with tiny fonts but I don't want to ask too much.

suborb:
Interesting, the TV interface is our old friend the MC6847:

<img width="845" height="1249" alt="Image" src="https://github.com/user-attachments/assets/30a91a97-6c35-4427-a213-edbbcacdc6ad" />

suborb:
Lot more still to do, but we have pixel control now:

<img width="269" height="130" alt="Image" src="https://github.com/user-attachments/assets/69d4e5ce-7a49-48bc-be21-e7cad0b37376" />

suborb:
I've merged support for --generic-console (using the native font size) + pixel plotting.

There's a few more things I want to add, but hopefully that's enough to get you started?

Fabrizio-Caruso:
@suborb Thanks! Great news!
I am not sure any of my games would be playable. I may try to write a new game that only needs 4 rows of text.
On the other hand, people who want to use C for the game jam, may use it for their new homebrews.
Which API can I use to plot a single point? 
4x4 fonts would produce 8 rows and could be easier to exploit for a game.

suborb:
Plot/unplot/xorplot. 

The small font is 4px wide. But there’s no positioning with it. Thought I’d start off with the easy stuff to implement!

Fabrizio-Caruso:
Sure. Please, do not feel pushed by enthusiasm. I love to get improved support but this is a hobby project for everyone here. 

[...]

PR #2838, opened 2025-12-18
X07 improvements

suborb:
* Pixel plotting etc
* putc4x6 has now gained lower case characters
* Basic gencon with the standard x07 font size

Part of #2837


Fabrizio-Caruso:
I use put4x6 ! Is it 4 pixel wide and 6 pixel tall? or viceversa?

zx70:
> * Pixel plotting etc
> * putc4x6 has now gained lower case characters
> * Basic gencon with the standard x07 font size
> 
 this was a difficult port, good one!

PR #2839, opened 2025-12-19
[BBC Micro] specific appmake option

zx70:
This is a new appmake option including the (already existing) Acorn Z80 CP/M configuration and the Torch Z80 CPN one


suborb:
@zx70 can you use unix style paths please - forward slashes

zx70:
> @zx70 can you use unix style paths please - forward slashes

Oops, I pasted the wrong version

PR #2860, opened 2026-01-15
Check length in read() before trying to read data

iratahack:
When the ESP reaches the end of the file, it returns an offered length of 0. In the current implementation this causes the code to hang trying to read a byte which never comes. I’ve updated the logic to check for a zero length before attempting to read any bytes.


suborb:
Ooops, thank you

PR #2865, opened 2026-01-24
Fixed a couple compiler warnings for Aquarius+ target

iratahack:
Changes tested with SDCC and SCCZ80.


suborb:
Thank you once again.

PR #2889, opened 2026-02-18
Updates to tool chain config for cmake

iratahack:
These updates clean up a few of the warnings when cmake configures and also fix an issue where it would try to use the native archiver (ar).


iratahack:
Hi @suborb any feedback on this PR? Cheers.

suborb:
@iratahack I don't use `cmake` but the changes look sensible! Apologies for the delay

PR #2891, opened 2026-02-19
HDOS disk format support

zx70:


suborb:
@zx70 you'll need to add hdos.c into the appmake.vcxproj file

PR #2906, opened 2026-03-11
Added --no-nop-comment option to zcc to opt out of nop comment processing

iratahack:
I'm working with third-party code which has inline assembly in C files compiled using sdcc. The C files have multiple locations where the following assembly is observed.

```
    .
    .
    nop    ; 4 = VRAM safe
    .
    .
```

zcc is stripping the `nop ;` from the line, leaving `4 = VRAM safe`. Then the peep-hole formatter steps in and thinks `4 = VRAM safe` should be a constant and changes it to `defc 4 = VRAM safe`. Assembler errors ensue.

To resolve this I have added a new option to zcc, `--no-nop-comment` to disable the nop comment processing. There may be a better way to do this, but the implementation is simple and it defaults to the existing behavior so shouldn't break existing code.


zx70:
Please, just confirm that "Remove unintended sdcc_opt.1 change from PR" is reverting to the original file and that your change is simply based on a new opt rule.

iratahack:
@zx70, I was trying to separate 2 sets of changes for different issues into separate PR's this one for the zcc change and #2904 for the sdcc_opt.1 changes. I would like to submit both.They solve unrelated issues.
IH.

PR #2910, opened 2026-03-11
Removed labels and circular `defc` aliases in copt peephole optimizer

iratahack:
## Problem

Compiling an infinite loop as the last statement of a function with
`sccz80 -O1` fails with an assembler error:

```
error: undefined symbol: i_2
```

(The symbol number varies with the function.)

### Reproducer

```c
#include <stdio.h>
void main(void) { printf("Hello World\n"); for(;;); }
```

```bash
zcc +sms -compiler=sccz80 -O1 -o main.bin main.c -create-app
```

---

## Root cause

`sccz80` emits two consecutive labels at the same address when a loop follows
another statement: the loop-top label and the EOS (end-of-statement) marker of
the preceding statement.  With `-O1`, two peephole rules in `z80rules.1`
interact destructively on this pattern.

**Pre-optimiser fragment:**

```asm
        jp      i_4     ;EOS    ; forward jump to EOS marker
.i_2                            ; loop-top
.i_4                            ; EOS marker (same address)
        jp      i_2     ;EOS    ; loop-back
```

**Rule 1 — De-dupe labels** fires on `.i_2 / .i_4` (stacked labels at same
address).  It kept the *upper* label (`.i_4`), deleted `.i_2`, and emitted
`defc i_2 = i_4`:

```asm
.i_4
        jp      i_2     ;EOS
        defc    i_2 = i_4
```

**Rule 2 — Trampoline** then fires on `.i_4 / jp i_2` (label followed by jump
to a *different* label).  It deleted `.i_4` and emitted `defc i_4 = i_2`:

```asm
        jp      i_2     ;EOS
        defc    i_2 = i_4   ; i_4 has no real definition
        defc    i_4 = i_2   ; i_2 has no real definition
```

Both real labels are deleted.  The `defc` aliases form a closed cycle with no
real label at either end.  The assembler exhausts the chain
`i_2 → i_4 → i_2 → …` and reports `undefined symbol: i_2`.

---

## Fix

**Change De-dupe to keep the *lower* label** (`.i_%1`) instead of the upper.

The loop-back jump always targets the lower-numbered label (the actual
loop-top).  Keeping `.i_%1` means:

- The `jp i_%1 ;EOS` loop-back targets a real, extant label directly.
- The Trampoline's guard (`%1 != %2`) is false on `.i_%1 / jp i_%1`, so it
  does not fire.
- Only one alias is produced: `defc i_%2 = i_%1` (upper → lower).

**Diff (`z80rules.1`, De-dupe labels rule):**

```diff
 %once
-.i_%2
+.i_%1
 %activate
 ; --- Start of Optimiser additions ---
 =
 %once
 ; --- Start of Optimiser additions ---
-	defc	i_%1 = i_%2
+	defc	i_%2 = i_%1
```

**Safety-net rule** added immediately after De-dupe to catch any future
circular pair where real labels are still present:

```
%title Remove circular defc aliases
	defc	i_%1 = i_%2
	defc	i_%2 = i_%1
=
	defc	i_%1 = i_%2
```

---

## Correct output after fix

```asm
._main
        call    _printf
        pop     bc
.i_2                        ; loop-top preserved as a real label
        jp      i_2     ;EOS
; --- Start of Optimiser additions ---
        defc    i_4 = i_2   ; single alias, upper → lower, no cycle
```

### DIFF
--- a/testsuite/results/Issue_xxxx_infinite_for.opt
+++ b/testsuite/results/Issue_xxxx_infinite_for.opt
@@ -1,0 +1,50 @@
+
+
+
+
+
+	INCLUDE "z80_crt0.hdr"
+
+
+	SECTION	code_compiler
+
+._main
+	dec	sp
+	pop	hl
+	ld	l,0
+	push	hl
+.i_4
+	call	l_gint3sp	;
+	ld	de,2
+	ex	de,hl
+	call	l_gt
+	jp	nc,i_5	;
+	ld	hl,0	;const
+	add	hl,sp
+	ld	(hl),15
+.i_6
+	jp	i_6	;EOS
+.i_7
+	inc	sp
+	ret
+
+
+	SECTION	bss_compiler
+	SECTION	code_compiler
+	defc	i_2 = i_5
+	defc	i_3 = i_5
+	defc	i_8 = i_6
+	defc	i_5 = i_8
+
+
+
+	SECTION	bss_compiler
+	SECTION	code_compiler
+
+
+
+	GLOBAL	_main
+
+
+
+


suborb:
I'll look at this series of PRs tonight.

Just out of interest, what tool did you use for this one?

iratahack:
> I'll look at this series of PRs tonight.
> 
> 
> 
> Just out of interest, what tool did you use for this one?

I was testing the power of Claude. It thought long and hard over this one.

PR #2912, opened 2026-03-12
Reformat n$: labels to @__sdcc_local_n:

iratahack:
I messed up my PR's (sigh) so this change got lost. This is in reference to #2904.

Also a small change to the crt for SMS to move __SMSlib_theLineInterruptHandler (which has an init value) from the BSS section to the DATA section.


iratahack:
@zx70 , sorry. I messed things up and this change was lost when merged. I've re-created it with a reference to the original PR. Thanks.

PR #2949, opened 2026-04-20
ucpp: fix empty include line sync issue

vmorilla:
When a C file that includes the same header twice (which is common and harmless due to include guards), the #line markers in the preprocessed output get wrong. As a result, when the compiler reported an error or a debugger pointed to a location, it would show the wrong line number, making it harder to trace problems back to the actual source.

For example, given:
```c
#include "guarded.h"   // line 1 — included normally
#include "guarded.h"   // line 2 — skipped (already included)

int test(void) {       // line 4
    return 0;
}
```

After preprocessing, int test should be tagged as coming from line 4. But due to the bug, the line counter was off-by-one after the skipped include, so it would report the wrong line — causing confusing compiler errors or debugger locations downstream.


suborb:
Thank you @vmorilla

PR #2964, opened 2026-05-13
pps: improve crt0 startup and add low-memory layout

witchcraft2001:
Update PPS (Peters Plus Sprinter) crt0 to preserve the DSS console by default, use native DSS console output, normalize DSS command tails before argv parsing, and use the standard REGISTER_SP setting for the EXE stack  address.

Add a low subtype for the $4100/$7FFF low-origin layout, fix native newline output to emit DSS-friendly CR/LF, and document PPS startup, memory layouts, command-line parsing, and examples.


lib/config/pps_low.cfg
suborb, line 1: Given the difference is just the pragma-define argument you should be able to fold this into a subtype on the main pps.cfg.

```
SUBTYPE low -pragma-define:CRT_ORG_CODE=0x4100 -pragma-define:CRT_STACK=0x7FFF
```

Then a compile with `zcc +pps -subtype=low ....`

lib/target/pps/classic/pps_crt0.asm
suborb, line 1: `REGISTER_SP` is normal pragma used for sp

witchcraft2001: changed

lib/config/pps_low.cfg
witchcraft2001: fixed

suborb (approved):
All good thank you

Issue #2974, opened 2026-05-28
Please test SDCC 4.6.0 RC1


spth:
Since today, we have SDCC 4.6.0 RC1. Since z88dk uses SDCC, it would be good to test early that SDCC 4.6.0 doesn't break z88dk. Since SDCC 4.5.0, substantial time passed, and many new features were added (which also increases the risk of new bugs having been introduced).

SDCC 4.6.0 RC1 can be found in the usual place: https://sourceforge.net/projects/sdcc/files/

feilipu:
Hi Philipp @spth 

at this stage (resources / timing) there's not much we can do with detailed testing of sdcc 4.6.0 release candidates before the final release, beyond adding it to a build to ensure that (with our small patch) things build normally.

Going forward (in line with the last years), the strategy we will pursue is to build off the stable release of sdcc as our basis, and if bugs are discovered we will try to replicate them in vanilla sdcc and then report them into your bug tracker. Only if they are urgent/critical to z80 family would we fix between stable sdcc releases.

Our [small zsdcc patch](https://github.com/z88dk/z88dk/blob/master/src/zsdcc/sdcc-z88dk.patch) does three things. 1. update the version information. 2. switch the abi to version 0. and 3. add the intrinsic functions we use.

The only remnant patch that we have is to resolve an issue with Fedora #2717 to allow the build to complete. It has probably been resolved in 4.6.0.

Perhaps also to note that there is usually a discrepancy between the svn commit number and the Revision actually recorded in the last lines of ChangeLog file. Usually this has to be patched by hand to get the build number to be reported correctly.

Once we have a 4.6.0RC1 build available, then I'm sure that some people will exercise it on their code.

spth:
SDCC 4.6.0 has ben released yesterday.

feilipu:
Tag for 4.6.0.

https://sourceforge.net/p/sdcc/code/16639/

PR #2976, opened 2026-05-31
Issue #2974: Prepare patch and tarball for sdcc 4.6.0

suborb:
* Usual patch
* Plus small change to permit writing of adb files from --c1-mode (see https://www.z88dk.org/forum/viewtopic.php?f=13&t=12139)
* Tarball is in place for builds


feilipu:
Not important here, but for the final release (for OCD correctness), the sdcc svn doesn't update ChangeLog file every commit, so the command line info reports the release number (rxxxxx) incorrectly.

Fixing this to match the actual release number means adjusting sdcc ChangeLog file to match actual release.

Then sometimes diff doesn't read text files, so it has to be manually prompted to actually produce a patch for the ChangeLog file.

Example [here](https://github.com/z88dk/z88dk/blob/090ab9ae9c0e8344170b949d0e650cae0599e123/src/zsdcc/sdcc-14648-z88dk.patch#L1).

feilipu:
@suborb I have pushed a tarball for the nightly server. it should be deleted from here once you're done.
I think I've captured the relevant things, including the adb files fix.

suborb:
I'll take the tarball, rebuild and then recreate a PR - this branch needs to die completely or we'll have a 30mb tarball in the history.

I think we might need some work around the rules - I noticed whilst investigating #3063 that with  4.6.0 the condition flags are now lower case.

suborb:
Tarball has been removed, hosted on the nightly server. Win32 and Mac binaries have been rebuilt.

Realistically I think this is probably good enough to merge now I've updated the condition rules.

feilipu:
Perhaps delete the RC1 patch file, and the reference in the readme?

I think it will be unnecessary noise now.

PR #2986, opened 2026-06-30
Extending stencil_render to the 8080 CPU

zx70:


libsrc/classic/gfx/narrow/__generic_stencil_render.asm
```diff
@@ -134,7 +142,15 @@ pattern_shift:
...
 xloop:
     ;;;ld    h,a    ; X1
+IF  !__CPU_INTEL__&!__CPU_GBZ80__
     rrc     e                           ; shifted pattern
+ELSE
+	push    af
+	ld      a,e
+	rrca
+	ld      e,a
+	pop     af
```
ped7g, line 152: Isn't A always free to use here, it's zero from previous `pattern_shift` loop upon first entry and in further iterations of `xloop` it's whatever `plotpixel` and `respixel` leaves it to...

Oh, even worse, for Z80 path the carry flag from `rrc e` is used to plot/unplot pixel, so this patch for 8080 and Gameboy will not work correctly with push af/pop af, as it discards the carry flag setup done by fake `rrc e`.

so this is not just optimization, but fix:
```
ELSE
  ld a,e
  rrca    ; set carry to plot/unplot pixel by pattern
  ld e,a
ENDIF
```

disclaimer: I spent about ~5min looking at this code seeing it first time in life, so I may be wrong about everything, verify me.

zx70: I'm shocked, it's such an immediate review!
You are right, restoring AF will destroy the carry.
Oddly I checked it on stencil.c and dstar.c and couldn't see side effects, but it deserves another pass.

PR #2989, opened 2026-07-01
Base line drawing library extended to 8080 CPU

zx70:


libsrc/classic/gfx/narrow/liner.asm
```diff
@@ -242,4 +245,322 @@ abs:
...
+    ret     z                           ; integer    is zero, return 0...
+;    bit    7,h
+    ld      a, 128                      ; trying to be 8080 compatible  ;)
+    and     h
+    jr      nz, negative_int
+    ld      a, 1
+    ret
+negative_int:
+    ld      a, -1
+    ret
```
ped7g:
```suggestion
	ld		a, h
	add		a, a						; CF = sign bit of H
	sbc		a, a
	or      1							; A = +1/-1 based on CF, ZF=0
	ret
```
I think this is 19T vs 28T on 8080? (and 19T vs 30/25T on Z80)
And 6 bytes vs 12 if I'm counting it correctly.

disclaimer: written from head, didn't test it, but I would have to have quite a brain fart to mess this one... :D

libsrc/classic/gfx/narrow/liner.asm
```diff
@@ -242,4 +245,322 @@ abs:
...
+;    A.BCDE../IXIY    same
+;    .F....HL/....    different
+;
+abs:
+;    bit    7,h
+    ld      a, 128                      ; trying to be 8080 compatible  ;)
+    and     h
+    ret     z                           ; integer    is positive...
+;    push    de
+;    ex    de,hl
```
ped7g, line 545:
```suggestion
;    Registers    changed after return:
;    ..BCDE../IXIY    same
;    AF....HL/....    different
;
abs:
	ld		a, h
	add		a, a
	ret		nc							; integer    is positive...
;    push    de
;    ex    de,hl
```
if my previous sgn suggestion works, the test for H.7 can be done by `add a,a` also here... -3T, -1B
And comment for 8080 version still claims A is preserved, but it is not. :)

(JFYI, I'm not doing thorough reviews of full merge request, I'm just taking quick look on some Z80 asm code here and there while I'm waiting for my paid job project to build, so looking at like 10-30 lines of instructions at most, without bigger context or testing anything, just having some fun...)

zx70: ..and I do appreciate it !    Sometime it feels like the code is being written for nobody, the whitebox check is totally unexpected.
I think I will leave it the way it is, I tend to do a lot of mistakes in such massive revisions,  I already spotted some from my past activity.    In the future the two versions should merge back into a single one with conditional branches, I'm sure your suggestion can improve the Z80 side as well.

Issue #3008, opened 2026-07-10
+cpm: uninitialised BSS is written into the flat .COM image


ravn:
By default CP/M binaries have uninitialized data area included in the generated FOO.COM file, making it larger than needed (the area is usually zeroed out anyway).

PR adds a failing but ignored test case showing this.

----
**copilot word salad**

`zcc +cpm` writes the program's uninitialised BSS region into the flat `.COM` file as trailing zero bytes, even though the classic CRT already zeroes BSS at startup (`lib/crt/classic/crt_initialise_bss.inc`: `ld hl,__BSS_head` / … / `ldir`). Those bytes on disk are pure redundancy and bloat every CP/M `.COM` by the size of its BSS.

**Repro**

```
$ cat t.c
unsigned char buf[8000];
int main(void) { buf[0] = 1; return buf[0]; }

$ zcc +cpm t.c -o t -create-app -m
$ ls -l T.COM ; grep __BSS_head t.map
-rw-r--r--  13189  T.COM
__BSS_head  = $12BA
```

`__BSS_head = $12BA` (4794), so code+data end at 4794 and the real image is `4794 - 0x100 = 4538` bytes. The remaining ~8.6 KB of the 13189-byte `.COM` is nothing but the zeroed `buf[]`. Same result under sccz80, sdcc and clang front ends — it's a target/CRT layout issue, not front-end specific.

**Expected**

The `.COM` should contain only code+data (4538 bytes here); the uninitialised BSS should not be materialised in the file. For contrast, `+z80`/newlib org's BSS as a separate binary and already excludes it from the image.

**Testcase**

The linked PR adds `src/z80asm/t/issue_NNNN.t`: it builds a `+cpm` program with a large uninitialised array, reads `__BSS_head` from the map, and asserts the `.COM` size equals `__BSS_head - 0x100`. The assertion is RED on current master; it is wrapped in a `TODO` block so it documents the bug without turning CI red (it reports as an expected failure now, and as "unexpectedly succeeded" once the `.COM` excludes BSS).

suborb:
This is fairly trivially fixed with:

```
diff --git a/lib/crt/classic/crt_section_bss.inc b/lib/crt/classic/crt_section_bss.inc
index f5596fa8bd..4659f3c508 100644
--- a/lib/crt/classic/crt_section_bss.inc
+++ b/lib/crt/classic/crt_section_bss.inc
@@ -2,6 +2,8 @@
 IF __crt_org_bss
[...]

PR #3009, opened 2026-07-10
Testcase for #3008: +cpm bundles uninitialised BSS into the .COM

ravn:
Adds a testcase for #3008: `zcc +cpm` writes the program's uninitialised BSS region into the flat `.COM` image, even though the classic CRT already zeroes BSS at startup, so those trailing bytes are pure bloat.

`src/z80asm/t/issue_3008.t` builds a `+cpm` program with a large uninitialised array, reads `__BSS_head` from the map, and asserts the `.COM` size equals code+data (`__BSS_head - 0x100`). The assertion is RED on current master and is wrapped in a `TODO` block, so it reports as an expected failure today (CI stays green) and as "unexpectedly succeeded" once the `.COM` excludes BSS — at which point the `TODO` should be removed.

This PR contains **only the testcase, no fix**.

### DIFF
--- a/src/z80asm/t/issue_3008.t
+++ b/src/z80asm/t/issue_3008.t
@@ -1,0 +1,53 @@
+#!/usr/bin/env perl
+
+# Issue 3008: +cpm writes the uninitialised BSS region into the flat .COM image.
+#
+# A CP/M program's BSS is zeroed at startup by the classic CRT
+# (lib/crt/classic/crt_initialise_bss.inc: ld hl,__BSS_head / ... / ldir), so
+# the BSS bytes need not be stored in the .COM.  On +cpm they nevertheless end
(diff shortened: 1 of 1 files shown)


pauloscustodio:
@ravn , thanks for contributing.

The test case is failing, because the BSS is bundled in the .COM image. Can you please change to a SKIP block, so that the overall z80asm tests pass? Then we can remove the SKIP after the bug is fixed.

pauloscustodio:
> The test case is failing, because the BSS is bundled in the .COM image. Can you please change to a SKIP block, so that the overall z80asm tests pass? Then we can remove the SKIP after the bug is fixed.

Sorry, my bad. I had a wrong config and the test failed. Passed after fixing the environment, all good.

PR #3010, opened 2026-07-12
Update z80nm to support the new object file format

pauloscustodio:
Current z80asm has object format v18. The new z80asm under development will 
update to v19 to add:

- Relocations of symbol+addend to avoid parsing expressions during linking

- Symbol index per cpu/swap-ixiy configuration for the library to
  parse only the needed modules instead of whole library

Added a tool z88dk-z80objcopy.pl to read/write any version of object file
and to export/import to an ASCII dump that can be used to change object 
files.


suborb:
> Relocations of symbol+addend to avoid parsing expressions during linking

We're quite dependent on this ability so removing expression parsing at link time is going to be very difficult

pauloscustodio:
>> Relocations of symbol+addend to avoid parsing expressions during linking

> We're quite dependent on this ability so removing expression parsing at link time is going to be very difficult

Version 19 does not remove expressions, it adds symbol+addend for the cases where this is possible, and keeps expressions for the more complicated cases. I expect to be able to reduce the time in the linker parsing and evaluating expressions, as most relocations of the form JP label are of the simple type.

PR #3015, opened 2026-07-16
80cc: More fixes, opts

suborb:
## Correctness fixes
  - **Signed `>>` was lowered as *logical*. ** A runtime `signed >> n` gave
    the unsigned result (`-16 >> 2` → `0x3FFC`, not `-4`); constant shifts folded
    early and hid it. Now marked arithmetic and lowered via `sra` (z80 family),
    `ARHL` (8085) or `l_asr`/`l_asr_dehl` (8080/gbz80). New 808x `l_asr_dehl`
    helper. Regression test in `long_ir/irgaps.c`. Unsigned shifts byte-identical.
  - **fp compile aborts / miscompiles** (`umaxd` and friends): kind/width builder
    fixes, byte `NEG` for width-1 dst, `load_to_dehl` sign-extension, width-4
    compare operand ordering.
  - **qsort/bsearch IX-frame corruption** under the fp frame pointer: classic
    `qsort`/`bsearch` rewritten to share the newlib core via a comparator closure,
    preserving the caller's IX (sccz80 + sdcc).
  - copt peephole fixes: post-increment reload → `dec hl`; const/sym via DE across
    an sp-adjust → reload in HL; drop redundant frame-slot reloads; `#285r`
    `(ix+%eval)` sign drop; strip adjacent `push/pop` no-ops.

  ## Performance
  - **Shift codegen** — inline 8085 native long const shifts (`rl de`/`rra`);
    `>>8` is a byte move; residual via `rrca`/`rlca` + mask; gbz80 TOS pop/push for
    32-bit slots. crcbench 8085 −20.8%; searchbench/histbench/maskbench 8085
    −12…14% (signed shifts no longer call a helper); gbz80 long-slot code smaller.
  - **Word residency / DE-home** — DE-clean pointer-chase accumulators, keep word
    call results in HL (fp), coalesce word copies, don't spill dead/reg-allocated
    word results.
  - **BC-pack** (default on) — live-range-pack call-free word temps into BC.
  - **stack-park** — NO_SLOT bytes parked on the stack (kills a latent below-sp
    write) + `IR_PR_STACK` word spill; matrixbench −8.5…12.9%.

  ## Infrastructure / measurement
  - Test harness: `FP=1` build mode + `run-matrix.sh` for fp/CPU coverage;
    exit-check every CPU's runtest; `#include <stdlib.h>` in the framework header.
  - LRA measurement scaffolding (inert): `IR_SPILL_STATS` slot-traffic counter and
    `IR_ALLOC_PROBE` residency estimates — off by default, no codegen effect.
.


feilipu:
I can't see the link from #289 (minor RC2014 and YAZ180 fixes) to this work. I guess it is a typo, that has propagated pretty widely. Might be good to fix the reference for future readers.

suborb:
There is no link! It’s an internal reference that I didn’t spot escaping. 

I’ve removed it from here but I think it’s in a commit so it will stay.

PR #3020, opened 2026-07-18
math32 - 8085 support

feilipu:
The IEEE754 single float solution for the z80 / z180 / z80n / kc160 and rabbits has been extended to support the 8085.

The solution was derived from the am9511 solution for the 8085, with new code written by grok-build 4.5.

The solution via grok-build required multiple steps, generally reflected as each interim commit. The detailed testing of each stage was handled by grok-build and this was a substantial time saver. Extensive testing on a whim "write me a test for..." is a godsend.

grok-build was guided by this [8085-skills](https://github.com/feilipu/8085-skills) document. Additions were made during development to refine the AI product.

Accuracy should be similar to the z80 solution, and performance will be addressed as part of follow up commits.

__EDIT__  optimisation produces about a 20% saving in the test suite and about 30% on the 32x32 multiply and add path for inverse, divide, and poly (used by all transcendental functions).


suborb:
There's math tests in test/suites/maths - might want to turn on 8085/math32

feilipu:
The helper already added some non-integer tests to the math suite too.

But, it seems to have broken other things. Will fix soonest.

PR #3025, opened 2026-07-27
newlib - +cpm migration

feilipu:
See [migration plan](https://github.com/z88dk/z88dk/blob/4f041b723b434eb08ec0412f2e60718907494dff/newlib_cpm_migration_plan.md) at top of repository.
Wheels turning now.


suborb:
I think the serial port emulation is hanging on msys2 - Cygwin is historically the slower one and both of the recent runs seem to get to roughly the same point.

feilipu:
Yes that was my thought, too. So I investigated that point exactly. Something like a stray ^c escaping the test somehow was the theory.

But now (with library fix) the Cygwin run passes, so not sure anymore.

The failure seems consistent at 6hr. Perhaps it is hung and is being killed at that moment?

Can't explain why the msys2 implementation and run is different to all others.

Anyway the test is just to prove the migration, so once we're ok with it I can pull out the whole (or part) of the test later.

I will minimise the test and fix library bodge (once nightly exists), if that's ok?

---

Trying a fix to scanf in the RC2014 emulation. If that works, then great!
The new target_io test tries to cover sufficient cases to ensure that nothing got broken, and the newlib cp/m build now has a file interface too.

feilipu:
The scanf problem seems fixed.

I will delete the Workplan from the repository, and attach it here.
Then merge the PR on soonest full completion.

[initial implementation plan (from above)](https://github.com/z88dk/z88dk/blob/4f041b723b434eb08ec0412f2e60718907494dff/newlib_cpm_migration_plan.md)

[newlib cpm migration outcomes](https://github.com/user-attachments/files/30443215/newlib_cpm_migration_plan.md)

Noting that the migration is fully "dual-stack". There is no integration of classic +cpm target into newlib +cpm. As future work, it might be interesting to add some of the classic machine subtypes into newlib +cpm. But, that was out of scope for this.

---

Future work is to migrate ZX and ZXN newlib machines. Using the JNext emulator to check for accuracy locally, as part of the process.

Issue #3022, opened 2026-07-25
newlib: status of FILE I/O (asm_target_open) — and would a contributed CP/M implementation be welcome?


ravn:
Human abstract: 

In the process of getting llvmz80 to use the existing z88dk standard libraries, it has been found that file support appears to be lacking in newlib and is needed as newlib as far as I have understood from claude supports the sdcccall(1) (register passing) convention directly that llvmz80 generates (but classic doesn't so each library routine call requires a register swapping bridge). 

Is the project interested in a AI-assisted attempt at writing the missing code? 

I am only interested in the CP/M target, but it could be a good start.

---

(rest written by Claude)

**Technical context**

On newlib, opening a file via stdio fails at link:

```
zcc +cpm -clib=new t.c     # fopen("f","w") -> undefined symbol: asm_target_open_p1 / _p2
```

`libsrc/newlib/fcntl/z80/asm_vopen.asm` (header dated *"October 2014"*) delegates the actual open to two target hooks, `asm_target_open_p1` / `asm_target_open_p2`. No target under `libsrc/newlib/target/*` implements them — the symbols appear only as that one `EXTERN`, and `git log -S asm_target_open` shows no implementation. `stdin`/`stdout`/`stderr` work because the CRT statically instantiates the console drivers, bypassing `vopen`, so no newlib z80 target can open a disk file through stdio (the `rc2014`/`yaz180` `diskio` drivers are raw block drivers, not wired to `fopen`).

This lines up with the "last mile" noted in #1426 and the forum guidance to use the classic library for file I/O. The wiki [stdio](https://github.com/z88dk/z88dk/wiki/stdio) page documents `fopen`/`fread`/`fwrite` with no newlib caveat, so today it surfaces only as a cryptic link error. Reproduces identically on sccz80 (`-clib=new`), sdcc (`-clib=sdcc_ix`) and clang/llvmz80 (`-clib=newlib_iy`).

**Questions**

1. Is newlib FILE I/O intentionally unsupported (use classic), or is `asm_target_open` meant to be completed?
2. If a contribution would be welcome, does this CP/M approach look right — an `asm_target_open` dispatcher (device-name vs filename) plus a disk-file driver delivering `STDIO_MSG_*` onto BDOS FCB calls, modelled on the existing driver classes and the classic `libsrc/target/cpm/fcntl/` FCB layer?

The implementation would be AI-assisted, reviewed and tested by me before any PR.

feilipu:
[...]

PR #3031, opened 2026-07-28
[cpm][newlib] console output is lost after an intervening fopen (#3022)

ravn:
During testing the new CP/M FCB file driver (#3025), Claude noticed that any console output after an `fopen()` call appears to be misrouted into the file.  Testing revealed this to be a newlib CP/M stdio/fcntl defect, not a compiler calling-convention one — the classic clib is unaffected.

This PR includes a test case failing under newlib and working under classic written by Claude showing the behavior. I have tested using the ntcvm emulator

I do not understand fully what is going on so the test case may be wrong.

---

_claude word salat:_

Adds a minimal repro for a newlib `+cpm` defect found while wiring the new CP/M
FCB file driver (#3025): opening a file corrupts the stdout console stream, so
console output after an `fopen` is misrouted into the file.

**Repro** (stock z88dk, no clang):

```
zcc +cpm -clib=new     -create-app -o caf console_after_fopen.c   # newlib
zcc +cpm -clib=default -create-app -o caf console_after_fopen.c   # classic
```

Run under a CP/M emulator (verified with ntvcm):

- **classic** prints both `CONSOLE-BEFORE-FOPEN` and `CONSOLE-AFTER-FOPEN` ✓
- **newlib** prints only `CONSOLE-BEFORE-FOPEN` — the second `puts()` is
  misrouted into the file ✗

The file write path itself is fine (the file is created correctly); it is the
console stream that gets rebound onto the file driver. Reproduces identically
under SDCC (`-clib=sdcc_iy`), so it is a newlib CP/M stdio/fcntl issue, not a
compiler calling-convention one. The classic clib is unaffected.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/test/suites/target_io/console_after_fopen.c
+++ b/test/suites/target_io/console_after_fopen.c
@@ -1,0 +1,36 @@
+/*
+ * console_after_fopen.c -- console output must survive an intervening fopen().
+ *
+ * Bug demonstrator for z88dk/z88dk#3022 (newlib +cpm CP/M FCB file driver).
+ *
+ * The program prints a marker to the console, opens (and closes) a file, then
+ * prints a second marker:
+ *
+ *     puts("CONSOLE-BEFORE-FOPEN");
+ *     fopen("caf.dat","w"); fclose(...);
+ *     puts("CONSOLE-AFTER-FOPEN");
+ *
+ * EXPECTED (and observed on the classic clib, -clib=default): both markers
+ * reach the console.
+ *
+ * ACTUAL on newlib (-clib=new): only the first marker appears.  Opening a file
+ * rebinds the stdout console stream onto the CP/M file driver, so the second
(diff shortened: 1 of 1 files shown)


feilipu:
The bug is confirmed on default. It is a memory reuse issue that is only in the default subtype. 

Will provide a fix then merge your test PR.

ravn:
> The bug is confirmed on default. It is a memory reuse issue that is only in the default subtype.
> 
> Will provide a fix then merge your test PR.

Great.   I'll continue when you have had a look at it.

feilipu:
Ok. I'll commit the test for the record, but it won't be run every time.

feilipu:
The heap fix (and a further wider heap audit) are building now.

The test you added passed locally for me. Let me know if same result for you.

PR #3034, opened 2026-07-29
Optimization ideas for float32

ped7g:
Just few low-hanging-fruit changes done in web interface of github (so it's mess split into 3 commits, not cleaned up).

NOT tested (beyond CI), and gains are calculated in my head, so I may be wrong about something.

So this is for someone else to pick up, verify, look for other spots in the lib and measure the real impact of these changes.

Summary of ideas:
implicit 1 creation:
```
  ld a,255
  add a,h  ; CF from (255 + exp) means CF = (exp != 0)
```

callee with two 16bit args: `ex (hl),sp` is 19T, while `pop bc : push bc` is 21T

caller with two 16b args: a quick guesstimate in head is 3x pop+push is 63T, the hl<-sp+2 and fetching them byte by byte is like 71T? (I assume `ld e,(hl+)` is `ld e,(hl) : inc hl`)

And the floor for negative value with fraction can do `+(-1)` instead of `-1` saving the sign flip of `m32_fssub_callee`, as the constant is set in floor anyway, so the sign flip can be baked in.

Sorry for not doing this properly and deeply, but this was just some quick fun for me, feel free to even ignore/close if nobody wants to pick on this. (I just have read talk about float32 being behind MS implementation and I got slightly curious to read into it for few minutes I had, just to figure out I would need to extract it into separate project and actually step over it in debugger as the code flow gets quickly more tricky to follow just in head ... but on the first read it feels like there should be more to gain with decent effort, not talking about effort like squeezing out last byte of 256 byte intro, just first-iteration effort).

### DIFF
--- a/libsrc/math/float/math32/asm/z80/d32_fsadd.asm
+++ b/libsrc/math/float/math32/asm/z80/d32_fsadd.asm
@@ -71,26 +71,20 @@ PUBLIC m32_fsadd, m32_fsadd_callee
     add hl,hl                   ; unpack op1
     ld c,h                      ; save op1.e in c
 
-    ld a,h
-    or a
-    jr Z,faunp1                 ; add implicit bit if op1.e!=0
-    scf
-
-.faunp1
+    ld a,255
+    add a,h                     ; CF = (op1.e != 0) to add implicit bit
     rr l                        ; rotate in op1.m's implicit bit
     ld a,b                      ; place op1.s in a[7]
 
     exx
 
-    ld hl,002h                  ; get second operand off of the stack
-    add hl,sp
-    ld e,(hl+)
-    ld d,(hl+)
-    ld c,(hl+)
-    ld h,(hl)
-    ld l,c                      ; hlde = seeeeeee emmmmmmm mmmmmmmm mmmmmmmm
+    pop bc                      ; get second operand off of the stack (preserve stack)
+    pop de
+    pop hl                      ; hlde = seeeeeee emmmmmmm mmmmmmmm mmmmmmmm
+    push hl
+    push de
+    push bc
     jp farejoin
-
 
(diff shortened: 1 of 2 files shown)


feilipu:
Thanks 🙏

Just hold fire for a day or two. I'm in the middle of a full rewrite, adding 8085 across math32 and math16. And optimising z80 paths where possible too.

The ability to specify alternative strategies for full implementation and a/b hotspot comparison by machine takes the guesswork out of what is actually faster.

It seems that some of my implementations, although clean and optically pretty, are non-optimal. 🤷‍♂️

Will come back to this later.

feilipu:
Just to follow up on the Microsoft "advantage".

It does a few things that don't make the cut today. Self modifying code is one. The other is static temporary register storage, which breaks multi threading.

So it has a few advantages that are difficult to overcome with self imposed rules of engagement.

ped7g:
@feilipu I don't plan to work on this further (I may, if I get similar spare time curiosity, but it's unlikely).

So feel free to "steal" any of this if you are already working on the math libraries, test it out, incorporate... this PR is like "hint".

feilipu:
Thanks. 🙏

feilipu:
closed by #3037

ped7g:
@feilipu ok, but I don't see any of this used by #3037 so far. Are you preparing more changes yet for that PR and I should check later?

Or should I comment there with suggestions to not throw away these completely?

feilipu:
I've pulled your critical path ideas into the main optimisation PR.

Overall, some of the benchmarks are now 10% to 18% better. Squeezed as much as possible out of it, after testing many different theories, and burning lots of A/B testing of multiple rounding strategies. Pretty happy with the outcome.

And, we're beating Microsoft at Whetstones 🥇

ped7g:
@feilipu nice... I will have few more review comments on the #3037 as well.

PR #3038, opened 2026-07-30
Optimize l_small_mul_16_16x16.asm

PsychedelicPalimpsest:
I noticed l_small_mul_16_16x16 was a bit slow and fixed it. I figured for this routine 100 extra bytes was worth it for something so critical (If not feel free to close).

I tested this in an emulator with a few million random numbers to make sure it worked.

llm summery of differences:
```
+----------------------+-------------------------+-------------------------+
| Metric / Scenario    | New Version             | Current Version         |
|                      | (Loop-Unrolled)         | (Iterative / Dynamic)   |
+----------------------+-------------------------+-------------------------+
| Approx. Code Size    | ~132 bytes              | ~32 bytes               |
+----------------------+-------------------------+-------------------------+
| 16x16 Multiplication |                         |                         |
|  - Best Case         | 454 T-states            | 844 T-states *          |
|  - Average Case      | 502 T-states            | 892 T-states *          |
|  - Worst Case        | 550 T-states            | 940 T-states *          |
+----------------------+-------------------------+-------------------------+
| 16x8 Multiplication  |                         |                         |
|  - Best Case         | 240 T-states            | 414 T-states            |
|  - Average Case      | 264 T-states            | 438 T-states            |
|  - Worst Case        | 288 T-states            | 462 T-states            |
+----------------------+-------------------------+-------------------------+
```

### DIFF
--- a/libsrc/math/integer/small/l_small_mul_16_16x16.asm
+++ b/libsrc/math/integer/small/l_small_mul_16_16x16.asm
@@ -15,29 +15,24 @@ l_small_mul_16_16x16:
    ;         hl = 16-bit multiplicand
    ;
    ; exit  : hl = 16-bit product
-   ;         carry reset
    ;
-   ; uses  : af, bc, de, hl
+   ; uses  : af, b, de, hl
+   ld b, l
+   ld a, h
 
-   inc h
-   dec h
-   jr z, eight_bit_1
-   
-   inc d
-   dec d
-   jr z, eight_bit_0
+   ld hl, 0
+   REPT 8
+       add hl, hl \ add a, a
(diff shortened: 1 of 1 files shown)


suborb:
The clue is in the name `_small_`!!

There's also a `_fast_` variant which should be much faster. 

If you're using classic, add `-lfastmath` to use it (or fiddle around with a pragma-redirect). If you're on newlib then there's config.inc tuning options and a library rebuild to switch over.

PsychedelicPalimpsest:
Ah shoot, I didn't see a fast version of mul_16_16x8, only mulu_16_8x8 and mulu_16_16x16.asm

In that case I'll close

PR #3054, opened 2026-08-08
math32 math16 - division rework

### LINKED ISSUE #3041: Discussion: math32 float division (m32_fsdiv) is ~2x slower than a direct restoring-division approach on z80
# Discussion: is math32's `float` division slower than it needs to be?

All files referenced below live in one folder: https://github.com/ravn/z88dk/tree/02ebba1d3e427834376d52407145bb84aee8dfc3/tasks/upstream-div-discussion-2026-08-01

**From @ravn (human):**

It was found during adaptation of llvmz80 to math32, that the math32
"div" routine for 32-bit floats was only half the speed of the implementation provided with the
compiler itself. 

Apparently this is because math32 calculates
1/x first using Newton-Raphson, which is slower on the z80 than just doing the division directly.   Note that the result is correct, but the algorithm is slower than it could be.  

License of llvmz80 code is permissive, so it is possible to adapt this implementation to math32 if desired.  

AI blurb below provide more detail.

---

**From Copilot (AI):**

## Summary

Comparing `m32_fsdiv` (math32's float division) against LLVM compiler-rt's
`__divsf3` for z80 showed compiler-rt's routine is markedly faster
(measured ~2.3x on a raw T-state comparison, see below). The two use
different algorithms: math32 computes `a/b` as `a * (1/b)`, finding the
reciprocal via Newton-Raphson iteration; compiler-rt does direct 24-bit
restoring binary long division on the mantissas, no reciprocal, no
multiply. This issue is a **discussion**, not a bug report or a patch:
NR is a legitimate, correct design choice that simply loses to direct
long division on the Z80 specifically, because the Z80 has no hardware
multiplier -- each NR iteration's cost (several float multiplies) is
disproportionate to what a fixed 24-step compare/subtract/shift loop
costs on this ISA.

## Why: algorithm, not a bug

**math32** (`libsrc/math/float/math32/asm/z80/f32_fsdiv.asm`,
`m32_fsinv_fastcall`) computes the reciprocal `1/b` via a degree-2
polynomial seed (`X0 = 140/33 + (-64/11 + 256/99*D')*D'`) followed by two
refinement steps `X := X + X*(1 - D'*X)`. Counting the seed, that's
roughly 8 full float multiplies + 7 float adds worth of sub-calls
(`m32_fsmul24x32`/`m32_fsmul32x32`/`m32_fsadd24x32`/`m32_fsadd32x32`),
plus a final multiply by the dividend to get the quotient. NR halves the
number of *iterations* needed for a given precision, which is the right
trade-off when multiplication is cheap (e.g. hardware FPU/multiplier) --
but on Z80, `m32_fsmul` is itself a multi-byte shift-and-add routine, not
a single instruction, so "fewer iterations, each an expensive multiply"
does not win here.

compiler-rt's z80 `__divsf3` (`compiler-rt/lib/builtins/z80/divsf3.asm`,
LLVM project, permissively licensed -- see licensing note below) instead
unpacks the 24-bit mantissas and runs a fixed 24-iteration
compare-subtract-shift loop (`djnz`-driven) -- no multiply at all. More
iterations, but each one is cheap.

This is also the reason for a known 1-ULP rounding difference between the
two approaches (e.g. on `-3/3`): NR's accumulated rounding differs
slightly from a direct divider's exact guard/round/sticky-bit tracking.
Neither is "wrong"; they're different valid roundings arrived at via
different paths.

## Reproducible benchmark (attached)

To let this be checked independently rather than taken on faith, **three**
builds are attached (two rebuildable by anyone with just z88dk, one
pre-built since it needs an llvm-z80 toolchain most z88dk maintainers
won't have):

- [`divtest.c`](https://github.com/ravn/z88dk/blob/02ebba1d3e427834376d52407145bb84aee8dfc3/tasks/upstream-div-discussion-2026-08-01/divtest.c) -- shared test source for the first two builds (N=2000-call
  loop, fixed operands `3.14159f / 2.71828f` every iteration), selected
  via a compile-time `-DUSE_MATH32` switch so the two builds differ only
  in which division routine is exercised.
- [`directdiv.h`](https://github.com/ravn/z88dk/blob/02ebba1d3e427834376d52407145bb84aee8dfc3/tasks/upstream-div-discussion-2026-08-01/directdiv.h) -- a portable-C reference implementation of the "direct
  restoring division" family (unpack sign/exponent/24-bit mantissa,
  24-iteration compare-subtract-shift loop, round, repack) -- structurally
  the same approach as compiler-rt's `__divsf3`, but written as plain
  portable C rather than hand-tuned Z80 assembly, precisely so it can be
  built and inspected with nothing but a stock z88dk/SDCC toolchain (no
  llvm-z80/clang involved).
- [`compilerrt_div_test.c`](https://github.com/ravn/z88dk/blob/02ebba1d3e427834376d52407145bb84aee8dfc3/tasks/upstream-div-discussion-2026-08-01/compilerrt_div_test.c) -- a freestanding test harness (same operands,
  same N=2000 loop) linked directly against LLVM compiler-rt's real z80
  `__divsf3` assembly implementation -- the actual hand-tuned routine,
  not the portable-C reference above. Build commands are in the file's
  header comment; **also attached pre-built** ([`compilerrt_div_test.bin`](https://github.com/ravn/z88dk/blob/02ebba1d3e427834376d52407145bb84aee8dfc3/tasks/upstream-div-discussion-2026-08-01/compilerrt_div_test.bin)
  + `.elf`, produced on my machine from this exact source) so it can be
  run/inspected without needing an llvm-z80 build at all.
- [`build_and_measure.sh`](https://github.com/ravn/z88dk/blob/02ebba1d3e427834376d52407145bb84aee8dfc3/tasks/upstream-div-discussion-2026-08-01/build_and_measure.sh) -- reproduces the first two builds from a clean
  checkout unconditionally, and the third if `LLVM_Z80_BUILD` happens to
  be set:
  ```sh
  PATH=<z88dk>/bin:$PATH ZCCCFG=<z88dk>/lib/config ./build_and_measure.sh
  ```

### Results (N=2000 divisions, z88dk-ticks, reproducible via the script above)

| build | implementation | total T-states | T-states/call |
|---|---|---|---|
| [`math32_div_test`](https://github.com/ravn/z88dk/blob/02ebba1d3e427834376d52407145bb84aee8dfc3/tasks/upstream-div-discussion-2026-08-01/MATH32_DIV_TEST.COM) | math32 `m32_fsdiv` (Newton-Raphson reciprocal) | 48,980,572 | 24,490.3 |
| [`directdiv_test`](https://github.com/ravn/z88dk/blob/02ebba1d3e427834376d52407145bb84aee8dfc3/tasks/upstream-div-discussion-2026-08-01/DIRECTDIV_TEST.COM) | direct restoring division, portable C reference | 33,642,572 | 16,821.3 |
| [`compilerrt_div_test`](https://github.com/ravn/z88dk/blob/02ebba1d3e427834376d52407145bb84aee8dfc3/tasks/upstream-div-discussion-2026-08-01/compilerrt_div_test.bin) | compiler-rt `__divsf3`, real hand-tuned Z80 asm | 21,594,032 | 10,797.0 |

- The **naive C reference** of the same algorithm is already ~1.46x
  faster than math32 -- a conservative lower bound, since it has none of
  the register-allocation tuning a hand-written routine would.
- The **real compiler-rt implementation** (same algorithm family, hand
  optimized) is ~2.27x faster than math32 -- consistent with the
  independent measurement in the llvmz80 bridge work
  (`z88dk/libsrc/l/llvmz80/MATH32_BRIDGE.md` Sec. 5, itself reproducible
  via `z88dk/test/clang/bench_math32_vs_compilerrt.sh`).
- The gap between the naive-C row and the real-asm row (16,821 vs.
  10,797) is the realistic remaining headroom a hand-tuned z80 asm
  implementation of the same algorithm could still capture beyond what
  this portable-C reference demonstrates.

Confirmed deterministic: rerunning [`build_and_measure.sh`](https://github.com/ravn/z88dk/blob/02ebba1d3e427834376d52407145bb84aee8dfc3/tasks/upstream-div-discussion-2026-08-01/build_and_measure.sh) (both with and
without `LLVM_Z80_BUILD` set) reproduced identical T-state totals across
repeated runs on the same machine.

## What this issue is (and isn't) asking

This is a **discussion starter**, not a proposed patch: is it worth
z88dk considering a direct-division implementation for `m32_fsdiv` on
z80/z180/z80n (either replacing the current NR approach, or offering it
as a build-time alternative)? The trade-off is real -- NR is correct and
reasonable in the abstract, and there may be reasons (code size, other
targets sharing this routine, historical compatibility) to keep it as-is
-- but the timing gap seemed large enough, and easy enough to demonstrate
reproducibly, to be worth raising.

## Licensing note

compiler-rt's z80 `divsf3.asm` (the routine linked into
[`compilerrt_div_test.bin`](https://github.com/ravn/z88dk/blob/02ebba1d3e427834376d52407145bb84aee8dfc3/tasks/upstream-div-discussion-2026-08-01/compilerrt_div_test.bin)) is triple-licensed `Zlib OR Apache-2.0 WITH
LLVM-exception OR MIT` (see the SPDX header in the file) -- all three are
permissive with no copyleft/share-alike obligation, so if z88dk
maintainers want to look at or adapt that algorithm/implementation
directly, that's available under any of those three licenses (attribution
notice would need to be retained, per each license's terms). The source
itself is not attached here since the discussion is about the algorithm
choice, not a proposed direct code lift, but it can be provided on
request, and the compiled binary linked against it ([`compilerrt_div_test.bin`](https://github.com/ravn/z88dk/blob/02ebba1d3e427834376d52407145bb84aee8dfc3/tasks/upstream-div-discussion-2026-08-01/compilerrt_div_test.bin)/`.elf`)
is attached for the timing comparison.

## Attached files

- [`divtest.c`](https://github.com/ravn/z88dk/blob/02ebba1d3e427834376d52407145bb84aee8dfc3/tasks/upstream-div-discussion-2026-08-01/divtest.c), [`directdiv.h`](https://github.com/ravn/z88dk/blob/02ebba1d3e427834376d52407145bb84aee8dfc3/tasks/upstream-div-discussion-2026-08-01/directdiv.h) -- math32 vs. direct-division reference,
  rebuildable with z88dk alone.
- [`compilerrt_div_test.c`](https://github.com/ravn/z88dk/blob/02ebba1d3e427834376d52407145bb84aee8dfc3/tasks/upstream-div-discussion-2026-08-01/compilerrt_div_test.c) -- pure compiler-rt harness source (needs
  llvm-z80 to rebuild).
- [`compilerrt_div_test.bin`](https://github.com/ravn/z88dk/blob/02ebba1d3e427834376d52407145bb84aee8dfc3/tasks/upstream-div-discussion-2026-08-01/compilerrt_div_test.bin), [`compilerrt_div_test.elf`](https://github.com/ravn/z88dk/blob/02ebba1d3e427834376d52407145bb84aee8dfc3/tasks/upstream-div-discussion-2026-08-01/compilerrt_div_test.elf) -- pre-built from
  the above on my machine, so the compiler-rt comparison can be run/
  inspected without an llvm-z80 toolchain.
- [`build_and_measure.sh`](https://github.com/ravn/z88dk/blob/02ebba1d3e427834376d52407145bb84aee8dfc3/tasks/upstream-div-discussion-2026-08-01/build_and_measure.sh) -- reproduces all of the above (first two
  unconditionally, the third if `LLVM_Z80_BUILD` is set).

feilipu:
# Discussion: Restoring fsdiv vs Newton–Raphson (math16 / math32)

Results from an optimised restoring fsdiv (implemented via l_fast_long_udiv example with reference to llvm-z80).

There is a substantial win in Whetstone, taking the result to over __11 kWIPS__!
Best result by a long way. But that cost comes with some extra code.

So not sure if this should be included, or not. For discussion.

Also, if the prior 1/n direct `fsinv `call is to be reversed, then __sccz80 will need to be adjusted__.

## Setup

| Item | Value |
|------|-------|
| Tree | `/home/phillip/Z80/z88dk` |
| **NEW** | restoring `fsdiv` / `asm_f16_div` (working tree) |
| **OLD** | NR inv×mul trampoline from `067aec9b70` (`18a723d^`) |
| Method | same `zcc +test` TIMER recipes; only div objects swapped; libs rebuild + reinstall |
| Sources | `support/benchmarks/*/z88dk-classic/*.c` |
| Timer | `z88dk-ticks -start TIMER_START -end TIMER_STOP` (8085: `-m8085` before bin) |
| Counter | `999999999999` |
| Date | 2026-08-08 |

### Recipes (classic sccz80)

```bash
# math32 whetstone
zcc +test -vn -DSTATIC -DTIMER -D__Z88DK -O2 z88dk-classic/whetstone.c --math32 -m -lndos
zcc +test -clib=8085 … -O2 … --math32 -m -lndos

# math32 n-body
zcc +test … -O2 z88dk-classic/n-body.c --math32 -m -lndos

# math32 / math16 spectral-norm / n-body math16
zcc +test … -O3 --opt-code-speed=inlineints z88dk-classic/… --math32|--math16 -m -lndos
# math16 8085 also: -lmath32_8085
```

## Summary (TIMER ticks; lower is faster)

| Bench | CPU | Lib | Flags | OLD (NR) | NEW (restoring) | Δ ticks | Δ % | OLD size | NEW size |
|-------|-----|-----|-------|---------:|----------------:|--------:|----:|---------:|---------:|
| whetstone | z80 | math32 | `-O2` | 529_019_818 | 363_261_050 | +165_758_768 | +31.33% | 9_397 | 10_170 |
| whetstone | 8085 | math32 | `-O2` | 1_062_512_692 | 766_999_051 | +295_513_641 | +27.81% | 11_070 | 11_767 |
| n-body | z80 | math32 | `-O2` | 920_897_733 | 920_897_733 | +0 | +0.00% | 5_446 | 6_219 |
| n-body | 8085 | math32 | `-O2` | 1_855_235_630 | 1_855_235_630 | +0 | +0.00% | 6_570 | 7_267 |
| n-body | z80 | math16 | `-O3 --opt-code-speed=inlineints` | 295_597_764 | 295_597_764 | +0 | +0.00% | 4_080 | 4_303 |
| n-body | 8085 | math16 | `-O3 --opt-code-speed=inlineints` | 344_362_740 | 344_362_740 | +0 | +0.00% | 3_906 | 4_089 |
| spectral-norm | z80 | math32 | `-O3 --opt-code-speed=inlineints` | 9_004_174_578 | 9_004_159_458 | +15_120 | +0.00% | 5_316 | 6_089 |
| spectral-norm | 8085 | math32 | `-O3 --opt-code-speed=inlineints` | 17_756_492_837 | 17_756_461_279 | +31_558 | +0.00% | 6_389 | 7_086 |
| spectral-norm | z80 | math16 | `-O3 --opt-code-speed=inlineints` | 4_726_369_202 | 4_726_362_472 | +6_730 | +0.00% | 3_661 | 3_884 |
| spectral-norm | 8085 | math16 | `-O3 --opt-code-speed=inlineints` | 5_834_341_846 | 5_834_333_707 | +8_139 | +0.00% | 3_680 | 4_012 |

Positive Δ% = faster (fewer ticks).

## Whetstone (math32) — primary win

Whetstone is divide-heavy; restoring replaces NR inv + mul.

| CPU | OLD ticks | NEW ticks | Speedup | OLD KWIPS@4MHz | NEW KWIPS@4MHz |
|-----|----------:|----------:|--------:|---------------:|---------------:|
| z80 | 529_019_818 | 363_261_050 | **1.46×** | 7.5612 | **11.0114** |
| 8085 | 1_062_512_692 | 766_999_051 | **1.39×** | 3.7647 | **5.2151** |

KWIPS = `1000 / (ticks / 4e6)` = `4e9 / ticks` (same formula as classic readmes).

### Link proof (whetstone math32)

- NEW maps reference `m32_fsdiv` / `m32_fsdiv_callee` from restoring `f32_fsdiv.asm`.
- OLD maps reference the NR trampoline + `m32_fsinv_fastcall` + mul.
- Binary sizes rise with restoring as expected.

## n-body and spectral-norm

These workloads are dominated by **mul / add / sqrt** (and math16 **invsqrt**), not general divide.

| Observation | Detail |
|-------------|--------|
| **n-body math32 z80** | OLD and NEW TIMER ticks **identical** |
| **n-body math32 8085** | OLD and NEW TIMER ticks **identical** |
| **n-body math16 z80** | OLD and NEW TIMER ticks **identical** |
| **n-body math16 8085** | OLD and NEW TIMER ticks **identical** |
| **spectral-norm** | See summary table (often ~0%) |
| Size | Restoring grows math32 binaries; math16 z80 often unchanged |

Conclusion: restoring `fsdiv` is a large win where `/` is hot (whetstone), and is essentially **neutral** on n-body / spectral-norm TIMER scores.

## Raw result dump

```
# old
nbody_8085_math16: ticks=344362740 size=3906
nbody_8085_math32: ticks=1855235630 size=6570
nbody_z80_math16: ticks=295597764 size=4080
nbody_z80_math32: ticks=920897733 size=5446
spectral_8085_math16: ticks=5834341846 size=3680
spectral_8085_math32: ticks=17756492837 size=6389
spectral_z80_math16: ticks=4726369202 size=3661
spectral_z80_math32: ticks=9004174578 size=5316
whet_8085_math32: ticks=1062512692 size=11070
whet_z80_math32: ticks=529019818 size=9397

# new
nbody_8085_math16: ticks=344362740 size=4089
nbody_8085_math32: ticks=1855235630 size=7267
nbody_z80_math16: ticks=295597764 size=4303
nbody_z80_math32: ticks=920897733 size=6219
spectral_8085_math16: ticks=5834333707 size=4012
spectral_8085_math32: ticks=17756461279 size=7086
spectral_z80_math16: ticks=4726362472 size=3884
spectral_z80_math32: ticks=9004159458 size=6089
whet_8085_math32: ticks=766999051 size=11767
whet_z80_math32: ticks=363261050 size=10170
```

## Notes

1. Only the four div sources were swapped between OLD and NEW; rest of tree identical.
2. After the OLD pass, restoring sources were copied back and libs rebuilt.
3. Published wiki Jul 2026 math32 whet (~524.7M z80 / ~1.14B 8085) aligns with this OLD NR band.
4. No math16 whetstone classic TIMER recipe; math16 covered via n-body + spectral-norm.


feilipu:
## Microbench: `1/n` via restoring `fsdiv` vs `fsinv` (math32)

**Question:** for a reciprocal `1/n`, is restoring division faster than the dedicated Newton–Raphson `inv` / `m32_fsinv_fastcall` path?

**Answer:** **yes on both z80 and 8085.** Restoring `one / x` (real `m32_fsdiv`) is about **2.8×** (z80) / **2.3×** (8085) faster than `inv(x)` when net of the shared empty loop.

### Method

| Item | Value |
|------|-------|
| Recipe | `zcc +test [-clib=8085] -vn -DSTATIC -DTIMER -D__Z88DK -O2 -DN=2000 … --math32 -m -lndos` |
| Timer | `z88dk-ticks` `TIMER_START`…`TIMER_STOP`, counter `999999999999` (8085: `-m8085` before bin) |
| N | 2000 |
| Inputs | rotating finite values in ~\[0.625, 1.875\] |
| **div** | `do_div(one, x)` → `return a / b` with `static float one = 1.0f` → **`m32_fsdiv`** |
| **inv** | `do_inv(x)` → `return inv(x)` → **`m32_fsinv_fastcall`** |
| **empty** | same loop, `r = x` then `sink += r` (shared float-add + loop cost) |
| net/call | `(mode_ticks − empty_ticks) / N` |
| Date | 2026-08-08 |

**Compiler note:** a literal `1.0f / x` is rewritten by sccz80 to `inv(x)` / `l_f32_invf`, so both binaries would hit **only** `fsinv` and match. The microbench uses a non-constant numerator (`static float one = 1.0f`) and a small C wrapper so the map really contains the restoring divider (`m32_fsdiv` / `div_body`, …).

### Results (math32 restoring tree)

| CPU | Mode | TIMER total | T/call (total) | T/call (net of empty) | vs inv (net) |
|-----|------|------------:|---------------:|----------------------:|-------------:|
| z80 | empty | 3_128_908 | 1_564 | — | — |
| z80 | **div** `one/x` (restoring) | **15_541_920** | 7_771 | **6_207** | **1.00×** (baseline) |
| z80 | **inv** `inv(x)` (NR) | 37_367_420 | 18_684 | 17_119 | **2.76× slower** |
| 8085 | empty | 5_685_442 | 2_843 | — | — |
| 8085 | **div** `one/x` (restoring) | **36_848_814** | 18_424 | **15_582** | **1.00×** (baseline) |
| 8085 | **inv** `inv(x)` (NR) | 76_275_588 | 38_138 | 35_295 | **2.27× slower** |
| z80n | empty | 3_128_908 | 1_564 | — | — |
| z80n | **div** `one/x` (restoring) | **15_541_920** | 7_771 | **6_207** | **1.00×** (baseline) |
| z80n | **inv** `inv(x)` (NR + `mul de`) | 24_829_670 | 12_415 | 10_850 | **1.75× slower** |
| z180 | empty | 2_832_583 | 1_416 | — | — |
| z180 | **div** `one/x` (restoring) | **13_164_254** | 6_582 | **5_166** | **1.00×** (baseline) |
| z180 | **inv** `inv(x)` (NR + `mlt`) | 23_319_754 | 11_660 | 10_244 | **1.98× slower** |

| CPU | Restoring div vs NR inv | Speedup | HW mul in path? |
|-----|-------------------------|--------:|-----------------|
| z80 | −63.7% ticks (net) | **2.76×** | no (soft 8×8 in mul helpers) |
| 8085 | −55.9% ticks (net) | **2.27×** | no |
| z80n | −42.8% ticks (net) | **1.75×** | **yes** — inv uses `f32_z80n_mulu_*` / `mul de` |
| z180 | −49.6% ticks (net) | **1.98×** | **yes** — inv uses `f32_z180_mulu_*` / `mlt` |

**Takeaways**

1. Restoring **`fsdiv` still beats `fsinv` for `1/n` on all four CPUs**, including those with hardware multiply.
2. HW mul **narrows** the gap (z80 2.76× → z80n 1.75×) because NR **`fsinv` is mul-heavy**; restoring **`fsdiv` does not use the 24×32/32×32 mul cores**, so z80n div matches plain z80 div tick-for-tick.
3. z180 shows a slightly cheaper restoring loop in TIMER (CPU model timings differ from plain z80 even without mul in the div path).

So for **explicit** `1/n`, prefer **`a / n` with a runtime `1.0`** (or any general divide) over **`inv(n)`** with the current restoring `fsdiv` and NR `fsinv`.  
`inv` remains useful when the reciprocal is needed as a building block for other NR-based cores (and when the compiler already lowers `1.0/x` to it — that path does **not** get the restoring win until codegen stops rewriting, or until `fsinv` itself is replaced).

ravn:
Nice work!

I cannot relate whetstone numbers - is 11 kWIPS very impressive?

feilipu:
> I cannot relate whetstone numbers - is 11 kWIPS very impressive?

Historically unprecedented. So yes. 👍 
Previous best from any compiler was just over 7 kWIPS (all figures assume 4MHz z80).
[See benchmarks here](https://github.com/z88dk/z88dk/wiki/Benchmarks#whetstone-12).

So I'll wait until the sccz80 compiler is reverted to the original situation (before modified for issuing `fsinv`) before finalising the actual numbers. But, it is likely to be a bit more than 11 kWIPS.

ravn:
If the previous record was 7 kWIPS then 11 is very, very impressive.  Congratulations  🙂 

The linked page says that the package is available for download, but not where, and I'd like to give it a shot to see how math32 performs with llvmz80 🙂

suborb:
@feilipu I think that's the change you wanted?

feilipu:
> I think that's the change you wanted?

Yes. Thanks.
I'll do some updates around benchmark comments and special paths too.
Leaving the 80cc `irbuild.c` updates to you too.

Running some new benchmarks tomorrow, once this is merged.

feilipu:
A few quick tests, for good measure. Definitely worth doing.
Thanks for the push to get this done @ravn 🙏 

### z80 classic sccz80 / math32 — TIMER remeasure

After restoring `fsdiv` (this PR) and removing the sccz80 / spectral-norm `1.0/n` → `inv` special path.

**Recipes** (classic `+test`, `z88dk-ticks -start TIMER_START -end TIMER_STOP -counter 999999999999`):

- whetstone / n-body: `-O2 -DSTATIC -DTIMER -D__Z88DK --math32 -m -lndos`
- spectral-norm: `-O3 --opt-code-speed=inlineints -DSTATIC -DTIMER -D__Z88DK --math32 -m -lndos`

**Columns**

| Column | Meaning |
|--------|---------|
| OLD NR | Prior `fsinv`×mul divide |
| Restoring (pre-sccz80) | Restoring `fsdiv` only; sccz80 still rewrote `1.
| Final HEAD | Restoring `fsdiv` + sccz80 no longer rewrites IEEE/f16 `1.0/

**%** = `(OLD − ticks) / OLD × 100` vs OLD NR (higher is faster).
**Total %** = full improvement OLD NR → Final HEAD.

| Bench | OLD NR | Restoring (pre-sccz80) | % vs OLD | Final HEAD | Total %
|-------|-------:|-----------------------:|---------:|-----------:|--------
| whetstone | 529019818 | 363261050 | +31.33% | 363261050 | **+31.33%** |
| n-body | 920897733 | 920897733 | +0.00% | 790941952 | **+14.11%** |
| spectral-norm | 9004174578 | 9004159458 | +0.00% | 4250214704 | **+52.80%**

**Whetstone KWIPS @ 4 MHz:** OLD 7.56 → Final HEAD 11.01 (`4e9 / ticks`).

**Why n-body / spectral move only in Final HEAD**

- **Whetstone:** hot path is general `a/b` → already `fsdiv` after restoring; no change from dropping the `1.0/x` rewrite.
- **n-body:** classic uses `1.0/sqrt(...)`; sccz80 used to lower that to NR `inv(sqrt(...))`; HEAD uses restoring `div(1.0, sqrt(...))`.
- **spectral-norm:** was explicitly calling `inv()` under `__MATH_MATH32`; HEAD uses `1.0/…` → restoring `fsdiv` (~2.12× vs restoring column alone).

**spectral-norm correctness (N=100, math32)**

| Build | Printed result | Notes |
|-------|----------------|-------|
| Final HEAD (`1.0/…`) | 1.274219155 | abs err ≈ 5e-6 vs ref 1.274224153 |
| Explicit `inv()` (same flags) | 1.274219155 | identical print; ~9.00e9 ti

Speedup is fewer cycles for the same float32 answer, not a wrong/short run.

suborb:
If I'm reading correctly, the size delta is 700 bytes? That's a pretty hefty size bump - can you unroll less?

feilipu:
It is very expensive special case (zeros, nan, etc) handling that takes bytes. I've pushed on that a little, but will push again.

But, I also just took a lot of bloat out of the normalising code in #3053, so it should balance out a little.

feilipu:
## Size ledger — restoring `fsdiv` cost vs normalise savings

Assembled `code_fp_math32` / `code_fp_math16` only.

| Platform | Normalize Δ | Div Δ | **Net (norm + div modules)** |
|----------|------------:|------:|-----------------------------:|
| **math32 z80** | **−655** | **+580** | **−75 B** |
| **math32 8085** | **−619** | **+690** | **+71 B** |
| **math16 z80** | 0 | **+203** | **+203 B** |
| **math16 8085** | 0 | **+302** | **+302 B** |

### How to read it

1. **z80 math32:** the normalize campaign more than pays for restoring `fsdiv` on a pure object tally (**−75 B** net).
2. **8085 math32:** nearly flat (**+71 B** net).
3. **math16:** only div grew; no parallel normalize offset → **+200–300 B** per CPU.
4. **Not included:** NR div’s old pull of mul / add32 / helpers when `/` went through `inv`. Restoring `fsdiv` is self-contained, so **div-only** apps can still win on total link size even when the `fsdiv` object alone is larger.

ravn:
> Den 8. aug. 2026 kl. 20.01 skrev Phillip Stevens ***@***.***>:
> 
> 
> feilipu
>  left a comment 
> (z88dk/z88dk#3054)
>  <https://github.com/z88dk/z88dk/pull/3054#issuecomment-5227399422>
> A few quick tests, for good measure. Definitely worth doing.
> Thanks for the push to get this done @ravn <https://github.com/ravn> 🙏
> 
You’re welcome 🙂  Hope you had fun!
> Whetstone KWIPS @ 4 MHz: OLD 7.56 → Final HEAD 11.01 (4e9 / ticks).
> 
That is an impressive leap just by improving division a bit 🙂 

I tried getting whetstones to work with dcc resulting in kwips around 9 and llvm giving around 14 using —math 32.   It will be very interesting to see if you can replicate these numbers later 🙂

Issue #3043, opened 2026-08-02
float32 fsadd alignment implementation: can't figure out some details about implementation


ped7g:
I guess this one is mostly for @feilipu :

I'm working on squeezing few more T-states out of the float32 addition implementation (nothing significant, but I hope it will *add* up in the end... pun intended).

And I did notice two peculiarities which I either can't understand or smell like borderline bugs.

## Q1:

https://github.com/z88dk/z88dk/blob/9f5773a7b23261b927b7db23a9faa9016e739433/libsrc/math/float/math32/asm/z80/d32_fsadd.asm#L354-L358

This exit through `m32_fsnormalize` sets carry in F' with comments suggesting as if there was some alternate path in normalize:

https://github.com/z88dk/z88dk/blob/9f5773a7b23261b927b7db23a9faa9016e739433/libsrc/math/float/math32/asm/z80/d32_fsnormalize.asm#L26-L27

But from the implementation of normalize I can't figure out how the F' matters, except it is being returned with final value, so technically that `scf` in fsadd will make the addition/subtraction result (requiring full normalization) to return with carry set, which in other exit paths means "error".

I don't think it was meant as "error" just because normalization was done, and it's rather based on some historical implementation which did have different code paths based on F' carry set/unset upon call?

So IMHO fsadd exiting through full normalize should instead clear F' carry to not report "error" (unless zeroing result in case of underflow counts as "error", but  that should be IMHO enforced in `.normzero` branch, not by caller).

Am I missing or misunderstanding something?

(if you agree carry should be cleared, I can adjust it in my optimized version)

## Q2:

https://github.com/z88dk/z88dk/blob/9f5773a7b23261b927b7db23a9faa9016e739433/libsrc/math/float/math32/asm/z80/d32_fsadd.asm#L196-L221

when shifting smaller number's mantissa by 16 bits during aligning the code says it discards collected lost bits in A so far because they are too remote to shift by 16, and instead it does focus on content of D (from LDE) which the highest byte to be gone -> so far this all makes sense to me.

So after `ld a,d` it does `or a,h` -> here I'm lost, the `h` meanwhile could have accumulated some even less significant lost bits (from shift by 2 and 4 places due to those `rr hl` used in the unrolled chains). So if the lost bits in A are discarded, are those other lost bits from H intentionally mixed in here?

IMHO it should have been just `or a` to test the D itself, discarding any other lost bits from earlier.
[...]

PR #3056, opened 2026-08-08
math32 - further diet of normalisation (f32, z80)

ped7g:
From 152 to ~~96~~ ~~85~~ ~~53~~ 55 bytes.
Also marginally faster, no code path is slower.

Related to #3043

### DIFF
--- a/libsrc/math/float/math32/asm/z80/f32_fsnormalize.asm
+++ b/libsrc/math/float/math32/asm/z80/f32_fsnormalize.asm
@@ -1,5 +1,6 @@
 ;
 ;  feilipu, 2026 August
+;  ped7g, 2026 August
 ;
 ;  This Source Code Form is subject to the terms of the Mozilla Public
 ;  License, v. 2.0. If a copy of the MPL was not distributed with this
@@ -9,7 +10,7 @@
 ; m32_fsnormalize - z80 / z180 / z80n normalisation
 ;-------------------------------------------------------------------------
 ;
-;  unpacked: h==0; mantissa=lde, sign in b, exponent in c
+;  unpacked: h==0 (discarded); mantissa=lde, sign in b, exponent in c
 ;
 ;  ex de,hl → E:HL; D = total left-shift count (byte align + residual).
 ;  Sign stays in B (no push).  Residual: jp m tree, add hl,hl / rl e.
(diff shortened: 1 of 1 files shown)


ped7g:
Looking at the final resulting code...

This must be "wrong", the bitwalk and shifts are doing the same work on bits 16:23 twice and it should be possible to merge it with everything else kept "as is", no extra cost, so that should be another minus `rl e` per every shift?

Switching this to "draft", I will take a look tomorrow, bed time...

EDIT: ah, it's the initial extra `add hl,hl` to shift implicit 1 out and detect it, which will hurt a bit (to revert)
Still worth a double-check, maybe I will figure out a way to remove at least some of it, maybe going back to `jp m` test instead, it's probably slower on average because `jr` is faster to fail, but if it removes extra `rl e`, it should be still gain in total.

feilipu:
Yes there are some things to fix without any loss of function.

Background: I was trying to write something that was common for 8085 and z80, but the different IEEE packing shifts blew up that idea. I've carelessly forgotten to remove some then unnecessary baggage from the z80 solution.

I've already pulled the safe things into master. Thanks for identifying them.

The restoring division function I added added initially over 700 bytes. But now it is down under 600. So with your savings here adding another 50 bytes it is a big win. 🥳

[See results](https://github.com/z88dk/z88dk/pull/3054#issuecomment-5229605192).

ped7g:
what do you mean "safe" things to master? This code was correctly working, it just can be further optimized, but there was nothing wrong about it in terms of how it works?

If you have spot some bug, surely let me know.

feilipu:
Just your comment and setting to pr draft. A critical review of the savings found some things in pr safe, no question, and others needed further thought (as you rightly pointed out). So skipped those "unsafe", for now.

ped7g:
Smells like quite some AI slop is going on here... :shrug: 

Well, I will prepare final version of my code anyway, but this cherry-picking of code changes based on "safe"/"unsafe" and the resulting code on master branch is not convincing to me that Phillip personally was involved, I would appreciate if responses would be clearly marked as AI-only/well-reviewed, so I have better idea when to just safely ignore some changes as usual noise.

feilipu:
Yeah. No. The code and algorithm is all mine.

So a tool provided me a ranked list based on byte values, from which I picked things I could see are obviously "safe" and pick things to get pushed today.

Anyway another 50 bytes saved takes us into positive balance with the new `fsdiv` code. Excellent!
👍

ped7g:
@feilipu what's your opinion on keeping "noise" bits when normalizing by byte alignment?

The code so far does zeroing of those, but it feels to me like it is mathematically/IEEE valid to set them to anything (as rounding error), but I haven't read IEEE spec, so I'm more like guessing.

Example:
+$0000:4582 comes into normalise with exp=100, so the byte-align code would turn it into $0045:8200 and exp -8, then further bit shifts to get +$008B:0400 exp -9 (writing it from head, so hopefully I got it right :D ), positioning the implicit 1 into 24th bit.

How about not bothering to clear lowest 8 bits, ie. the byte align would result in $0045:8282 and then bit shifts would result in $008B0504  (so +$104 in lowest mantissa bits)?

It would save only about two-three bytes and like -4T per byte-alignment code path, so just as performance optimization it's very insignificant, but it gut-feels to me like it can also somewhat even out lost precision of very small numbers. My intuition is suggesting it would be either neutral or positive change, but my intuition may be just performance jerk looking forward to get even those -4T per code path, maybe there are hard data reasons why this is not a good idea at all.

ped7g:
Turns out it was really possible to merge shifting and bitwalking to do no duplicate work on the values.
And it saves another ~10 bytes and the source looks IMHO even more streamlined and simpler to grok.

ped7g:
ARGH ... one more optimization incoming... it's always one more thing... ridiculous, how it becomes obvious *after* I think I'm done with it. :)

feilipu:
Looks very good. Everything, except 0 shift path, is better and cheaper.
Can you have a look at the fast path (zero shifts) solution, please?
The hot path is quite a bit slower (12 ticks).

ped7g:
About 0 shift path - and that's still faster compared to original (152 bytes one, doing push/pop bit sign).

But I guess it's technically slower against current master, ok, I will take a look, if there's a good landing spot for the extra `jp m`, it should cost just +3 bytes to make it fastest case.

--------------

About fsadd... I'm planning to "exploit" the fact the normalize doesn't depend on h==0 and just discards it, so the fsadd code can then skip clearing of H reg. before jump.

So two questions are open right now feilipu:

1) zeroing or "noise" of made up low precision bits during byte alignment

2) is it ok internally for fsadd to call with H != 0, assuming the normalize code will discard the nonsense in H anyway

ped7g:
Also if you think the already-normalised path is lot hotter than other, it can be tested even earlier and have completely separate exit path doing specific pack (no exponent adjustment, etc).

feilipu:
> Also if you think the already-normalised path is lot hotter than other, it can be tested even earlier and have completely separate exit path doing specific pack (no exponent adjustment, etc).

It is enough to have a cheap Path 0 without adding too much more. A cheap `jp m` fast path (+~3 B) is welcome if it restores the original zero-shift cost.

> So two questions are open right now feilipu:
> zeroing or "noise" of made up low precision bits during byte alignment.

1. Byte-align low bits — keep zeros.

Left-shift normalise is exact rescale of the known mantissa. IEEE says new low bits must be zero unless a wider field already holds real bits below LDE. Filled “noise” is not IEEE rounding. Rounding uses lost bits on right-shift (our jam-sticky on add, RNE residual on mul/div). Fake LSBs can add bias and possibly false precision after cancel.

> is it ok internally for fsadd to call with H != 0, assuming the normalize code will discard the nonsense in H anyway

2. H != 0 into m32_fsnormalize — yes as do not care, with a contract.

Current code does `ex de,hl` then `ld d,0`, so entry H does not affect the packed result. You may skip a redundant `ld h,0` when sticky/overflow is already done and the 24-bit magnitude is only in LDE. Document that: "H is ignored on entry." Do not put extra precision in H. High overflow must be fixed before this path. Extra precision needs a real wider field (e.g. using fsnormalize32), not a silent discard.

ped7g:
Ok, this is now 55 bytes, already normalised value has hot-path and I removed the into-loop-entry point completely (mostly to make the source easier to follow while reading, performance wise it will be same on average with +-12T per particular path).

So this is now about ~1/3 of size from where I started. I didn't expect such gain, interesting, on first look it did seem to be like in 20-30 bytes to shave off... :)

funny coincidence - in my text editor this now fits exactly one page from `.m32_fsnormalize` to last `ret`

And now back to fsadd where I have only about -55B so far...

Obviously feel free to reuse any idea of this for the other files (8085 or normalise32), I like to focus on particular file one at time, but with some luck the LLM can maybe spot the pattern and suggest other files matching these ideas.

feilipu:
> with some luck the LLM can maybe spot the pattern and suggest...

I'll ask it in due course.

Actually the split walk and shift idea originally arises from my 8085 design using `add hl,hl` `rl de` to shift without moving bytes through `a`. So IDK if there is translation there, but I'll let it loose and see. Hope there's a win to find.

ped7g:
@feilipu back after "fresh look"

I think this can be still sped up on average, but it will cost some code size (probably like +15..30 bytes).

The idea is to do `cp 8` (or maybe just `cp 4`) on non-zero leading byte (before entering loop), and if CF=1 -> 5+ leading zero bits -> enter specific loop which will instead shift right (3 to 1 times right instead of 5 to 7 times left) and reshuffle the bytes afterward to get correct 24b result. Shifting right is slower and "lost bits" have to be recovered, but probably all of that can be achieved with the difference of 3 loops right vs 5 loops left (or at least 2 loops right vs 6 loops left) to make it on average slightly faster.

Should I take a look or the compact code size here has more value?

feilipu:
The way it is (or will be post CI gremlins), is best as it is small and quite fast for the common cases. There was the fast large digi inspired nybble based version I just ditched to make space. math32 is generally already "too big".

So I would not spend time on that path.

PR #3064, opened 2026-08-14
math32 - diet of z80/fsconv.asm

ped7g:
From 159 to 127 bytes and faster (I know there is one specific path being +5T slower, but everything else should be considerably faster, from 10+ to 100+ T states).

### DIFF
--- a/libsrc/math/float/math32/asm/z80/d32_fsconv.asm
+++ b/libsrc/math/float/math32/asm/z80/d32_fsconv.asm
@@ -8,6 +8,9 @@
 ;  feilipu, 2019 April
 ;  adapted for z80, z180, and z80n
 ;
+;  ped7g, 2026 August
+;  optimized for code size and performance
+;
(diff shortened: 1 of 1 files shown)


ped7g:
JFYI: I'm testing these optimizations outside of z88dk (I have custom test runner in sjasmplus, so I can iterate quickly and debug in my usual tools), luckily the syntax is almost same.

Anyway, point is, that there is some minuscule chance I messed something up while flipping it to z88dk form... I guess CI would catch that.

WRT to bit-results testing, the i8/u8/i16/u16 results were compared with original code for all possible values, i32/u32 were compared in loop with +19777 increment (ie. ~217k test values with all variants in upper 16b).

So if I somehow managed to break it, I will buy all maintainers at least one beer... :D

feilipu:
Have stress tested it already. No issues. Will merge once green (cygwin takes a minute, over 280 of them).

Thanks 🙏

PR #3066, opened 2026-08-15
80cc mandelbrot pr

feilipu:
Some minor patches resolving issues with the Mandelbrot benchmark. Affecting 80cc math32.
Benchmarks being updated now.


src/80cc/ir_lower_ops.inc.c
```diff
@@ -1914,6 +1914,29 @@ static int gen_shl(FILE *out, Func *f, const Op *op)
...
         }
+        /* Variable-count byte <<. The imm path below uses op->imm as
+           the count (0 here) and would emit <<0. Count in B, value in
+           A, `add a,a` loop — same portable body as const, no djnz. */
+        if (op->src[1] >= 0) {
+            int n = L.cmp_label_counter++;
+            int bc_live = (L.rs.bc >= 0);
+            if (!hl_has(op->src[1]))
+                load_to_hl(out, f, op->src[1]);
+            if (bc_live) emit(out, "push\tbc");
```
suborb: I think this needs to be:

  if (bc_live) emit_sp(out, 2, "push\tbc");

and similarly (-2) on the pop or the stack tracking offset is incorrect for the load_byte_to_a

feilipu: just a suggestion.

the issue was causing the last row of the Mandelbrot vs gold to be offset. and there was a related dereference problem. Let me see. I'll paste the text.

feilipu:
### 80cc PR 3066 — two codegen patches

Two commits on 80cc_mandelbrot_pr fix 80cc code generation for classic math32 Mandelbrot.

1. Variable-count byte << (c6ae24efc8)

Variable-count byte << (c6ae24efc8). Mandelbrot leftover packing uses byte_acc <<= (8 - w % 8). A variable count fell through to the immediate path, which used count 0. The leftover byte stayed unshifted.

gen_shl handled a byte result only when the count was immediate. A vreg count fell through to the immediate path. That path reads op->imm, which is 0. The compiler then emitted a shift by 0.

The leftover byte stayed at 0x0c. For w == 60 the packed value must be 0xc0.

The new path loads the count into B and the value into A. It then loops add a,an and decrements B. The loop is portable. It does not use djnz or the CB shift set. If BC holds a live value, the path saves BC first. A count of 0 skips the loop

The ca_char_shl_var test locks this. It expects HL = 0x00c0.

2. Byte deref through BC or DE (4f99e41975)

Byte deref through BC or DE (4f99e41975). return *p_local assigned BC as the pointer home but never loaded the pointer into BC. The assigned home is not the same as a live value in BC. The compiler then emitted ld a,(bc) with BC equal to 0.

The old path used vreg_in_pr_bc and vreg_in_pr_de. Those tests look at the allocator home, not at current tenancy. The compiler then emitted ld a,(bc) while BC was 0.

The new path uses bc_has and de_has. Those tests are true only when the pointer is in BC or DE now. If the pointer is not there, the path loads it into HL and reads (hl)..

src/80cc/ir_lower_ops.inc.c
suborb: Yeah, there's a bug there where the case wasn't handled. 

I was pointing out that the push/pop bypass the sp accounting so load_byte_to_a may load a from the wrong stack offset.

feilipu: yes. pushed the fix. thanks.

suborb:
Merging this since it's a bug fix.

PR #3072, opened 2026-08-18
80cc: Further work - 8085, 

suborb:
Mostly 8085 tweaks, but some generic stuff as well.

### 8085 codegen
  - `dsub` (`sub hl,bc`) for 16-bit subtract where BC is free to stage the subtrahend
  - LHLX/SHLX for long frame slots and for word derefs
  - an 808x word compare against a constant subtracts the immediate halves instead of staging DE
  - a fused long bitop keeps its operands in BC/DE and writes the result in place, telling its sink that BC already holds the low half
  - the DE park around a slot load (`push de / ld de,sp+N+2 / ld hl,(de) / pop de`) is dropped where nothing reads the pair
  - one copt rule: a constant long half before a slot store needs no `ex de,hl`

  ### 80cc, all CPUs
  - a branch to a bare `ret` becomes the `ret`
  - a store's address is built before a wide value, with the reorder skipped when the  address is a link-time constant
  - a store-then-reload of the same sp-relative word slot reduced to a store only
  - BC packing asks whether the candidate crosses a BC clobber rather than vetoing the whole function

  ### Bug fixes
  - `l_f32_f2sint` (mbf32) unwound the sign byte and the saved IX in the wrong order,
  - a static local's storage is in bss, not on the frame
  - a wide accumulator op is a call, so a stack park cannot span it
  - a LEA escapes rematerialisation when it is parked on the stack, not when the CPU is 8080

Unintended side effect is that md5 has gone on a diet and lost bytes and ticks.


libsrc/l/sccz80/7-8085/i32/l_long_div_0.asm
```diff
@@ -32,6 +45,9 @@ PUBLIC  l_long_div_0
     or      c
     ld      e,16                ; E = 16 if MSW half, else 0
     jp      nz,half_ready
+    IF DIV16_FASTPATH
```
suborb, line 48: @feilipu not sure how to handle this (And we need to replicate for 8080 and gbz80), however this 32/16 optimisation is pretty useful.

It's 86 bytes, so I'm unsure whether it's worth doing by default, or creating a `fastmath` equivalent.

feilipu: Leave it in, at least as a fast math equivalent. The fast path options take quite a few ticks out of the big benches, so worth doing.

For the gbz80 implementation, there are skills for that ;-)
I guess that can extend to 8080 too.

feilipu: Unrelated comment: on whether test before subtraction is required, rather than subtract, test and revert.

The stack traffic to test only is much less than needed to actually subtract, so the test first path is faster for 8085. Tested by dual implementation A/B analysis.

suborb: I'll leave it in for the moment and revisit for 8080/gbz80 later.

feilipu:
The msys2 build is failing here.
Something in the CI run is off.

suborb:
The PR build passed: https://github.com/z88dk/z88dk/actions/runs/32185379989

The merge into master failed: https://github.com/z88dk/z88dk/actions/runs/32231138703 with:

```
long_ir.c:2010: could not flush output (disk full ?)
```

It doesn't buy much space, but a make clean in libsrc before running the tests will claw back about 400mb.

I suspect ultimately we may need to change the build so it's pipelined a little. Build binaries + libs. 

This then allows us to choose not to build libraries on the windows machines and avoid running into a timeout.

Store them as a temporary pipeline artefact and then run the tests in a separate stage.

Edit:

* 14GB partition on D: shared with msys and the z88dk build
* I've pushed a `make -C libsrc clean` as a temporary bodge

PR #3073, opened 2026-08-19
zcc - multi compiler tool

feilipu:
## [SEE SPECIFICATION OF `zcc-multi` TOOL](https://github.com/z88dk/z88dk/blob/zcc_multi_compile/src/zcc-multi/zcc-compiler-multi.md)

Per-function mix of sccz80 and 80cc after copt. One stitched assembly file goes to the existing z80asm path.

This branch does not change 80cc. FatFs and FreeRTOS compile because master already has #3079.

## What this adds

| Piece | Role |
|-------|------|
| `-compiler=multi` | Compile one C translation unit with more than one compiler variant |
| `-compiler-metric=ticks` | Default. Pick the lower static T-state sum |
| `-compiler-metric=size` | Pick the smaller assembled body |
| `-compiler-multi-report=path` | Write a TSV of winners and scores |
| `z88dk-zcc-multi` | Parse, measure, select, stitch |
| `test/suites/zcc-multi/` | Stitch, 8085 two-variant, helpers, leftover `defc`, fallback, loops, `ldir`, CPU extras, sdcc ABI reject |

zcc still preprocesses, compiles, and runs copt. The new tool only replaces the single-compiler `.asm` that z80asm already accepts.

## Variants

The variant set follows 80cc CPU capability. There is no private CPU list.

1. `sccz80` — default stack locals. Multi does not pass `-frameix`.
2. `80cc-sp` — 80cc with `-fomit-frame-pointer`. Always.
3. `80cc-fp` — 80cc with `-fframe-pointer`. Only when 80cc treats the CPU as having IX.

No-IX CPUs (8080, 8085, gbz80) get two variants. Z80-family CPUs get three.

Each variant has its own preprocess. Benches that use `#ifdef __80CC` keep that path.

CPU flags come from `cpu_map[]` / `select_cpu()`. z80asm gets `--cpu=` and `--asm-flags=` from `CPU_MAP_TOOL_Z80ASM`, including extra flags such as `-IXIY`.

## Ticks metric

Size is the listing opcode bytes after the function label. The tool does not use listing line numbers.

Ticks come from the assembly source. `ticks_for_src` reads the mnemonic and the operands. It does not use a listing byte count. `ld hl,n` is 10 T. `ld a,n` is 7 T. `ld a,b` is 4 T.

One expansion rule covers pair copies, word rotates, `ex`, word loads, and CPU extras. If the chip has the op, use the native time. If z80asm expands the mnemonic to a short inline sequence, use the sum of those ops. If z80asm emits a helper `call` and the cost is documented, use that cost. Repeating block ops (`ldir` and friends) use `21 × BC` on every CPU. An unknown mnemonic or a helper with no documented cost sets fallback. 80cc writes the mnemonic. The assembler chooses the expansion. Names are Zilog on every model. Intel and Nintendo aliases are not matched.

Examples of that rule:

| Form | Score |
|------|-------|
| `ld` among `bc` / `de` / `hl` | 8 T (Z80-family, 8085, gbz80). 10 T (8080). 4 T (Rabbit) |
| `rl` / `rr` / `sra` on a pair | 16 T as two CB ops on Z80 / Z180 / Z80N / gbz80. Native where the chip has it (8085 `rl de` 10, `sra hl` 7. Rabbit `rr hl` 2) |
| `ld hl,(ix+d)` | 38 T on Z80 (two indexed loads). 11 T on Rabbit |
| gbz80 `ex de,hl` | 56 T (`push hl` / `push de` / `pop hl` / `pop de`) |
| gbz80 `ex (sp),hl` | 148 T (helper) |
| gbz80 `ld hl,sp+*` | 12 T |
| Rabbit `bool hl` | 2 T |
| Rabbit `add sp,n` | 4 T |

Documented extra-op bundle sums: 8085 97, gbz80 144, z180 94, z80n 175. `slp` and `jp (c)` stay fallback.

A backward `jr` / `jp` / `djnz` to a label in the same function is a loop. The span is multiplied only for a literal trip. Overlapping edges multiply (B × C). Load 0 wraps. An 8-bit load is 256 trips. A 16-bit pair load is 65536. `dec rr` does not set Z. The portable 16-bit test is `or` through A. 8085 `jp k` / `jp nk` after `dec rr` is a general 16-bit loop.

`ldir` / `cpir` and the other Z80 block repeats use `ld bc,N` when that load is visible. `ld bc,0` is 65536. Unknown BC uses `ZCCMULTI_BLOCK_REP` (2). The block-repeat pass owns that line score.

Classic helpers are a differential. Shared `l_*` / `d*` calls cancel. Extra sites add `(count[v] − min) × helper`. The loop product on the call line is in `count[v]`.

`jp (hl)` / `(ix)` / `(iy)`, `halt`, or an unknown mnemonic make that body unusable for ticks. If any present variant of a function is unusable, the pick is size among all bodies. The tool does not mix ticks of two usable bodies with size of a fallback body. `reason` is `fallback`. A function that exists in only one variant is still emitted (`reason` `only`).

## Stitch

Named file-scope data comes from sccz80 only.

Local labels (`i_N`, `L_f*_bb_*`) are rewritten per variant.

Optimiser and literal pools come only from variants that contributed a function.

Leftover copt `defc i_N = i_M` is kept only when the chain ends at a label in a selected body. The selector tracks the whole alias pool. There is no 256-entry cap.

IX mix is costed. An 80cc-fp caller must not call a selected sccz80 or 80cc-sp body. Elevate versus demote: cheaper legal fix. Tie keeps elevate.

`-v` prints one winner line per function and a `zcc-multi-summary:` block.

`-compiler-multi-report` on a `@lst` job overwrites the TSV once per `.c` file. The file that remains is the last translation unit. Use `-v` summaries for a multi-file job, or pass a distinct report path per file.

## Example

```text
zcc +test -compiler=multi -compiler-metric=ticks -v \
  -DSTATIC -DTIMER -D__Z88DK -O2 --opt-code-speed \
  fannkuch.c -o fannkuch.bin -lndos -m
```

Shape of `-v` lines:

```text
zcc-multi: _foo selected=80cc-fp sccz80=42/130 80cc-sp=40/125 80cc-fp=38/120
zcc-multi-summary: selected sccz80=0 80cc-sp=0 80cc-fp=1
zcc-multi-summary: ticks selected=120 sccz80=130 saved_vs_sccz80=10
```

Numbers are `bytes/static-ticks` after differential helper charge. Missing variant columns show `-`.

## Measured on this tree

Do not paste these TIMER numbers into published RESULT tables. The metric is a compile-time proxy.

### Classic `+test` TIMER (2026-08-22)

Same recipes as the original 19 Aug table. Baseline is sccz80 on this tree. Percent is `(sccz80 − multi) / sccz80 × 100`. Positive is faster.

| Bench | sccz80 | multi | Δ |
|-------|-------:|------:|--:|
| sieve | 4,769,143 | 4,769,143 | 0% |
| fannkuch | 83,264,076 | 69,691,670 | **+16.3%** |
| fasta | 123,137,988 | 150,201,804 | **−22.0%** |
| mandelbrot | 1,030,978,885 | 1,030,978,885 | 0% |
| n-body | 797,522,480 | 809,458,163 | −1.5% |
| pi | 1,992,348,390 | 1,997,952,293 | −0.28% |
| binary-trees | 149,258,200 | 142,510,225 | **+4.5%** |
| sorting | 44,474,249 | 38,866,533 | **+12.6%** |
| spectral-norm | 4,210,446,635 | 4,210,428,394 | 0.00% |
| whetstone | 361,981,590 | 359,722,820 | +0.62% |

sieve and mandelbrot stay all-sccz80. sorting is the same mix as 19 Aug. fannkuch still wins. The hot loop is 80cc-sp, not 80cc-fp.

fasta is a TIMER loss. Static ticks prefer 80cc-sp `_gen_random`. TIMER disagrees. Keep `-compiler=sccz80` on that file.

pi sccz80 TIMER on this tree is 1.99e9, not the 19 Aug 4.03e9 row. PRINTF digits match. Multi then picks 80cc-sp and is 0.28% slower than sccz80 on this tree.

### FatFs `ff.c`

Package recipe. No source fold.

```text
zcc +rc2014 -clib=new -compiler=multi -x -O2 --opt-code-speed=all --math32 @ff.lst
zcc +rc2014 -clib=new -m8085 -compiler=multi -x -O2 --opt-code-speed=all -D__DISABLE_BUILTIN --math32 @ff.lst
```

`@ff.lst` completes on z80, 8085, z180, z80n, gbz80, and r2ka. `+rc2014` for z80 / 8085 / z80n / gbz80 / r2ka. `+yaz180` for z180.

| CPU | sccz80 / 80cc-sp / 80cc-fp | mixed | fallback | ticks vs sccz80 | bytes vs sccz80 |
|-----|----------------------------|------:|---------:|----------------:|----------------:|
| z80 | 25 / 14 / 11 | 25 | 0 | −288926 | +3721 |
| 8085 | 28 / 22 / — | 22 | 0 | −468998 | −966 |
| z180 | 25 / 14 / 11 | 25 | 0 | −288478 | +3794 |
| z80n | 25 / 14 / 11 | 25 | 0 | −290262 | +3641 |
| gbz80 | 4 / 46 / — | 46 | 0 | −658764 | +7406 |
| r2ka | 17 / 33 / 0 | 33 | 9 | −287642 | +1428 |

SDCC recipes print `-compiler=multi does not support the sdcc ABI` and do not invoke zsdcc.

On 8085, ticks tries to assemble Z80-tree `l_long_ne` (`pop ix`). z80asm reports `illegal identifier`. The stitch still succeeds. Helper differential for those symbols is call-only.

### FreeRTOS

```text
zcc +yaz180 -clib=new -compiler=multi -x -O2 --opt-code-speed=all --math32 @freertos.lst
```

| CPU | `@lst` | note |
|-----|--------|------|
| z180 `+yaz180` | OK 165824 | package recipe |
| z80 `+rc2014` | OK 165648 | needs a local `configISR_ORG` stub. Not in the package |
| z80n `+rc2014 -mz80n` | OK 164480 | same stub |
| 8085 / gbz80 / r2ka | fail | `ld a,i` / `di` / `ei` in `portENTER_CRITICAL` |

`tasks.c` on z180: 110 functions, 59 / 26 / 25, mixed=51, fallback=0, −27420 ticks versus all-sccz80.

The FreeRTOS port is Z80 / Z180. Failure on 8085, gbz80, and Rabbit is illegal opcodes in `portmacro.h`. It is not a missing multi stitch.

sccz80 headers still emit a 0-byte `_name` trampoline and a `___name` body. The trampolines tie-break to sccz80. The code that moves is on `___*`.

## Caveats

1. **Not LTO.** Selection is per function in one file. It is not whole-program.

2. **Static ticks is not TIMER.** The metric does not run the program. Literal trips, nested loops, and differential helpers are a compile-time proxy. Expect about 90% of TIMER-ish picks. The score does not match the map. fasta is the file that shows the miss.

3. **Size can miss a TIMER win.** A small sccz80 body that calls `l_gint` / `l_lt` can beat 80cc on size and lose on TIMER. Use `-compiler-metric=ticks` when speed is the goal.

4. **Do not pass frame-pointer flags.** Multi owns `-fframe-pointer` and `-fomit-frame-pointer`. A leftover flag is an error.

5. **Any variant error fails the file.** There is no silent fallback to one compiler.

6. **sdcc is out of scope.** Multi is the classic small-C ABI only. An sdcc clib recipe is an error.

7. **80cc-fp must not call a non-fp body in the same file.** 80cc-fp keeps a live IX frame. sccz80 and 80cc-sp may clobber IX. Pick the cheaper legal fix. Indirect `jp (hl)` is not walked.

8. **math32 is IX-clean.** math48 uses IX. Both compilers already call that library. Multi does not drop 80cc-fp for math48.

9. **Assume the mbf32 IX fix (#3072) is in the tree.** Multi does not special-case mbf32.

10. **Helpers are classic only.** The tool walks `libsrc/l/sccz80`, `libsrc/math/float`, and `libsrc/math/integer`. It does not assemble newlib `libsrc/l/util` or `config_private.inc` dispatchers. An unmeasured helper charges the `call` only.

11. **Named data must match.** If `#ifdef __80CC` changes a global layout, the file is invalid for multi.

12. **Hand-written `.asm` is not rewritten.** copt does not run on library asm.

13. **sccz80 has no frame-pointer path.** The three-way mix is sccz80 + 80cc-sp + 80cc-fp.

14. **80cc itself does not change in this PR.** A later in-compiler host can reuse `z88dk-zcc-multi`. That work is not here.

15. **Dhrystone classic/sccz80 does not compile** (two-dimensional array). Multi needs every variant, so that bench cannot use `-compiler=multi` today.

## Files

| Path | Change |
|------|--------|
| `src/zcc/zcc.c` | `CC_MULTI`, variant compile, metric flags before `-compiler`, `--asm-flags=`, sdcc ABI reject |
| `src/zcc-multi/zccmulti.c` | Parse, measure, select, stitch |
| `src/zcc-multi/zccmulti_ticks.c` | CPU T-states, extras, pair copies, word rotates, `ex`, word loads |
| `src/zcc-multi/zcc-compiler-multi.md` | Specification |
| `Makefile` | Build and install `z88dk-zcc-multi` |
| `changelog.txt` | v2.5 note |
| `test/suites/zcc-multi/` | Acceptance suite |
| `src/zcc-multi/t/` | Host fixtures for the selector |

## Tests

```text
make -C test/suites/zcc-multi
```

Covers two functions, 8085 without 80cc-fp, link, `-v` summary, helper report, discarded-body `defc` stitch past 256 aliases, `jp (hl)` fallback, size-among-all when one variant is unusable, `reason` `only`, counted `djnz`, load-0 wrap, `dec rr` without `or`, 8085 `jp nk`, `ldir` (including gbz80 58 T), differential `mul.c`, operand ticks, Z80 `sra hl` / `ld hl,(ix+d)`, Rabbit pair and `bool` / `add sp` / indexed word load, CPU extra bundles (8085, gbz80, z180, z80n), gbz80 `ex de,hl`, and sdcc ABI reject.


feilipu:
# Why `-compiler=multi` is worth it

## Remeasure 2026-08-22 (after CPU extras / word-op ticks)

Classic Z80 `+test` TIMER, same recipes as the 2026-08-19 table below. Default metric is ticks. sccz80 TIMER was remeasured on this tree the same day.

Percent is `(sccz80 − multi) / sccz80 × 100` against **today’s** sccz80, not the 2026-08-19 sccz80 column. Positive is faster.

| Bench | sccz80 now | PR multi (19 Aug) | multi now | Δ vs sccz80 now | vs PR multi |
|-------|-----------:|------------------:|----------:|----------------:|-------------|
| sieve | 4 769 143 | 4 769 143 | 4 769 143 | 0% | same. all sccz80 |
| fannkuch | 83 264 076 | **68 730 757** | 69 691 670 | **+16.3%** | hotter loop is 80cc-sp, not 80cc-fp |
| fasta | 123 137 988 | 124 859 727 | 150 201 804 | **−22.0%** | 80cc-sp `_gen_random` (PR ticks kept sccz80) |
| mandelbrot | 1 030 978 885 | 1 030 978 885 | 1 030 978 885 | 0% | same. all sccz80 |
| n-body | 797 522 480 | 797 320 160 | 809 458 163 | −1.5% | 80cc-sp won `_advance` / `_energy` / `_offset_momentum` / `_scale_bodies` |
| pi | **1 992 348 390** | 4 028 061 102 | 1 997 952 293 | −0.28% | sccz80 baseline halved vs 19 Aug. multi picks 80cc-sp |
| binary-trees | 149 258 200 | 143 203 779 | **142 510 225** | **+4.5%** | 80cc-fp `_BottomUpTree` lost. TIMER matches the old **size** row |
| sorting | 44 474 249 | 38 866 533 | 38 866 533 | **+12.6%** | same mix |
| spectral-norm | 4 210 446 635 | 4 201 071 927 | 4 210 428 394 | 0.00% | `_main` still 80cc-sp; TIMER ≈ all-sccz80 |
| whetstone | 361 981 590 | 360 699 646 | 359 722 820 | +0.62% | more 80cc-sp (`_main` / `_PA` / `_P0`) |

Do not paste these numbers into published RESULT tables.

### What changed in the stitch

Word-op and `(ix+d)` scoring moved several 80cc-fp bodies off ticks. `ld hl,(ix+d)` is 38 T on Z80. That used to be unknown (`fallback`). 80cc-fp is now more expensive on the static score, so the default ticks metric prefers 80cc-sp or sccz80.

| Bench | PR ticks mix | Mix now |
|-------|--------------|---------|
| fannkuch | 80cc-fp `_fannkuchredux`; 80cc-sp `_max` `_main` | 80cc-sp `_fannkuchredux` `_max`; sccz80 `_main` |
| fasta | sccz80 `_gen_random` | 80cc-sp `_gen_random`; 80cc-fp `_makeRepeatFasta` |
| n-body | math32 cores sccz80; 80cc-sp `_main` | 80cc-sp four cores; sccz80 `_main` |
| pi | all sccz80 | 80cc-sp `_main` (3228 vs 7505 static T) |
| binary-trees | 80cc-fp `_BottomUpTree` | sccz80 `_BottomUpTree`; 80cc-sp alloc/free |
| whetstone | weak 80cc-sp stitch | 80cc-sp `_main` `_PA` `_P0`; sccz80 `_P3` |

### Issues

1. **fasta is a TIMER loss (−22%).** The static ticks metric now prefers 80cc-sp `_gen_random` (2238 vs 3482). TIMER disagrees. Use `-compiler=sccz80` on this file, or `-compiler-metric=size` is not a fix (size was worse on 19 Aug).

2. **n-body is a small TIMER loss (−1.5%).** 80cc-sp wins the float cores on static ticks and is larger (selected bytes 5516 vs sccz80 3748).

3. **fannkuch is still a win (+16.3%), not +17.5%.** 80cc-fp lost the hot loop on ticks (4469 vs 80cc-sp 4206). TIMER of 80cc-fp was better. Static ticks is not TIMER.

4. **pi sccz80 TIMER is half the 19 Aug number** (1.99e9 vs 4.03e9). PRINTF digits match between sccz80 and multi (`31415926…`). Whole-program PRINTF sccz80 is 1.996e9. The 19 Aug 4.03e9 row is stale vs this tree (integer runtime, not the selector). Multi then picks 80cc-sp and is 0.28% slower than **today’s** sccz80. Static ticks (3228 vs 7505) overstates the 80cc win.

5. **binary-trees TIMER now matches the old size row** (+4.5%). 80cc-fp `_BottomUpTree` is no longer cheapest on ticks.

6. **Dhrystone** was not remeasured. sccz80 still rejects the 2-D array. Multi needs every variant.

## Original note 2026-08-19 (PR #3073)

Classic Z80 `+test` TIMER, 2026-08-19. Baseline is the same flags with sccz80. Multi ran both metrics. The default is now **ticks** (static T-state sum). Size is `-compiler-metric=size`.

Percent is `(sccz80 − multi) / sccz80 × 100`. Positive is faster.

Dhrystone compiled, then hung in `z88dk-ticks` (TIMER_STOP not reached). No TIMER for that bench.

## The three outcomes

### 1. Stitch wins — mix two compilers in one file

These files have more than one C function. The cheaper body is not always from the same compiler.

| Bench | sccz80 TIMER | multi ticks | Δ | What was mixed |
|-------|-------------:|------------:|--:|----------------|
| **fannkuch** | 83 264 076 | **68 730 757** | **+17.5%** | 80cc-fp `_fannkuchredux`. 80cc-sp `_max` and `_main`. |
| **sorting** | 44 474 249 | **38 866 533** | **+12.6%** | 80cc-sp `_g_rand`, `_main`, `_ascending_order`. sccz80 `_perform_sort`. |
| **binary-trees** | 149 258 200 | 143 203 779 | +4.1% | 80cc-sp alloc/free. sccz80 `_ItemCheck` and `_main`. 80cc-fp `_BottomUpTree`. |

A single `-compiler=80cc` or `-compiler=sccz80` cannot do this. One file, one compiler, one set of bodies.

fannkuch is the proof. Size keeps the small sccz80 hot loop (`_fannkuchredux` 634 vs 729 bytes). That loop calls `l_gint` / `l_lt`. TIMER only drops 1.2%. Ticks charges those helpers per call and picks 80cc-fp. TIMER drops 17.5%. Four extra bytes in the binary.

sorting is the same mix on both metrics. 80cc-sp wins the RNG. sccz80 keeps the thin sort wrapper.

### 2. Multi picks the right compiler for the whole file

These files are one hot function, or every function scores the same way. Multi selects all sccz80. The binary matches a plain sccz80 build. You do not need to know that in advance.

| Bench | TIMER | Size | Selection |
|-------|------:|-----:|-----------|
| sieve | 4 769 143 | 9191 | `_main` sccz80 |
| mandelbrot `--math32` | 1 030 978 885 | 3750 | `_main` sccz80 (834/4730 vs 80cc 1408/7480 and 2031/9581) |
| pi | 4 028 061 102 | 7160 | `_main` sccz80 |
| whetstone size | 361 981 590 | 9950 | all four functions sccz80 |
| spectral-norm size | 4 210 446 635 | 5718 | all five functions sccz80 |

n-body is almost this case. The math32 cores stay sccz80. Only `_main` goes to 80cc-sp. TIMER +0.03%. That is noise.

Ticks will mix whetstone `_PA`/`_P0` and spectral `_eval_A_times_u` into 80cc-sp. TIMER gain is 0.2–0.4%. The binary grows. That is a weak stitch, not a reason to turn multi off.

### 3. Stitch is a bad idea — fasta

fasta `--math32` is slower than all-sccz80.

| Metric | TIMER | Δ vs sccz80 | Bytes |
|--------|------:|------------:|------:|
| sccz80 | 122 977 988 | | 4275 |
| multi size | 159 098 446 | **−29.4%** | 6059 |
| multi ticks | 124 859 727 | −1.5% | 5213 |

Size picked 80cc-fp for `_makeRandomFasta`. That caller uses IX. The elevate rule then forced `_gen_random` and `_selectRandom` to 80cc-fp even though sccz80 is smaller and cheaper on the static scores (`_gen_random` 452/2211 vs 737/3431). The mix is legal. It is the wrong mix for TIMER.

Ticks keeps sccz80 `_gen_random` and is only 1.5% worse. Still not a win.

Use `-compiler=sccz80` on this file, or accept a small loss.

## TIMER table

| Bench | sccz80 | multi size | Δ size | multi ticks (default) | Δ ticks |
|-------|-------:|-----------:|-------:|----------------------:|--------:|
| sieve | 4 769 143 | 4 769 143 | 0% | 4 769 143 | 0% |
| fannkuch | 83 264 076 | 82 253 475 | +1.2% | **68 730 757** | **+17.5%** |
| fasta | 122 977 988 | 159 098 446 | −29.4% | 124 859 727 | −1.5% |
| mandelbrot | 1 030 978 885 | 1 030 978 885 | 0% | 1 030 978 885 | 0% |
| n-body | 797 522 480 | 797 320 160 | +0.03% | 797 320 160 | +0.03% |
| pi | 4 028 061 102 | 4 028 061 102 | 0% | 4 028 061 102 | 0% |
| binary-trees | 149 258 200 | **142 510 225** | **+4.5%** | 143 203 779 | +4.1% |
| sorting | 44 474 249 | **38 866 533** | **+12.6%** | 38 866 533 | +12.6% |
| spectral-norm | 4 210 446 635 | 4 210 446 635 | 0% | 4 201 071 927 | +0.22% |
| whetstone | 361 981 590 | 361 981 590 | 0% | 360 699 646 | +0.35% |

binary-trees is the one bench where **size** beat ticks on TIMER (4.5% vs 4.1%). The default still picks ticks.

## What to take from this

1. **Use multi when the file has more than one function and the hot loop may inline better in 80cc.** fannkuch and sorting are that case.
2. **Use multi when you do not want to guess the compiler.** On mandelbrot, pi, and sieve it selects sccz80 and you get the sccz80 binary.
3. **Do not expect a win on a helper-heavy float file with one hot function.** 80cc-fp glue around `l_f32_*` is larger and slower. Multi leaves sccz80 in place.
4. **IX elevate can undo a good metric pick.** fasta size is the example. An 80cc-fp caller pulled worse callees with it.
5. **Default ticks, not size.** Size missed the fannkuch 17% TIMER win. The static ticks metric is not a profiler. It was right enough here.

## Recipes

All `zcc +test -vn -m -D__Z88DK` plus:

| Bench | Extra flags |
|-------|-------------|
| sieve | `-O2 -DSTATIC -DTIMER -lndos` |
| fannkuch | `-O2 -DSTATIC -DTIMER -lndos` |
| fasta | `-O2 -DSTATIC -DTIMER --math32 -lndos -pragma-define:CRT_HEAP_AMALLOC=1` |
| mandelbrot | `-O3 --opt-code-speed=inlineints -DSTATIC -DTIMER --math32 -lndos` |
| n-body | `-O2 -DSTATIC -DTIMER --math32 -lndos` |
| pi | `-O2 -DSTATIC -DTIMER -lndos` |
| binary-trees | `-O2 -DSTATIC -DTIMER --math-mbf32 -lndos -pragma-define:CRT_HEAP_AMALLOC=1` |
| sorting | `-O2 -DTIMER -DSTYLE=0 -DNUM=5000 -lndos` |
| spectral-norm | `-O3 --opt-code-speed=inlineints -DSTATIC -DTIMER --math32 -lndos` |
| whetstone | `-O2 -DSTATIC -DTIMER --math32 -lndos` |

Raw TSV: `/tmp/zcc-multi-classic-benches/results.tsv`.

suborb:
This is (genuinely) an interesting idea. I'd be interested to see how it behaves on "real-life" projects such as fatfs, zxnext_magnetic (https://gitlab.com/strandgames/brahman/-/tree/master/zxnext_magnetic - I have a version that runs on +test).

To up-the-permutations it might be worth adding some calling convention command line options to 80cc and sccz80 that can force save frame, sdccdecl etc so that we can get a blend of all the compilers?

feilipu:
I will put some time into testing and extending it over the next week, and see how it looks at the end.

My original shower thought was "how to automate selecting the right 80cc compiler mode? Can we optimise per function?" And, as sccz80 has an identical ABI, I thought it would make a nice project stretch goal to add in the old faithful as a benchmark.

So my design thought is eventually to shrink it down so that it just merges into a 80cc as a function.

The decision metric problem is the interesting one. It is not binary size that is the key attribute, but rather static ticks (including counting helpers per call too). Obviously the 100% solution is to build a full function profiler, but is it possible to get 90% of the way just by counting static ticks in a function? I think the answer is probably, yes!

Another goal was to understand how a frontier LLM (grok-build & grok4.6) would treat a project problem that was not part of their training set. i.e. not another space-invaders or web shop. And, I would say that it performed that task perfectly. It even found and resolved link errors because of sccz80 local symbols not resolving properly, if the function was replaced by 80cc.

Anyway, I'll play with this on some real code (beyond our benchmarks) and see what happens over the next week(s).

feilipu:
The stitch tool now scores T-states from the mnemonic and the operands. It does not use listing line numbers or listing byte counts.

## What changed in the score

- **Ticks from operands.** `ld hl,n` is 10 T. `ld a,n` is 7 T. `ld a,b` is 4 T. A z80asm synthetic `ld de,hl` is 8 T. Listing `nbytes` must not drive T-states. 80cc listings can put a wrong source line on an opcode.
- **Size from listing bytes.** Size still uses the listing. Sum opcode bytes after the function label. Stop at the next function, a data section, or a non-function file-scope label.
- **Unknown and `halt`.** An unknown mnemonic or `halt` fails ticks for that function. `reason` is `fallback`. `jp (hl)` / `(ix)` / `(iy)` still fall back the same way.
- **Load 0 wraps.** `ld b,0` / `djnz` is 256 trips. `ld rr,0` with a valid 16-bit test is 65536. A load of 0 is not an empty loop. The last load of that register wins.
- **`dec rr` does not set Z.** This is true on Z80, 8080, and 8085. `dec rr` then `jp nz` with no `or` is not a counted loop. The portable 16-bit test is `ld a,hi` / `or lo` / `jp nz`.
- **8085 K is general.** `jp k` / `jp nk` after `dec rr` is a 16-bit loop for any literal N. K sets when the pair goes 0 to −1. `ld rr,0` still wraps to 65536.
- **Pair loads feed halves.** `ld bc,N` / `ld de,N` / `ld hl,N` feed B/D/H from the high byte and C/E/L from the low byte. A 0 byte is 256 for that half.
- **Counter rewrite kills the trip.** `inc` or `dec` of the counter inside the span is not a counted loop. Trip stays 1.
- **Block wrap.** `ldir` / `cpir` and friends use `ld bc,0` as 65536 repeats. Unknown BC is still `ZCCMULTI_BLOCK_REP` (2).
- **Helper walk.** Classic integer leaves sit under `libsrc/math/integer` as well as `libsrc/l/sccz80` and `libsrc/math/float`. Skip `obj/`. Do not assemble `config_private.inc` dispatchers.

The `ldimm` fixture is the operand-tick proof. `ld hl,1` / `ret` is 20 T. `nop` / `nop` / `ret` is 18 T. The 80cc-sp nops win. A listing-byte count would score `ld hl,1` as 4 T, the same as `nop`. That pick is wrong.

## What did not change

Differential helper charge is the same. Shared `l_*` calls still cancel.

IX mix is the same. Elevate versus demote. Cheaper legal fix. Tie keeps elevate. 80cc-sp stays IX-unsafe for an 80cc-fp caller.

The TIMER picks in the previous comment were not remesured. Operand ticks can move a static score. This pass did not run `z88dk-ticks` on the benches.

## Clean-up in the tool

Dead seams are gone. The score path has one ticks API.

- Ticks API no longer takes a listing byte count.
- Listing parse no longer returns a source line number.
- Control ops share `parse_control` (`CTL_JP` / `JR` / `DJNZ` / `CALL` / `RET`). Callee names use that parse.
- Intra-TU helper extras add to the base score. They do not reset it.
- Unused line-kind tags and discarded counters are gone.

## Suite

```text
make -C test/suites/zcc-multi
```

New host fixtures:

| Fixture | Lock |
|---------|------|
| `t/loop_b0.asm` | `ld b,0` / `djnz` is 256 |
| `t/loop_bc0.asm` | `ld bc,0` / `or` through A is 65536 |
| `t/loop_bc_nz.asm` | `dec bc` / `jp nz` without `or` stays small |
| `t/loop_8085_k.asm` | 8085 `ld bc,8` / `jp nk` multiplies |
| `t/loop_8085_k0.asm` | 8085 `ld bc,0` / `jp nk` wraps |
| `t/ldir_bc0.asm` | `ldir` with `ld bc,0` is 65536 repeats |
| `t/ldimm.asm` vs `t/nops.asm` | `ld hl,1` is 20 T. Two `nop` then `ret` is 18 T |

## Files in this update

`src/zcc-multi/zccmulti.c`, `src/zcc-multi/zccmulti_ticks.c`, `src/zcc-multi/zccmulti_ticks.h`, `src/zcc-multi/zcc-compiler-multi.md`, `src/zcc-multi/t/*`, `test/suites/zcc-multi/Makefile`.

feilipu:
# zcc-multi split on FatFs and FreeRTOS

Live compile of z88dk-libraries **ff** and **freertos** with `-compiler=multi`. Default metric is ticks. Re-run 2026-08-21 on `zcc_multi_compile` (`4f83fa1db1`).

The sccz80 recipe is the Preparation line in each package readme, plus `-compiler=multi -v` and `-clib=new`. Multi compiles each C function three ways on Z80/Z180 (sccz80, 80cc-sp, 80cc-fp) and two ways on 8085 (sccz80, 80cc-sp). It then picks one body per function and stitches one object.

That split is the result. On these packages, the stitch is not one compiler for the file. It is a mix of bodies from all live variants.

## The split

### FatFs core (`ff.c`, bound folded)

`ff.c` in the tree does not reach the selector (see blockers below). One local change, `BYTE buf[SZDIRE]`, lets 80cc compile the file. The selector then splits 50 functions on `+rc2014` z80:

| | sccz80 | 80cc-sp | 80cc-fp | selected |
|--|--------|---------|---------|----------|
| Functions won | 27 (54%) | 1 (2%) | 22 (44%) | 50 |
| Bytes | 100754 | 119509 | 112484 | 101587 (**+833** vs sccz80) |
| Ticks | 1069716 | 865897 | 594453 | 862420 (**−207296** vs sccz80) |

mixed=23, ix-elevate=9, fallback=41.

23 of 50 functions do not stay on sccz80. 80cc-fp wins the cheap private helpers (`get_fat`, `put_fat`, `follow_path`, `mount_volume`). sccz80 keeps the public `f_*` APIs. The ticks metric pays 833 extra bytes to cut about 207k static T-states versus all-sccz80.

`ffsystem.c` on yaz180/scz180/hbios is one function. The split keeps sccz80 (`_get_fattime` 42/270 vs 80cc 48/299). On RC2014, NORTC leaves that file empty. `ffunicode.c` is empty with `FF_USE_LFN=0`.

### FreeRTOS (`+yaz180` / z180)

Every translation unit that 80cc accepts is a split, not a whole-file pick:

| File | Funcs | sccz80 | 80cc-sp | 80cc-fp | mixed | Bytes vs sccz80 | Ticks vs sccz80 |
|------|-------|--------|---------|---------|-------|-----------------|-----------------|
| `list.c` | 10 | 5 | 3 | 2 | 5 | −65 | −2797 |
| `event_groups.c` | 19 | 11 | 3 | 5 | 8 | −343 | −6383 |
| `queue.c` | 54 | 31 | 17 | 6 | 23 | −345 | −5280 |
| `stream_buffer.c` | 42 | 21 | 5 | 16 | 21 | −486 | −15529 |
| `timers.c` | 41 | 25 | 10 | 6 | 16 | −81 | −5157 |
| `heap_3.c` | 4 | 4 | 0 | 0 | 0 | 0 | 0 |
| **Total** | **170** | **97** | **38** | **35** | **73** | **−1320** | **−35146** |

73 of 170 functions leave sccz80. 38 go to 80cc-sp. 35 go to 80cc-fp. `stream_buffer.c` is the strongest 80cc-fp file (16 of 42). `queue.c` and `timers.c` take more 80cc-sp.

sccz80 headers emit a 0-byte `_name` trampoline and a `___name` body. The trampolines are 0/0 on every variant and tie-break to sccz80. That inflates the sccz80 count. The split that moves code is on `___*`.

`hooks.c` is empty (`configSUPPORT_STATIC_ALLOCATION=0`). `port.c` and `tasks.c` do not reach the selector.

### How to read the split

The stitch is per function, not per file. A translation unit can contain sccz80, 80cc-sp, and 80cc-fp bodies together.

Ticks is the metric. Selected ticks fall. Selected size may rise (FatFs +833). IX mix can demote an 80cc-fp winner (`fallback`) or force a callee onto 80cc-fp (`ix-elevate`). Those are part of the split, not a failed compile.

8085 (`ff_85`) runs two variants only. No 80cc-fp.

---

## Library compile

The package `@lst` recipes fail because multi requires every variant. One 80cc error stops the file, and that stops the library job. Files that already split are discarded with it.

```text
zcc +<target> -clib=new -compiler=multi -x -O2 --opt-code-speed=all --math32 @ff.lst
zcc +rc2014 -clib=new -m8085 -compiler=multi -x -O2 --opt-code-speed=all -D__DISABLE_BUILTIN --math32 @ff.lst
zcc +<target> -clib=new -compiler=multi -x -O2 --opt-code-speed=all --math32 @freertos.lst
```

| Package | Target | CPU | Result |
|---------|--------|-----|--------|
| ff | `+rc2014` | z80 | Fail on `ff.c` |
| ff | `+yaz180` | z180 | Fail on `ff.c` |
| ff | `+scz180` | z180 | Fail on `ff.c` |
| ff | `+hbios` | z80 | Fail on `ff.c` |
| ff_85 | `+rc2014 -m8085` | 8085 | Fail on `ff.c` |
| freertos | `+yaz180` | z180 | Fail on `portable/port.c` |
| freertos | `+scz180` | z180 | Fail on `portable/port.c` |

sccz80-only controls of the same lines succeed: ff `+rc2014` → `ff.lib` 125128 bytes. freertos `+yaz180` → `freertos.lib` 213696 bytes.

FatFs list order: `ffsystem.c` and `ffunicode.c` split (or are empty), then `ff.c` fails. FreeRTOS: `heap_3.c` splits, then `port.c` fails. `tasks.c` is not reached.

SDCC recipes (`sdcc_ix` / `sdcc_iy`) are not a split path. The driver prints `-compiler=multi does not support the sdcc ABI` and does not invoke zsdcc. Use `-clib=new -compiler=multi`.

Read-only FatFs was not rebuilt. It uses the same `ff.c`. Library sources were not changed except the diagnostic fold of the FatFs array bound.

## Blockers (80cc)

These stop the selector. They are not a split failure.

**FatFs `ff.c:5303`.** `BYTE buf[FF_FS_EXFAT ? SZDIRE * 2 : SZDIRE]`. `FF_FS_EXFAT` is 0. sccz80 folds it. 80cc reports `Expecting constant expression`. That is the only 80cc error in `ff.c`. Fold the bound to `SZDIRE` and the split above is the result.

**FreeRTOS `portable/port.c:162`.** `__naked` `timer_isr` is not one asm block. On YAZ180, `configINCREMENT_TICK()` is a C call. SAVE/RESTORE are `do { __asm__(…) } while(0)`. 80cc: `__naked function 'timer_isr' may contain only an asm block`. sccz80 accepts this port.

**FreeRTOS `tasks.c`.** After `port.c`, 80cc still fails `vTaskPrioritySet`: `AST_DECL: kind 1 not yet supported` / `cannot translate 'vTaskPrioritySet' to IR`. sccz80 compiles `tasks.c`.

A package `.lib` from multi needs 80cc to accept those constructs, or a source change the packages can keep. The split on every file that does compile is already the multi behaviour this PR ships.

feilipu:
# PR #3073 — FatFs and FreeRTOS after the 80cc parse fixes

Paste for a PR comment. Not a tree or RESULT-table update.

Master #3079 (`80cc-multi-fixes`) is on this branch. 80cc now folds a constant-condition `?:`, accepts a `__naked` body with more than one asm block and argument-free C calls, folds a parenthesised `sizeof` in a static initialiser, and does not hoist a void-typed subexpression. Those were the FatFs and FreeRTOS compile blockers. The live packages were rebuilt with `-compiler=multi` on the package recipes (`-clib=new -x -O2 --opt-code-speed=all --math32 @lst`).

The stitch is still per function. That mix is the result.

## The split

### FatFs (`ff.c`, unpatched)

The ternary array bound in `f_rename` now compiles. The `@ff.lst` recipe completes. No source fold.

**Z80** (`+rc2014 -clib=new`): 50 functions. 20 stay on sccz80. 30 go to 80cc-fp. mixed=30. Selected ticks 682493 versus all-sccz80 1069716 (**−387223**). Selected size 104410 versus 100754 (**+3656**).

**8085** (`+rc2014 -clib=new -m8085`): two variants only. 50 functions. 28 sccz80, 22 80cc-sp. mixed=22. fallback=0. Selected ticks 573840 versus 1023884 (**−450044**). Selected size 92059 versus 93025 (**−966**). Four larger 80cc-sp bodies win on ticks (`_f_open`, `_f_lseek`, `_f_expand`, `_create_chain`).

The ticks table now models the ten 8085 extended ops under Zilog names. Stand-alone 80cc `L_*` labels are not opcodes. Without that, every 8085 function fell back to size.

`ffunicode.c` is empty with `FF_USE_LFN=0`. RC2014 `ffsystem.c` is empty without RTC.

### FreeRTOS (`+yaz180`)

`portable/port.c` and `tasks.c` now reach the selector. `port.c` keeps sccz80 (12 functions). The `@freertos.lst` recipe completes (`freertos.lib` 165880 bytes versus sccz80 213732).

| File | Funcs | sccz80 | 80cc-sp | 80cc-fp | mixed | Bytes vs sccz80 | Ticks vs sccz80 |
|------|-------|--------|---------|---------|-------|-----------------|-----------------|
| `list.c` | 10 | 5 | 3 | 2 | 5 | −65 | −2797 |
| `event_groups.c` | 19 | 11 | 3 | 5 | 8 | −343 | −6383 |
| `queue.c` | 54 | 31 | 17 | 6 | 23 | −345 | −5280 |
| `stream_buffer.c` | 42 | 21 | 5 | 16 | 21 | −486 | −15529 |
| `tasks.c` | 110 | 55 | 22 | 33 | 55 | −1224 | −45223 |

`tasks.c` is a mix (55 of 110 leave sccz80). An earlier stitch failed: copt left more than 256 `defc i_N = i_M` aliases, and aliases after that cap still pointed at dropped labels. The selector now tracks the whole pool. `z80asm` accepts the stitch.

sccz80 headers still emit a 0-byte `_name` trampoline and a `___name` body. The trampolines tie-break to sccz80. The code that moves is on `___*`.

## Library compile

```text
zcc +rc2014 -clib=new -compiler=multi -x -O2 --opt-code-speed=all --math32 @ff.lst
zcc +rc2014 -clib=new -m8085 -compiler=multi -x -O2 --opt-code-speed=all -D__DISABLE_BUILTIN --math32 @ff.lst
zcc +yaz180 -clib=new -compiler=multi -x -O2 --opt-code-speed=all --math32 @freertos.lst
```

| Package | Target | Result |
|---------|--------|--------|
| ff | `+rc2014` z80 | **OK** `ff.lib` |
| ff_85 | `+rc2014 -m8085` | **OK** `ff_85.lib` |
| freertos | `+yaz180` | **OK** `freertos.lib` |

SDCC recipes still print `-compiler=multi does not support the sdcc ABI` and do not invoke zsdcc.

These two live packages now build mixed libraries from the package recipes. There is no remaining compile or stitch blocker on this path.

## Ticks extras

The ticks table models CPU extras under Zilog names on every model. Unknown extras set `fallback`. Intel and Nintendo aliases are not matched. Bundle sums: 8085 97, gbz80 144, z180 94, z80n 175. `slp` and `jp (c)` stay fallback.

Pair copies (`ld de,hl`) and word rotates (`rl` / `rr` / `sra` on a pair) use native time when the chip has the op. Otherwise they use two 8-bit or CB ops. A documented helper cost is that cost. An unknown helper is fallback. 80cc writes the mnemonic. Rabbit `rr hl` is 2 T. `ld de,hl` is 4 T on Rabbit and 8 T on Z80. 8085 `rl de` is 10 T. Z80 `sra hl` is 16 T. 80cc-fp `ld hl,(ix+d)` is 38 T on Z80 (two indexed loads), 11 T on Rabbit. gbz80 `ex de,hl` is 56 T (`push hl` / `push de` / `pop hl` / `pop de`). gbz80 `ex (sp),hl` is 148 T. Repeating `ldir` uses `21 × BC` on every CPU.

If any variant of a function is unusable, the pick is size among all bodies. The tool does not mix ticks of two usable bodies with size of a fallback body. A function that exists in only one variant is still emitted (`reason` `only`).

## Remeasure 2026-08-22 — benches and CPU matrix

Word-op, `(ix+d)`, and CPU-extra ticks are in the selector. Classic `+test` TIMER and the live FatFs / FreeRTOS packages were rebuilt. Do not paste TIMER numbers into published RESULT tables.

### Classic benches vs the 19 Aug table

See `zcc-multi-classic-benches.md`. Short form:

- **Unchanged:** sieve, mandelbrot, sorting.
- **Still a win, smaller:** fannkuch +16.3% (was +17.5%). 80cc-fp lost `_fannkuchredux`.
- **Still a win, different mix:** binary-trees +4.5% (matches the old size row). 80cc-fp `_BottomUpTree` lost.
- **Loss:** fasta −22% (was −1.5%). 80cc-sp `_gen_random`. n-body −1.5% (was +0.03%).
- **pi:** sccz80 TIMER on this tree is 1.99e9, not 4.03e9. PRINTF digits match. Multi picks 80cc-sp and is 0.28% slower than today’s sccz80.

fasta is the file to keep on `-compiler=sccz80`.

### FatFs `ff.c` (`-clib=new -compiler=multi -O2 --opt-code-speed=all --math32`)

`@ff.lst` completed on every CPU in the matrix. `+rc2014` for z80 / 8085 / z80n / gbz80 / r2ka. `+yaz180` for z180. No source fold.

| CPU | sccz80 / 80cc-sp / 80cc-fp | mixed | fallback | ticks vs sccz80 | bytes vs sccz80 | `ff_*.lib` |
|-----|----------------------------|------:|---------:|----------------:|----------------:|-----------:|
| z80 | 25 / 14 / 11 | 25 | **0** | 806062 vs 1094988 (**−288926**) | +3721 | 106668 |
| 8085 | 28 / 22 / — | 22 | 0 | 579690 vs 1048688 (**−468998**) | **−966** | 107900 |
| z180 | 25 / 14 / 11 | 25 | 0 | 806648 vs 1095126 (**−288478**) | +3794 | 106672 |
| z80n | 25 / 14 / 11 | 25 | 0 | 804408 vs 1094670 (**−290262**) | +3641 | 106604 |
| gbz80 | 4 / 46 / — | 46 | **0** (was 43) | 732952 vs 1391716 (**−658764**) | +7406 | 109096 |
| r2ka | 17 / 33 / 0 | 33 | 9 | 640027 vs 927669 (**−287642**) | +1428 | 91380 |

Z80 fallback 38 → 0 is the ticks-table work. The 19 Aug / post-#3079 split (20 / 0 / 30, fallback=38, selected ticks 682493) mixed size on unusable 80cc-fp bodies. All 50 functions now use ticks. More 80cc-sp, fewer 80cc-fp.

8085 is the same mix as before (28 / 22, fallback=0). Ticks 579690 vs 573840. Size win −966 bytes is unchanged.

z180 and z80n follow Z80 (three variants, fallback=0).

**gbz80.** `ex de,hl` is 56 T. `ldir` uses the block-repeat pass. fallback **43 → 0**. Mix 4 / 46 (was 30 / 20 under size-among-all). A dump that strips tabs can make `ld hl,sp+n` look like Nintendo `ldhl`. 80cc writes the Zilog form; that op was already 12 T.

**Rabbit:** 80cc-fp wins nothing. `bool` / `rr` / `rl` appear in the stitch and score. Nine functions still fallback.

**8085 helper charge:** ticks tries to assemble `libsrc/l/sccz80/5-z80/i32/l_long_ne.asm` with `-m8085`. That file does `pop ix` / `jp (ix)`. z80asm reports `illegal identifier`. The stitch still succeeds. Helper differential for those symbols is not measured on 8085.

### FreeRTOS

`+yaz180` `@freertos.lst` still completes (`freertos_z180.lib` 165824 vs 19 Aug 165880).

`tasks.c` on z180: 110 functions, 59 / 26 / 25, mixed=51, fallback=0, **−27420** ticks vs all-sccz80. On z80: 59 / 27 / 24, mixed=51, fallback=0, **−27306**. The 19 Aug yaz180 row was 55 / 22 / 33, **−45223**. 80cc-fp lost share the same way as the benches.

| CPU | `@lst` | note |
|-----|--------|------|
| z180 `+yaz180` | **OK** 165824 | package recipe |
| z80 `+rc2014` | **OK** 165648 | needs harness `configISR_ORG` (below) |
| z80n `+rc2014 -mz80n` | **OK** 164480 | same harness |
| 8085 | fail `port.c` and most TUs | `ld a,i` in `portENTER_CRITICAL` |
| gbz80 | fail | `ld a,i`, `jp PO` |
| r2ka | fail | `ld a,i`, `di`, `ei` (Rabbit uses `ipset` / `ipres`) |

`list.c` compiles on 8085 / gbz80 / r2ka (little or no `ld a,i`). `tasks.c` / `queue.c` / `timers.c` / `stream_buffer.c` / `event_groups.c` / `port.c` do not.

The FreeRTOS port is Z80 / Z180. 8085, gbz80, and Rabbit are not a missing multi stitch. They are illegal opcodes in `portmacro.h`.

**Harness.** Off YAZ180/SCZ180, `FreeRTOSBoardDefs.h` does not define `configSETUP_TIMER_INTERRUPT`, `configISR_ORG`, or `configISR_IVT`. `timer_isr` emits `PHASE configISR_ORG` and z80asm fails. A local stub is enough for z80 / z80n:

```c
#define configTICK_RATE_HZ            (256)
#define configISR_ORG                 ASMPC
#define configISR_IVT                 0xFFE6
#define configINCREMENT_TICK()        xTaskIncrementTick()
#define configSWITCH_CONTEXT()        vTaskSwitchContext()
#define configSETUP_TIMER_INTERRUPT() do { } while(0)
#define configRESET_TIMER_INTERRUPT() do { } while(0)
#define configSTOP_TIMER_INTERRUPT()  do { } while(0)
```

Plus `-D__CPU_CLOCK=7372800` on `+rc2014`. That stub is not in the package. It is only a compile probe.

FatFs `ff.c` already has diskio on `+rc2014` (`FF_FS_NORTC=1`). No extra diskio stub was required for this matrix. A target without `__RC2014` / `__YAZ180` / `__SCZ180` / `__HBIOS` still hits `#error - No diskio functions available for your target`.

### Tool issues seen on this run

1. **`-compiler-multi-report` on `@lst` overwrites.** One path, one TU at a time. The remaining TSV is the last file (`ffsystem.c` / `heap_3.c`). Use `-v` or a per-file report.
2. **gbz80 `ex de,hl` is 56 T. `ldir` uses `21 × BC`.** FatFs `ff.c` fallback is 0. A tab-stripped dump of `ld hl,sp+n` looks like `ldhl`; that is not the source.
3. **Static ticks vs TIMER** is visible on fasta, n-body, fannkuch, and pi. The metric is still a compile-time proxy.
4. **8085 helper listings** from the Z80 `5-z80` tree (`pop ix`) do not assemble. Charge is call-only for those helpers.

## CI on #3073

Failed jobs on this PR are `build-on-windows-msys2`. The zcc-multi suite passes on those runs. The fail is later, in `test/suites/long_ir` with `-compiler=80cc -clib=8085`: `long_ir.c:2010: could not flush output (disk full ?)`. That is the MSYS2 runner filling the disk on an 80cc 8085 compile. It is not a multi stitch or ticks bug. The `pull_request` MSYS2 run on `ddb5f54` is green.

PR #3074, opened 2026-08-19
Build changes - speed up + fixes

suborb:
Working through:

* Speeding up the build
* Fixing msys so that it builds
* Dual architecture docker files (will only test when we hit master)
* My dislike of yaml


suborb:
* Msys2 is now fixed - double bug: 80cc builds preprocessed twice and the erroneous second one didn't close the pipe properly
* Parallelised libsrc + test
* Caches actually erm, cache now
* Add more caches for the perl dependencies
* Trim down Cygwin significantly: compile, build enough libraries to run test/suites. This means the z80asm tests no longer run on Cygwin (they need +zx and +zxn which are slow library builds)

Docker dual arch will be validated when merged.

suborb:
Right, here goes...

PR #3094, opened 2026-09-01
libsrc: Build changes

suborb:
* Move targets over to .mak format
* Remove recursive make for graphics and games
* Perl cache should be keyed by arch + perl version


suborb:
Libraries compared against pre-commit:

* STANDARDESCAPECHARS handled properly
* esxdos.lib now only has a single copy of the routines.

Issue #3071, opened 2026-08-18
math32 - fsadd overflow detection is wrong? (of finite + finite numbers)


ped7g:
https://github.com/z88dk/z88dk/blob/1bc5bfd4b04d465c048670db05f7e0772e4b515f/libsrc/math/float/math32/asm/z80/d32_fsadd.asm#L304-L305

considering the latest extension of that code with special "gates" for infinity/NaN handling, it struck me this test `inc c` is checking for 255 -> 0 increment, but that's already too late? Two 254 exp. args with some non zero mantissa bits will produce unpacked exp 255 with non zero mantissa, which as packed value is technically NaN, and the `inc c` test will not trigger to return just Inf with carry set in such case.

Is my understanding correct?

So the technically more correct way would be
```
  inc c
  ld a,c
  inc a  ; detect new exp reaching 255
  jr Z,foverflow
```
(or inc c : inc c : jr Z : dec c if A would become used there).

I'm working on optimized version of this file, so I can patch it in my new version (I have to extend it with Inf/NaN handling anyway because I started before that was added). I just need confirmation my understanding is right.

feilipu:
The goal was/is to minimise impact of specials to the hot path, being the regular numbers, and to keep specials handled off that hot path.

So it is possible that some specific cases will slip through when the goal is framed that way.

The current impact on regular numbers flow is about 0.7%. So, not expensive.

I'll comment further, but this just to understand the goal I'm trying to achieve. And guidance for further analyses.

feilipu:
Yes. That is the right way to fix it.

```
  inc c
  ld a,c
  inc a  ; detect new exp reaching 255
  jr Z,foverflow
```

In consideration to just drop the test, leaving just `inc c`, but that would allow INF*INF to be a quiet NaN.
So your suggestion is most correct.

__EDIT__ I've checked and the [cost is only 0.013% by adding the two instructions](https://github.com/z88dk/z88dk/commit/86d1096058ae4221fb8853d0961c2ad0a1aa907e) to the fsadd32 path.
So it is not an issue to make the improvement.

I'll leave it open for now, pending your wider review.
But do let me know if your timeline changes, and I'll close it with a simple patch.

ped7g:
fsadd32 -> this one is reused in multiplication IIRC? So falling out of range with correct INF vs NaN is probably of more value.

[...]

PR #3096, opened 2026-09-03
math32 - mechanical review

feilipu:
Pushing the code through 7 mechanical reviewers found some issues I'd overlooked.
And then I found some more issues.

And drive-by am9511 fixes too.

And uncovered unused z80asm paths which needed fixed.

---

## math32 reviews
 
Source: untracked `math32_reviews/math32-convergence-review.md` plus seven provider files. Z80-focused. Same class of bugs on 8080 / 8085 / gbz80 where the code exists.
 
The arithmetic cores (mul, restoring div, jam-sticky add, RNE pack) are sound. Most real defects were C wrappers and a few edges.
 
| ID | Claim | Verdict | Action |
|----|--------|---------|--------|
| C1 | `expf` guards `#if 0` | True (`exp2`/`exp10` too) | Guards live. `exp(100)` → `HUGE_POS_F32`, `exp(-100)` → `+0` |
| C2 | `sqrt(-0)` is NaN | False path. Result was **+0** | Return **−0**. `invsqrt(-0)` was already −Inf |
| C3 | `sinf` SP reduction / `uint16_t j` | True. Same in `cosf` | `\|x\|≥128` uses `fmod(2π)`. Stays in `[-1,1]`. Not Payne–Hanek |
| C4 | `fsmul32*` one-sided zero check | True as code. Helpers are not public | Removed the extra check. Callers already classify ±0 |
| C5 | Share specials gates | Architecture, not a defect | Not extracted. Hot path unchanged |
| C6 | Square kernel for sqrt NR | Kernel exists for IEEE `sqr()`. NR uses 32-bit `fsmul32x32` | Not rewired |
| C7 | normalize packs exp=0 + mant | True vs FTZ | Flush exact exp 0. Signed zero |
| C8 | `ld d,b` after one `exx` | False. Wrong bank | No change |
| C9 | Factor NR copies | Size vs ticks | **Forbidden.** NR stays unrolled |
| C10 | NaN payload | Policy `0x7FFFFFFF` | No change |
| MiMo E1 | add `rra` align wrong | False. `1+2^{-n}` bit-identical to host | Suite lock only |
| Ling #1 | div misses `0x7FC00000` | False. Quiet bit in `E` | Suite lock only |
| Muse E1 | add overflow → NaN | Already fixed (`#3071`) | No change |
| Muse T3 | drop `or a` in div restore | False. Restore leaves CF=1 | No change |
 
## math32 C wrappers
 
All 25 files under `libsrc/math/float/math32/c/*.c` were read.
 
| Function | Bug | Fix |
|----------|-----|-----|
| `acosf` | `atan(y/x)` drops quadrant | `atan2(y,x)` |
| `asinf` | `asin(±1)` sign wrong | `atan2(x,y)` |
| `asinhf` | always `fabs` (even) | restore sign |
| `acoshf` | no domain | `x<1` → NaN |
| `atanhf` | `\|x\|≥1` hangs / junk | ±1 → ±Inf, else NaN |
| `powf` | `x≤0` → 0 | integer negative base. `pow(-2,3)` ≈ −8 |
| `log*` | `x≤0` all NaN | `log(0)` → −Inf. `x<0` still NaN |
| recip | `invf` | `1.0/x` (restoring `div`) |
| `pow(x,-0.5)` | — | keep `invsqrtf` |
 
`hypot` stays `sqrt(sqr(x)+sqr(y))`. That already uses `invsqrt` inside `sqrt` (`s * invsqrt(s)`). Do not rewrite as `1.0/invsqrt(s)`.
 
`sinhf` is `(e^x − 1/e^x)/2`. No sqrt.
 
Bit punning uses `union float_long` from `c/m32_math.h`. Cast `NAN_*` / `INFINITY_*` to `int32_t` for zsdcc.
 
C regen: `make -C libsrc/math/float/math32/c` plus `8080` / `8085` / `gbz80`. Then rebuild products into `lib/clibs/`.


libsrc/math/float/math32/asm/z80/f32_fsnormalize.asm
```diff
@@ -74,6 +74,7 @@ PUBLIC m32_fsnormalize
     ld a,c
     sub d                       ; a = final exp
     jr c,normzero
+    jr z,normzero               ; FTZ: exp==0 is signed zero, not a denormal
```
ped7g: you can replace extra `jr` with `inc a` by incrementing all `ld d,<init_value>` by one (ie. 1, 17, 9).
That's -1 byte and -3 T states on non-zero path and -12T on become-just-zero path.

I'm wondering about the signed zero, is this just to conform to some spec of IEEE, or is this practical in some way? I may be blind, but I'm struggling to see how +0 differs from -0, ie. why/where losing the sign of zero matters. I guess there may be some "obvious" use case I'm just missing.

feilipu: Thanks. I'll make that change, with a few additional comments to make clear what is happening since it makes the code non-obvious.

My instructions to the reviewers was to comply against IEEE-754 policy, and generally they found minus zero handling lacking. I won't bore you with math theory, as I struggled with first year engineering maths and haven't recovered from that trauma, but [here is a wiki link](https://en.wikipedia.org/wiki/Signed_zero) to why.

feilipu:
## Why MSYS2 ticks --- not actually ticks at fault

The committed _m32_sinf asm is the same. The host that differs is the MinGW z88dk-ticks (MSYS2 MINGW64 gcc -O2). Cygwin gcc ticks and MSVC ticks execute that function correctly.

src/ticks/ticks.c is one huge cpu_run switch. MinGW gcc is the only compiler in this matrix that both builds that file and fails this path. That is the usual signature of a miscompiled opcode handler, not of a math32 source bug.

The suite never hit this poly before: exp(100) returns via the overflow clamp; fmod never calls sinf; add/sqrt/div do not run this octant loop. The 128 vector only made it visible.

Not pinned to one instruction. A ticks single-step of sin(2.336) (or probes of sin(1), sin(π/2), j after the (int) cast) would show whether ___fs2sint/___uint2fs or m32_polyf is the first bad value. Until that, treating this as a math32 C bug or loosening the assert would hide a MinGW ticks defect.

feilipu:
## Why z80asm

z80asm parsed integer literals in scan_num() with a host long and stopped at LONG_MAX. On MinGW LLP64, long is 32-bit signed, so LONG_MAX is 0x7FFFFFFF. A hex constant with bit 31 set, such as DEFQ 0xBE2AAAA3, overflowed that limit. The assembler then stored 0x7FFFFFFF instead of the given bits. Linux and Cygwin use LP64, so the same constant fits in long and assembles correctly.

Those DEFQ lines are the IEEE coefficients for math32 sin and cos. The MSYS2 library build therefore linked a polynomial with qNaN in place of the negative terms. Horner then treated that value as exponent 255 and returned the huge finite 0x7ece0965 for sin(128). fmod and sqrt did not use that table, so they still passed.

PR #3106, opened 2026-09-14
math48 - fix z180 library build

feilipu:
The m48 (math48) library only carries CPU:z80 members, which z80asm refuses to pull into -mz180 links.  Build a CPU:z180 m_z180 product from the same m48 sources so newlib z180 -lm resolves (zcc.c swaps -lm to -lm_z180 for newlib z180). 

The z80 m48 library and every other CPU delivery are untouched; z80asm links only pull members whose CPU matches -m.


pauloscustodio:
>The m48 (math48) library only carries CPU:z80 members, which z80asm refuses to pull into -mz180 links.

The reason is that z80 allows the undocumented instructions that z180 does not support, so z180 falls back to 8080 as the compatible CPU. Removing the limitation and allowing z80 code to be linked in a z180 would cause code using the undocumented instructions to fail.

Is this a limitation we should work on?

feilipu:
At the moment I've left it weirdly split, where classic uses math32 (with z180 specific core, and avoid genmath undocumented ops) and newlib uses math48 (in line with the current default).

By weird I mean z80 z80n etc use genmath as classic default, so z180 is the then only 4 bytes z80-ish classic solution.

But that seems to work. I've noted to clean it up at some stage.

PR #3108, opened 2026-09-15
z80n - fix push nnnn big endian code found little endian stacked

feilipu:
`push nnnn` does not round-trip; regenerate Issue_312 baseline.

Test with real hardware before committing.

__EDIT__ `push nnnn` behaves like `ld hl,nnnn` `push hl`. Adjust ticks etc to emulate hardware outcomes.


feilipu:
Spectrum Next v1. latest firmware.

IMG_9023

feilipu:
Hardware result from the photo on this PR (ZX Spectrum Next). The probe ran three times. Every run printed the same lines:

```
push 0x1234 : pop=0x1234  bytes@SP=34 12
push 0x0100 : pop=0x0100  bytes@SP=00 01
pair(10000,1): pop=0x0001  bytes@SP=01 00
OPTION: LITTLE-ENDIAN (push nn == ld hl,nn; push hl)
```

Silicon stores `PUSH nnnn` like `push hl`. For `push 0x1234` the byte at SP is `0x34` and `pop hl` returns `0x1234`.

`z88dk-ticks` does the opposite. `z80n_push_mn()` in `src/ticks/z80n_inst.c` writes the HI byte to the final SP. The same probe on ticks prints:

```
push 0x1234 : pop=0x3412  bytes@SP=12 34
OPTION: BIG-ENDIAN
```

The instruction encoding is still `ED 8A HI LO`. That is fetch order in the opcode stream, not stack order. `src/z80asm/t/issue_0312.t` is correct about the opcode bytes.

This PR removes the `lib/z80n_rules.1` `push nn` copt patterns because ticks does not round-trip. Real hardware does round-trip. Do not merge that deletion.

**Proposed resolution**

1. Change `z80n_push_mn()` so the stack write matches `push hl`: fetch HI then LO from the instruction, then `--sp = HI`, `--sp = LO`.
2. Keep the copt rules on master (`ld hl,nn; push hl` → `push nn`, and the push-pair form). Drop that part of this PR.
3. Keep the `issue_0312.t` encodings. Remove the new comment that `pop hl` returns a byte-swapped word.
4. Keep the probe under `support/benchmarks/z80n_pushnn/`. Return DEHL as `unsigned long` so `bytes@SP` is not taken from HL. After (1), `zcc +test -clib=z80n` must print `LITTLE-ENDIAN`.

F is settled: copt may emit `push nn` for a register push of a constant. Align ticks and the tests to silicon. Relates to #3104.

test/suites/z80n_pushnn/pushnn.asm
```diff
@@ -0,0 +1,32 @@
...
+
+_probe_0100:
+    ld (sptmp),sp
+    push 0x0100         ; ED 8A 01 00
+    jp _snap
+
+_probe_pair:
+    ld (sptmp),sp
+    push 10000          ; ED 8A 27 10
+    push 1              ; ED 8A 00 01
```
ped7g, line 23: I'm confused by this, it puts two words at stack, then falls through into `_snap:` which will read value `10000` into DE, then pop value `1` into HL and `ret` to 10000?

Feels like I'm missing something, but the photo from HW suggests the test did survive and reports also bytes `@SP=01 00` which IMHO should be the value 10000 from DE, ie. `10 27`?

edit: also it reads the value from SP as [D] [E], ie. ED in 16b word-way with swapped endianness. So D=0x10, E=0x27 IMHO in this case, before the `ret` goes to 10000?

feilipu:
The issue is that the assembler rightly encodes the opcode into big endian bytes in the ROM, but they are pushed little endian into RAM by the z80n CPU, so the stack is effectively just like a normal push via a register. But ticks and the copt rules incorrectly treated it as pushing big endian, and flipped the byte order.

test/suites/z80n_pushnn/pushnn.asm
```diff
@@ -0,0 +1,32 @@
...
+; (D = byte at SP, E = byte at SP+1), then pops and returns:
+;   HL = popped 16-bit value, DE = raw stack bytes (D=E0, E=E1).
+SECTION code_user
+PUBLIC _probe_1234, _probe_0100, _probe_pair
+
+sptmp:   defw 0
+
+_probe_1234:
+    ld (sptmp),sp
+    push 0x1234         ; ED 8A 12 34 (assembler encodes HI LO)
```
ped7g, line 12: the way this does `ld (sptmp),sp` here and then 2x `dec hl` in `_snap` this is also verifying the `push nnnn` does sp-=2. This feels a bit redundant as it's not truly protected against some mishap on the byte-count side (let's say the HW would do push only single byte or four bytes, the test will not recover in either case).

So you could as well just `push 0x1234 : jp _snap` and fetch the SP to HL directly in _snap and share that code across all three tests (that does not resolve my confusion about _probe_pair, I'm still not sure how that one works and why).

ped7g:
> The issue is that the assembler rightly encodes the opcode into big endian bytes in the ROM, but they are pushed little endian into RAM by the z80n CPU, so the stack is effectively just like a normal push via a register. But ticks and the copt rules incorrectly treated it as pushing big endian, and flipped the byte order.

yes, the big-endian with Z80N `push nnnn` is only about instruction encoding, ie. how assembler should encode the value of opcode. The functionality is regular push, ie. `push 0x1234` does push 0x1234 into stack from user's point of view. (the 0x1234 is then encoded as bytes `12 34` in the instruction and ends up as `34 12` in stack).

test/suites/z80n_pushnn/pushnn.asm
feilipu: The test is now redundant, now the test result is in. It may not be good code.

But I left it (unlinked) in the test directory just for the record. If you have a Spectrum Next try out the code and you can confirm the result.

feilipu:
> The functionality is regular push, ie. push 0x1234 does push 0x1234 into stack from user's point of view. (the 0x1234 is then encoded as bytes 12 34 in the instruction and ends up as 34 12 in stack).

Good to have a second opinion. 👍 
Another wrinkle smoothed out.

test/suites/z80n_pushnn/pushnn.asm
ped7g: well, this is working code, maybe more complicated than needed, but non-issue.

the `_probe_pair` should NOT work and jump at address 10000. Depending how you test this (ROM) it may somehow recover by accident and report something, if this was the code you tested on HW, then you must have been lucky that it recovered.

test/suites/z80n_pushnn/pushnn.c
```diff
@@ -0,0 +1,48 @@
...
+int main(void)
+{
+    unsigned int p1 = probe_1234();   /* low = pop, high = raw bytes (D high, E low) */
+    unsigned int p2 = probe_0100();
+    unsigned int pp = probe_pair();
+
+    unsigned int pop1 = p1 & 0xffff,  pop2 = p2 & 0xffff,  popp = pp & 0xffff;
+    unsigned int b01 = (p1 >> 8) & 0xff, b11 = p1 & 0xff;      /* b0 = byte@SP, b1 = byte@SP+1 */
+    unsigned int b02 = (p2 >> 8) & 0xff, b12 = p2 & 0xff;
+    unsigned int b0p = (pp >> 8) & 0xff, b1p = pp & 0xff;
```
ped7g:
```suggestion
    unsigned int pop1 = p1 & 0xffff,  pop2 = p2 & 0xffff,  popp = pp & 0xffff;
    unsigned int b01 = (p1 >> 24) & 0xff, b11 = (p1 >> 16) & 0xff;      /* b0 = byte@SP, b1 = byte@SP+1 */
    unsigned int b02 = (p2 >> 24) & 0xff, b12 = (p2 >> 16) &  & 0xff;
    unsigned int b0p = (pp >> 24) & 0xff, b1p = (pp >> 16) &  & 0xff;
```

IMHO you are extracting the same part of 32bit value twice, ie. `pop1` is sourced from same bits as `b01` and `b11` in your test code.

ped7g: in other words, your fixes to ticks and your conclusions are correct, but the test itself is lucky number generator 😁

PR #3109, opened 2026-09-15
CPC -  init floatpack

feilipu:
Move the init_floatpack into the CPC math library.


suborb:
I think if you add an override `fa` variable in the library then you only need to have the crt0_init section once.

Basically make a copy of the original fa.asm into cpc. -lcpcmath is added first to the library search list so that label will be found first.

